PR #148, opened 2015-01-01
Fixed class name resolution via get_class() and static::class.

BaldaI:
Hi Barry.

My Base Model has some of relations:
```php
	/**
     * Relation to the parent.
     *
     * @return BelongsTo
     */
    public function parent()
    {
        return $this->belongsTo(static::class, static::PARENT_ID);
    }

    /**
     * Relation to children.
     *
     * @return HasMany
     */
    public function children()
    {
        return $this->hasMany(get_class($this), static::PARENT_ID);
    }
```

It is incorrect result of `./artisan ide-helper:models -W` command
```php
/**
 * ModelName
 *
 * @property-read static $parent
 * @property-read \Illuminate\Database\Eloquent\Collection|\get_class($this)[] $children
 */
```
And it is expected result
```php
/**
 * ModelName
 *
 * @property-read ModelName $parent
 * @property-read \Illuminate\Database\Eloquent\Collection|ModelName[] $children
 */
```

The attached patch fix this.

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -331,9 +331,9 @@ class ModelsCommand extends Command
                         $search = '$this->' . $relation . '(';
                         if ($pos = stripos($code, $search)) {
                             $code = substr($code, $pos + strlen($search));
-                            $arguments = explode(',', substr($code, 0, stripos($code, ')')));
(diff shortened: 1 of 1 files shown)


barryvdh:
Thanks

PR #155, opened 2015-01-28
Check for driver method on Auth facade

Flynsarmy:
[October CMS](http://octobercms.com/) built on laravel includes an Auth facade in its [RainLab.User](https://github.com/rainlab/user-plugin) plugin but the resolved class doesn't contain a *driver()* method - so `php artisan ide-helper:generate` gives a nasty error during execution. This PR checks for the existance of the method before attempting to use it.

Ideally I'd like to see your other if statements in this method include the same check, but for the moment this is the only one erroring so I only updated that line.

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -114,7 +114,7 @@ class Generator
     protected function detectDrivers()
     {
         try{
-            if (class_exists('Auth')) {
+            if (class_exists('Auth') && method_exists('Auth', 'driver')) {
                 $class = get_class(\Auth::driver());
                 $this->extra['Auth'] = array($class);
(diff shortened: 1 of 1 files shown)


Flynsarmy:
An explanation would have been nice :(

barryvdh:
Sorry I screwed up with renaming branches which closed al PR s

barryvdh:
That didn't actually work, because the methods are called via the magic __call method.
I implemented a check on the correct facades:
https://github.com/barryvdh/laravel-ide-helper/commit/317ef794a579474e96eaec1d6c1c5e0f7ac6ee5e

Flynsarmy:
Ahh, my bad. Thanks Barry. You're a champ.

PR #169, opened 2015-02-17
Authenticatable Contract for User model

mattwells:
Hi Barry,

I have updated config so that it will map `Authenticatable` contract rather than `UserInterface` for the `App\User` model by default.

Thanks,
Matt


barryvdh:
Thanks

PR #178, opened 2015-03-03
Custom database types support for L5 branch

FractalizeR:
This adds an ability to have custom database types in your database and still have `ide-helper:models` generate correct documentation for them. 

For example, I extensively work with PostgreSQL `jsonb`, `array` and `point` types which are currently not mapped to a valid Doctrine type. This makes ide-helper skip the model entirely. With this PR you can just add the following to your ide-helper config file and have your models documentation generated just fine.

```php
    'custom_db_types' => array(
        'postgresql' => array(
            'citext' => 'text',
            'jsonb' => 'string',
            'point' => 'text',
            'array' => 'text',
            '_int4' => 'string',
        )
    ),
```

`jsonb` is mapped to a string for me (but you can have it mapped to array using `'json_array'` instead of `'string'`. `'_int4'` is used as an internal type for `ARRAY[INT]` so I mapped it to `'string'` also (but, of course you can have it mapped to array using `'array'` instead of `'string'` if you like.

If you are using different databases, you can map types differently with this PR.

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -84,6 +84,36 @@ return array(
     
     'interfaces' => array(
         '\Illuminate\Contracts\Auth\Authenticatable' => config('auth.model', 'App\User'),
-    )
+    ),
+
+    /*
+    |--------------------------------------------------------------------------
+    | Support for custom DB types
+    |--------------------------------------------------------------------------
+    |
+    | This setting allow you to map any custom database type (that you may have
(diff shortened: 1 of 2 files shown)


barryvdh:
Thanks.

FractalizeR:
Welcome. Thank you for the great addon!

danjesus:
:+1:

PR #179, opened 2015-03-03
Support for using namespaced models in relations (L4)

FractalizeR:
I'm using relations like this:

```php
    public function contractor()
    {
        return $this->belongsTo(User::class);
    }
```

But my `User` model is inside `FH\Stock\Model` namespace. Previously, `ide-helper` incorrectly formed docblock placing my model into root namespace. This patch fixes this.


barryvdh:
Rather not add new features to L4. THanks anyway.

Issue #108, opened 2014-08-27
Respect default file namespace for docblocks


FractalizeR:
``` php
namespace ForumHouse\Common\Conversion\Converters;

interface IConverter {}

class Factory
{
    /**
     * @param string $name Converter name
     * @returns IConverter
     */
    public function create($name) {}
}
```

`IConverter` in `Factory::create()` is actually `ForumHouse\Common\Conversion\Converters\IConverter`.

But in `_ide_helper.php` it is `\IConverter`

m4tthumphrey:
So there is no way to fix this? I've noticed it for a while but get around it by just defining the type with a `/** @var /Full/Path/To/Class $class */` tag.

mycarrysun:
@barryvdh any word on getting this fixed? showing me all kinds of warnings when trying to make commits...

barryvdh:
Is this still an issue then?

mfn:
@barryvdh this can be closed as duplicate of https://github.com/barryvdh/laravel-ide-helper/issues/627 :} (which IMHO has more/better details)

mfn:
Closing per my comment https://github.com/barryvdh/laravel-ide-helper/issues/108#issuecomment-650617938

PR #180, opened 2015-03-03
Support for using namespaced models in relations (L5)

FractalizeR:
I'm using relations like this:

``` php
    public function contractor()
    {
        return $this->belongsTo(User::class);
    }
```

But my `User` model is inside `FH\Stock\Model` namespace. Previously, `ide-helper` incorrectly formed docblock placing my model into root namespace. This patch fixes this.

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -512,8 +512,9 @@ class ModelsCommand extends Command
     }
 
     /**
-     * @param string $className
+     * @param string                              $className
      * @param \Illuminate\Database\Eloquent\Model $model
+     *
      * @return string
      */
     private function getClassName($className, $model)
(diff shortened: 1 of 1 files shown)


barryvdh:
So this prepends the namespace of the model it's in, so would only work when they have the same namespace right?
And if the model doesn't have a namespace, it just uses `\Classname`?

FractalizeR:
> So this prepends the namespace of the model it's in, so would only work when they have the same namespace right?
Yes, it prepends the namespace of the model, if the model's class name does not start from backslash.

> And if the model doesn't have a namespace, it just uses \Classname?
Then nothing should be appended.

danieliancu:
@FractalizeR Does this PR fixes #108?

FractalizeR:
No, this is for models only.

barryvdh:
Sorry, still not really sure. So what if I have just 'User' or 'App\User' as model? Will it prepend the namespace still? That doesn't seem right.

FractalizeR:
> Yes, it prepends the namespace of the model, if the model's class name does not start from backslash.

That's all it does. 

I think this should be closed because it's more than half a year old. I've switched to Yii now at my current job and have no time to work on this further unfortunately.

barryvdh:
Ok, thanks anways. I'll look into detecting the `::class` and trying to find the namespace on a different way.

mfn:
No need for this PR anymore, this is already working. I just tested this => https://github.com/barryvdh/laravel-ide-helper/pull/870/commits/4992d62226cfb0ae56b00a6dd333ab54c9250d79

mfn:
@barryvdh as mentioned in https://github.com/barryvdh/laravel-ide-helper/pull/180#issuecomment-570691856 , there's no need for this PR anymore, it already works

PR #190, opened 2015-03-29
Add Query\Builder return type to magic scope* methods

MalikinSergey:
It's needed for chain calls like 
`Model::popular()->orderBy('id','desc') /* popular() here is scopePopular() in model class */`


barryvdh:
Thanks

PR #192, opened 2015-03-31
Eloquent mixin

ababkov:
Option to add ```@mixin \Eloquent``` to model doc block if the model extends from Illuminate\Database\Eloquent\Model.

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -43,6 +43,19 @@ return array(
     'model_locations' => array(
         'app',
     ),
+    
+    /*
+    |--------------------------------------------------------------------------
+    | Model options
+    |--------------------------------------------------------------------------
+    |
+    | Define additional options for model docblock generation
+    |
+    */
+    
+    'model_options' => array(
+        'include_eloquent_mixin'=>true
+    ),
 
 
     /*
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -416,12 +416,15 @@ class ModelsCommand extends Command
 
         $properties = array();
         $methods = array();
+        $mixins = array();
         foreach ($phpdoc->getTags() as $tag) {
             $name = $tag->getName();
             if ($name == "property" || $name == "property-read" || $name == "property-write") {
                 $properties[] = $tag->getVariableName();
             } elseif ($name == "method") {
                 $methods[] = $tag->getMethodName();
+            } elseif($name == "mixin"){
+                $mixins[] = $tag->getContent();
             }
         }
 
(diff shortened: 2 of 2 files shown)


ababkov:
@barryvdh Any thoughts on this one?

Cheers.

barryvdh:
Hmm, I'm not 100% sure about this. It will probably solve a few cases, but not all, right? Why not just extend the facade instead?

ababkov:
Mostly on the grounds that the current documentation always extends the Model (http://laravel.com/docs/5.0/eloquent) which leads me to believe that most people will extend the base class rather than the facade. 

Additionally, if you're extending a third party class that extends Model rather than the facade you don't have much choice.

I've been using the mixin for a while with a fair amount of success - my only complaint is that the type hinting on some of the builder methods is a bit annoying as it never type hints static return types for methods like Eloquent::findOrFail() and firstOrFail(). That being said, at least they hint now..

Without putting in special exceptions this is maybe the best we can do.

Anyway, curious to hear your thoughts on potential issues - you've been doing this laravel thing for a while longer than me :)

barryvdh:
It should extend both the Eloquent Builder and the Query Builder, right?

The problem I had with it, was that it flagged some errors/warning, because it was behaving as if it was extending the class. Can't exactly remember what.

But if it is an option I guess it wouldn't hurt.

ababkov:
Not sure on extending the QueryBuilder - the methods in there start to get pretty low level - happy to take your thoughts on this.

On a slightly unrelated note and out of curiosity - do you have any idea why some of the query builder methods in Laravel hint Collection|static[] or Model|static and others omit static.

e.g. findOrFail
@return \Illuminate\Database\Eloquent\Model|\Illuminate\Database\Eloquent\Collection

Surely this should be:
@return \Illuminate\Database\Eloquent\Model|\Illuminate\Database\Eloquent\Collection|static|static[]

If not I might go through at some point and submit a pull request.

Doncode:
Hi. What about findOrFail
Autocomplete does not work for findOrFail (laravel 5)

Lavoaster:
What's the status on this? It'd be helpful for me because I have a preference for staying with the default and just extending `Model` instead of eloquent.
I would however suggest that the config option is disabled by default, so it won't affect anyone that this won't be useful for.

shehi:
+1 for this.

cablegunmaster:
+1 is there a update for this one? :)

ababkov:
@barryvdh - any chance this will be merged in at some point?

m3esma:
+1

m3esma:
+1 — this is a good thing.

CaporalDead:
:+1: please :)

barryvdh:
If we're going to use mixin, why not use the actual classes, instead of the facade?

CaporalDead:
Not sure to get what you said. Can you explain it to me plz ?

ababkov:
Yeah fair enough - I'll sort out next week

barryvdh:
Fixed with https://github.com/barryvdh/laravel-ide-helper/commit/c213be3d3ca95a788aa4f1d9ddf72d4f31700594

PR #205, opened 2015-05-06
Added functionality for platform specific datatypes

petermein:
I've been using ide-helper for a while with MySQL databases. But i switched to PostgreSQL for a recent project. However ide-helper didn't support all datatypes from PosgreSQL. This addition is aimed at creating better support for other database platforms. 

The config has an array with platform from \Doctrine\DBAL and the mappings are specified like a key value pair. 

I included the JSON mapping for PostgreSQL as example. However this needs to be extended to included more types.

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -84,6 +84,22 @@ return array(
     
     'interfaces' => array(
         '\Illuminate\Contracts\Auth\Authenticatable' => config('auth.model', 'App\User'),
+    ),
+
+    /*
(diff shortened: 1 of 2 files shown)


petermein:
Never mind this PR is duplicate the that of FractalizeR

barryvdh:
Okay I've merged that. You can PR some good defaults if you like?

petermein:
Sure, il will make a list for PostreSQL.

PR #211, opened 2015-05-13
Update ModelsCommand.php

MrCue:
Work with tables in different databases

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -226,7 +226,12 @@ class ModelsCommand extends Command
             $databasePlatform->registerDoctrineTypeMapping($yourTypeName, $doctrineTypeName);
         }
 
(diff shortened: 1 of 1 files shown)


EselDompteur:
+1 for that

MrCue <notifications@github.com> schrieb am Mi., 13. Mai 2015 um 17:49 Uhr:

barryvdh:
When is this needed?

EselDompteur:
if you are using different databases in the same project

Barry vd. Heuvel <notifications@github.com> schrieb am Do., 14. Mai 2015 um
10:09 Uhr:

MrCue:
When you generate the model documentation, if your table name references a different database (somedatabase.table) then the documentation doesn't contain any fields from the table.

This change allows for tables from more than one database to have documentation generated.

barryvdh:
Okay so the same db connection, but different table.

PR #215, opened 2015-05-26
Meta command: Support aliases and instances in service container + option to set methods to access container

vincekovacs:
I extended default \Illuminate\Foundation\Application to get service container instances and aliases.

For example:

``` php
   class Application extends \Illuminate\Foundation\Application
    {
        public function getInstances()
        {
            return $this->instances;
        }

        public function getAliases()
        {
            return $this->aliases;
        }
    }
```

And changed default Application instance in bootstrap/app.php:

From:

``` php
    $app = new Illuminate\Foundation\Application(
        realpath(__DIR__ . '/../')
    );
```

To:

``` php
    $app = new Application(
        realpath(__DIR__ . '/../')
    );
```

In config file simple overwrite `meta.methods` option to add custom container resolver method:

``` php
    'meta'            => array(
        'methods' => array(
            '\Illuminate\Foundation\Application::make',
            '\Illuminate\Contracts\Foundation\Application::make',
            '\Illuminate\Contracts\Container\Container::make',
            '\Illuminate\Container\Container::make',
            '\App::make',
            'app',
            '\Application::make', // For example I added Application class make method as container resolver
        ),
    ),
```


mfn:
@barryvdh this is about making `\Barryvdh\LaravelIdeHelper\Console\MetaCommand::$methods` configurable.

I think ppl doing such level of customization (extending Laravels `Application`) can as well customize this command already (`$methods` is protected, thus it can be overridden); I don't think such an edge case warrants the added code.

I vote for dropping the PR.

Issue #240, opened 2015-07-28
ide-helper:models --dir="app/Models" generates two docblocks


tzfrs:
I have a class, that looks like this

    <?php
    namespace FBFeedParser\Models;
    
    class FacebookPostData extends BaseModel
    {
        /**
         * The table name of this class, since the AI of Laravel would add an s to the end of this class
         * @var string
         */
        protected $table = 'facebook_post_data';
    
        /**
         * The attributes that are mass assignable.
         *
         * @var array
         */
        protected $fillable = [
            'facebook_post_id', 'shares', 'likes', 'comments', 'total_count'
        ];
    
        protected $parentMapper = [
            'facebook_post_id'  => 'id',
            'shares',
            'likes',
            'comments',
            'total_count'
        ];

        /**
         * Saves a new FacebookPostData entry into the database via values from a FacebookPost object
         *
         * @param FacebookPost $facebookPost
         */
        public function saveFromFacebookPost(FacebookPost $facebookPost)
        {
            parent::saveFromParentObject($facebookPost);
        }
    }

Now, when trying to generate the phpdocblock with `php artisan ide-helper:models --dir="app/Models"` and typing in yes to overvwrite existing model files I'm getting this output

>Written new phpDocBlock to /home/vagrant/work/FBFeedParser/app/Models/BaseModel.php
Written new phpDocBlock to /home/vagrant/work/FBFeedParser/app/Models/FacebookCategory.php
Written new phpDocBlock to /home/vagrant/work/FBFeedParser/app/Models/FacebookPage.php
Written new phpDocBlock to /home/vagrant/work/FBFeedParser/app/Models/FacebookPageData.php
Written new phpDocBlock to /home/vagrant/work/FBFeedParser/app/Models/FacebookPost.php
Written new phpDocBlock to /home/vagrant/work/FBFeedParser/app/Models/FacebookPostData.php
Written new phpDocBlock to /home/vagrant/work/FBFeedParser/app/Models/User.php
Written new phpDocBlock to /home/vagrant/work/FBFeedParser/app/Models/BaseModel.php
Written new phpDocBlock to /home/vagrant/work/FBFeedParser/app/Models/FacebookCategory.php
Written new phpDocBlock to /home/vagrant/work/FBFeedParser/app/Models/FacebookPage.php
Written new phpDocBlock to /home/vagrant/work/FBFeedParser/app/Models/FacebookPageData.php
[...]

PR #242, opened 2015-08-01
fix model docBlocks duplicates

bebnev:
This commit fixes the issue #240 
The issue takes place because we set --dir option with the folder that already in app folder, so the models are parsed from app folder and then from inner folder app/Models and that is why there are duplicates. As a workaround I add every parsed models to ignore list, so they will not be parsed twice.


barryvdh:
Thanks!

Issue #239, opened 2015-07-27
get_class($this) in relation methods breaks generated model docs


hailwood:
I have some self referential relations and as such to allow the class to be extended we have

```php
/**
  * Parent relation (self-referential) 1-1.
  *
  * @return \Illuminate\Database\Eloquent\Relations\BelongsTo
  */
  public function parent() {
    return $this->belongsTo(get_class($this), $this->getParentColumnName());
  }
```
Likewise for children I have 

```php
/**
  * Children relation (self-referential) 1-N.
  *
  * @return \Illuminate\Database\Eloquent\Relations\HasMany
  */
  public function children() {
    return $this->hasMany(get_class($this), $this->getParentColumnName());
  }
```

which results in

```php
 * @property-read \get_class($this $parent 
 * @property-read \Illuminate\Database\Eloquent\Collection|\get_class($this[] $children 
```

barryvdh:
Fixed in favor of the PR

PR #243, opened 2015-08-01
fix get_class($this) in relationship functions

bebnev:
The request fixes the issue #239 

The problem was in the strpos() in exploading argument list. It was looking for the first occurance of '(' which is in get_class($this), so it takes everything before ')' --> get_class($this. While in getClassName function strpos is checking - strpos($className, 'get_class($this)') !== false, so it fails. 
I have changed it to strrpos - in order to find the last occurance and explode the full argument list.

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -336,6 +336,7 @@ class ModelsCommand extends Command
                         $code .= $file->current();
                         $file->next();
                     }
+                    $code = trim(preg_replace('/\s\s+/', '', $code));
                     $begin = strpos($code, 'function(');
                     $code = substr($code, $begin, strrpos($code, '}') - $begin + 1);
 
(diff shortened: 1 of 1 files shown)


barryvdh:
Hmm, not sure if this fixes all the edge cases. Should probably need some test or make a better parser. What if there are chained methods calls? `$this->belongsTo(get_class($this))->orderBy('id')`?
Can't you use `static::class`? Or we could check for `get_class($this` instead.

bebnev:
Yee, I have totally forgotten about chain methods. You are right. 
So the simpliest workaround in that case would be - a check for `get_class($this`

barryvdh:
Can you verify that that works and submit a new PR? (Or fix this PR and squash?)
Thanks!

bebnev:
Yes, I'll do it today.

bebnev:
Hi. When you have time, please have a look. I've changed a bit the way of parsing relations function. I've taken into consideration different preferences of writing chaining functions, so I've make one string from the function code and parse it. Also I've changed the way of parsing arguments.
I've checked this on several different tests:

```php
public function posts()
    {
        return $this
            ->hasMany('App\Models\Post')
            ->orderBy('date_to')
            ->limit(10);
    }

public function phones()
    {
        return $this->hasOne('App\Phone');
    }

public function posts2()
    {
        return $this
            ->belongsToMany(static::class, 'table1', 'id4', 'id5')->orderBy('date');
    }

public function posts3()
    {
        return $this
            ->hasMany(Post::class)->orderBy('date_to')
            ->limit(10);
    }

    public function posts4()
    {
        return $this->hasMany(get_class($this));
    }
```
And as a result I have:
```
@property-read \Illuminate\Database\Eloquent\Collection|\App\Models\Post[] $posts
@property-read \App\Phone $phones
@property-read \Illuminate\Database\Eloquent\Collection|\App\User[] $posts2
@property-read \Illuminate\Database\Eloquent\Collection|Post[] $posts3
@property-read \Illuminate\Database\Eloquent\Collection|\App\User[] $posts4 
```

barryvdh:
Thanks!

Adelf:
Hello. Seems this patch broke something.

    public function roles()
    {
        return $this->belongsToMany(Config::get('entrust.role'), Config::get('entrust.role_user_table'), 'user_id', 'role_id');
    }

Now creates:

    * @property-read \Illuminate\Database\Eloquent\Collection|\Config::get('entrust.role[] $roles 

Previous version created: 

    * @property-read \Illuminate\Database\Eloquent\Collection|\Config::get('entrust.role')[] $roles 

which was a bit better :)

bebnev:
@Adelf Hello. Which version of the ide-helper are you using ? 
I have just installed 2.1.* and the generation is okay:

I have 
```php
public function roles()
    {
        return $this->belongsToMany(Config::get('entrust.role'), Config::get('entrust.role_user_table'), 'user_id', 'role_id');
    }
```
in the beginning, and 
```php
* @property-read \Illuminate\Database\Eloquent\Collection|\Config::get('entrust.role')[] $roles
```
after generation

But in version 2.0.6 I have the problem you have described.

Adelf:
@bebnev sorry. You're right. My fault.

PR #266, opened 2015-10-02
This is only used while developing

JeyKeu:


barryvdh:
Yeah but people should then also realize they can't just add it to the list of providers, otherwise it will crash.

shehi:
Yea this change is meaningless - most of us already have thought this through.

onigoetz:
Well, you could also do the following in your `config/app.php`:
```php
if (getenv('APP_ENV') == 'local') {
    $conf['providers'][] = 'Barryvdh\LaravelIdeHelper\IdeHelperServiceProvider';
}
```

barryvdh:
You can submit a PR to the readme if you want :)

PR #273, opened 2015-10-19
Add custom collection support

klaude:
Add support for models that override Eloquent's `newCollection()` method with a custom collection type. Do this by executing the relationship on the model then executing that relationship's `newCollection()` method to see what type of collection class is returned.

This also removes the the now-unused `Barryvdh\LaravelIdeHelper\Console\ModelsCommand::getClassName()` method since relationship properties aren't determined with as much string logic as before. 

This change has one side-effect. Since it determines relationship property types by executing the relationship it now puts fully-qualified model class names into the classes' docblock tags instead of class alias names or class names without their namespace if the model and its related model are in the same namespace. For instance, what may have been rendered as:

```php
use My\Related\SingularRelationship as Singular;
use My\Related\MultipleRelationship;
use My\Related\RelationshipWithACustomCollection; // Assume this uses a \My\CustomCollection collection type.

/**
 * ...
 * @property-read Singular $singular
 * @property-read \Illuminate\Database\Eloquent\Collection|MultipleRelationship[] $multiple
 * @property-read \Illuminate\Database\Eloquent\Collection|RelationshipWithACustomCollection[] $custom
 */
```
is now:

```php
use My\Related\SingularRelationship as Singular;
use My\Related\MultipleRelationship;
use My\Related\RelationshipWithACustomCollection; // Assume this uses a \My\CustomCollection collection type.

/**
 * ...
 * @property-read \My\Related\SingularRelationship $singular
 * @property-read \Illuminate\Database\Eloquent\Collection|\My\Related\MultipleRelationship[] $multiple
 * @property-read \My\CustomCollection|\My\Related\RelationshipWithACustomCollection[] $custom
 */
```

Note how every type hint is now fully qualified and unaliased, but it uses the custom collection class. It's a little wordier, but it's still readable by IDEs. I've tested this out on some of my Laravel projects and haven't seen any other side effects. Let me know what you think!

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -367,23 +367,20 @@ class ModelsCommand extends Command
                             $code = substr($code, $pos + strlen($search));
                             $end = strpos($code, ')->') ?:strpos($code, ');');
                             if (false !== $end) {
-                                $arguments = substr($code, 0, $end);
-                                $arguments = array_map(function($item) {
-                                    return trim($item, ' \'\"');
-                                }, explode(',', $arguments));
-                                //Remove quotes, ensure 1 \ in front of the model
-                                $returnModel = $this->getClassName($arguments[0], $model);
+                                //Resolve the relation's model to a fully-qualified class name.
+                                $relatedModel = '\\' . get_class($model->$method()->getRelated());
+
                                 if ($relation === "belongsToMany" or $relation === 'hasMany' or $relation === 'morphMany' or $relation === 'morphToMany') {
                                     //Collection or array of models (because Collection is Arrayable)
                                     $this->setProperty(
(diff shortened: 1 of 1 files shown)


shehi:
Is this for NoSQL support? I really look forward to this implementation.

barryvdh:
Changed it a bit with https://github.com/barryvdh/laravel-ide-helper/commit/56f89502e38dc3a2a8253259fee5f7d592d5a03e
Removed some left-over code, checked if it's an actual model.

klaude:
@shehi, it adds support for models that return custom collections via `newCollection()`. 

http://laravel.com/docs/5.1/eloquent-collections#custom-collections

Thanks, @barryvdh!

PR #279, opened 2015-11-17
PHPDoc for Fluent methods commonly used in migrations

Stalinko:
Laravel built-in "Fluent" class is commonly used in migrations but it doesn't give any hints about it's methods. Here we hardcode it so that when you write a new migration then IDE will hint you possible methods.

### DIFF
--- a/resources/views/helper.php
+++ b/resources/views/helper.php
@@ -37,3 +37,26 @@ namespace <?= $namespace == '__root' ? '' : $namespace ?>{
 }
 
 <?php endforeach; ?>
+
+namespace Illuminate\Support {
+    /**
+    * Methods commonly used in migrations
+    *
+    * @method Fluent nullable()
+    * @method Fluent default($value)
+    * @method Fluent unsigned()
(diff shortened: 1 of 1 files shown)


Stalinko:
I think many ppl are frustrating when write migrations and have to remember all possible modifiers...
There was a pull request in laravel: https://github.com/laravel/framework/pull/8506 (related discussion https://laracasts.com/discuss/channels/laravel/phpdoc-autocomplete-for-fluent-creating-a-migration)
But they fairly declined it because Fluent class is not for migrations only. But here we are not restricted in anything. This plugin is not part of laravel. Thus we can make life of many coders easier.

Probably this feature should be optional... Let's talk about that.

jonnywilliamson:
Yes please! This is a constant source of frustration when I forget the syntax!

barryvdh:
Please make this an option (not default)

barryvdh:
Fixed via https://github.com/barryvdh/laravel-ide-helper/pull/404

PR #298, opened 2016-01-03
Allow symfony/class-loader 3.0 to be installed

onigoetz:
Hi,

I made a quick change to the composer.json in order to allow symfony/class-loader 3.0 to be installed.


barryvdh:
Thanks!

PR #303, opened 2016-01-11
Don't use force aliases loading

dkulyk:
Another autoloader throw exception if aliases not exists.


barryvdh:
Which one?

dkulyk:
Yii1 autoloader throws error if alias not set in laravel.

barryvdh:
Thanks

inkvizytor:
After this change function class_exists('Auth', false) always returns false because 'Auth' class is not loaded before line 120. In result ide-helper does not generate autocomplete for Auth's magic methods.

barryvdh:
Reverted

Issue #269, opened 2015-10-13
Adding @method docblocks on the Eloquent/Model.php


imanghafoori1:
Hi, 
I am using your great package and I see that there is no auto-completion for my models if I do not extend the \Eloquent class in the _ide_helper.php. It makes a dependency on the file which should be removed before pushing to production.

Another way of providing the auto-completion is to add the proper docblocks on the Eloquent/Model.php before the class definition.

see offical Documentation :
http://www.phpdoc.org/docs/latest/references/phpdoc/tags/method.html#examples

Here is a sample which I use for my own project.

```
use Illuminate\Database\Eloquent\Model as EloquentModel;
use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
use Illuminate\Database\Query\Builder as QueryBuilder;

/**
 * @method EloquentModel|Collection|null static $this find($id, $columns = ['*']) Find a model by its primary key.
 * @method Collection static findMany($ids, $columns = ['*']) Find a model by its primary key.
 * @method EloquentModel|Collection static findOrFail($id, $columns = ['*']) Find a model by its primary key or throw an exception.
 * @method EloquentModel|EloquentBuilder|null first($columns = ['*']) Execute the query and get the first result.
 * @method EloquentModel|EloquentBuilder firstOrFail($columns = ['*']) Execute the query and get the first result or throw an exception.
 * @method Collection|EloquentBuilder[] get($columns = ['*']) Execute the query as a "select" statement.
 * @method mixed value($column) Get a single column's value from the first result of a query.
 * @method mixed pluck($column) Get a single column's value from the first result of a query.
 * @method void chunk($count, callable $callback) Chunk the results of the query.
 * @method \Illuminate\Support\Collection lists($column, $key = null) Get an array with the values of a given column.
 * @method \Illuminate\Contracts\Pagination\LengthAwarePaginator paginate($perPage = null, $columns = ['*'], $pageName = 'page', $page = null) Paginate the given query.
 * @method \Illuminate\Contracts\Pagination\Paginator simplePaginate($perPage = null, $columns = ['*'], $pageName = 'page') Paginate the given query into a simple paginator.
 * @method int increment($column, $amount = 1, array $extra = []) Increment a column's value by a given amount.
 * @method int decrement($column, $amount = 1, array $extra = []) Decrement a column's value by a given amount.
 * @method void onDelete(Closure $callback) Register a replacement for the default delete function.
[...]

Issue #259, opened 2015-08-25
Should ide-helper:models be generating phpDoc for the magic Eloquent\Model static methods?


jessedc:
I have recently setup a new Lumen project (v5.1.3) and noticed the instance methods from `Eloquent\Builder` provided to `Eloquent\Model` by using `Model::__callStatic()` and `Model->__call()` internally are not added to the phpDocs generated by `ide-helper:models`.

One example directly from  [the Larvel 5.1 documentation](http://laravel.com/docs/5.1/eloquent#eloquent-model-conventions) is the `::find()` method.

<img width="398" alt="screen shot 2015-08-26 at 6 50 18 am" src="https://cloud.githubusercontent.com/assets/17407/9479459/e09915e6-4bc0-11e5-9b45-786a0aaa9289.png">

My observation is this call to ::find() is caught by `__callStatic()` and `__call()` methods of Eloquent\Model and proxied to `\Illuminate\Database\Eloquent\Builder`.

<img width="728" alt="screen shot 2015-08-25 at 9 51 58 pm" src="https://cloud.githubusercontent.com/assets/17407/9480110/74b2a636-4bc4-11e5-888e-c7efeb6be163.png">

These methods can be added to the phpDocs of the model with @method hints like the following: 

``` php
/**
* @method static \Illuminate\Database\Eloquent\Model|\App\Models\Cafe find($id)
*/
```

My question is, **should these magic methods from Builder be added to Model subclasses by `ide-helper:models`**?  Or should the Laravel documentation be updated instead?

Worth noting that this whole situation can also be avoided by using the builder directly via the `query()` method.

``` php
$cafe = Cafe::query()->find($cafeId);
```

> The [find() method was defined in 5.0](http://laravel.com/api/5.0/Illuminate/Database/Eloquent/Model.html#method_find) but are no longer defined on the `Eloquent\Model` as of 5.1.

rentalhost:
I'll works if you extends from `Eloquent` facade. But I think that it need to works directly from `Eloquent\Model` too.

bebnev:
I think that having such methods in Model docblocks is a real overhead, because find is only one "magic" method from EloquentBuilder. Then somebody will want methods from the QueryBuilder which are also called with __call/__callStatic.

[...]

Issue #248, opened 2015-08-07
Model::findOrFail


miladrahimi:
It doesn't support this method (findOrFail) and some other methods like Model::save().

I'm using Laravel 5.0 and "barryvdh/laravel-ide-helper": "^2.0".

If there is a simple way to add it myself please explain.

Radau:
As far as findOrFail goes I've had luck using `class CLASSNAME extends \Eloquent` instead of `class CLASSNAME extends Model` within your Model. This workaround will cause it to use the helper file, so I believe you will need to deploy WITH the ide-helper file, but I'm not certain.

That will get most of them working it seems, however I'm still having issues getting `where` to work within my Model, not sure if it has to do with how I'm calling it `$query->where(...`, but using \Eloquent seems to fix most of the methods.

miladrahimi:
Definitely I don't want use \Eloquent instead of Model cause it's the Laravel recommendation,
But It's seems it's not hard to extend this helper. how can I do it?

barryvdh:
It's the same thing, Eloquent is an alias of Model. But because it's an alias, we can use the ide helper for it.

miladrahimi:
So if I use Eloquent (Laravel >= 5), It supports more. ok.

pahan35:
I faced with similar problem today.

I fixed it by adding `|static[]|static|null` to return signature for class `class Eloquent extends \Illuminate\Database\Eloquent\Model` method ` public static function findOrFail($id, $columns = array())` in **_ide_helper.php**

Before
```
            /**
             * Find a model by its primary key or throw an exception.
             *
             * @param mixed $id
             * @param array $columns
             * @return \Illuminate\Database\Eloquent\Model|\Illuminate\Database\Eloquent\Collection
             * @throws \Illuminate\Database\Eloquent\ModelNotFoundException
             * @static 
             */ 
            public static function findOrFail($id, $columns = array())
            {    
                return \Illuminate\Database\Eloquent\Builder::findOrFail($id, $columns);
            }

```

After
```
            /**
             * Find a model by its primary key or throw an exception.
             *
             * @param mixed $id
             * @param array $columns
             * @return \Illuminate\Database\Eloquent\Model|\Illuminate\Database\Eloquent\Collection|static[]|static|null
             * @throws \Illuminate\Database\Eloquent\ModelNotFoundException
             * @static 
             */ 
            public static function findOrFail($id, $columns = array())
            {    
[...]

Issue #209, opened 2015-05-11
The columns of eloquent db objects are not recognized in phpstorm


iarief:
Hi, i'm having quite a peculiar problem here, the ide-helper:model did generate the new file called _ide_helper_models.php but phpstorm doesn't seem to read from those. 

i.e.

$article = Article::find(1);
$article->(Ctrl + space)

in the second i believe that i should expect the columns of the object Articles to exist in the autocomplete suggestion list, but it doesn't. i had some column on that class (body, title, published_at) and they exist in _ide_helper_models.php file but for some reason it's still not working. the Article::find did get auto completed though.

Komayo:
As i know, it happens also with other situations. for example.

$user = User::where(xyz...), where method is also not detected.

For example also, in $message parameter of a Mail::send method,

$message->to, to is also not reconized!

And theres a few more.

I did some tests, and i know that for example at Sublime Text 3, the autocomplete works 100% fine, even with this methods.

Just happens in Phpstorm...

funsterisdafunniest:
@Komayo but did sublime 3 able to provide the auto completion for columns in the objects?
i already extend the Eloquent facade and created the model helper but still can't get it to work.

Komayo:
@funsterisdafunniest yes it provides auto completion for columns.

for example my user table has "username, password, activated, etc"

$user->activated (perfectly fine with autocomplete)

funsterisdafunniest:
@Komayo  hmm. that's strange.. it doesn't work for me even while i'm using sublime with code intel. did you write the php doc in the object or did you write them to the external file?

Komayo:
I am using the generated ide_helper.php. Works fine. But not on phpstorm.

aledmb:
@Komayo how do you deal with "where" and a few other methods not passing inspection? do you suppress those warnings on all your files?

akmumu:
so do i
![image](https://cloud.githubusercontent.com/assets/3242526/10018173/318aea04-6167-11e5-89b5-94b9d6d38bd0.png)

ruuter:
IMHO the best workaround I've found so far is to downgrade severity if __magic methods are present in class. PhpStorm has this exact option in inspections settings. [Check my answer at SO](http://stackoverflow.com/a/35347256/1296104)

zuhairtaha:
I faced this problem. Finally I solved it by following the structures under **Automatic phpDocs for models** in laravel-ide-helper.

[...]

PR #305, opened 2016-01-12
Add extend \Eloquent for ide-helper:models command

matiaspub:
It's help author to close many issues! )
Like: #269, #259, #248, #209, #191, #74, 
and https://github.com/laravel/framework/issues/7558 )))
etc. maybe

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -484,6 +484,10 @@ class ModelsCommand extends Command
             $phpdoc->appendTag($tag);
         }
 
+        if ($this->write) {
+            $phpdoc->appendTag(Tag::createInstance("@mixin \\Eloquent", $phpdoc));
+        }
+
         $serializer = new DocBlockSerializer();
         $serializer->getDocComment($phpdoc);
         $docComment = $serializer->getDocComment($phpdoc);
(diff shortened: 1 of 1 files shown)


barryvdh:
As I said before, shouldn't we `mixin`  the actual classes?

matiaspub:
Sorry, I didn't find it.
But Why?

barryvdh:
Now we're linking the methods to the ide_helper file. It would be nicer if PHPstorm would actually click directly to the correct class.

barryvdh:
Which are `Illuminate\Database\Eloquent\Builder`  and `Illuminate\Database\Query\Builder`

matiaspub:
It would be nicer... Right!
But it could be a temporary solution.
Or we will wait until PhpStorm begins to understand @mixin static ?

Lambik:
When I add the `mixin` to my model tag, phpstorm understands `::where()`, but subsequently complains that extra chained `->where()` and `->take()` and whatnot should be static too (`::where`, `::take`). Too bad that this is so hard to solve :(

barryvdh:
I tested this. You are correct about PHPstorm not understanding the statics. Merged this. Thanks!

PR #315, opened 2016-01-28
Ignore abstract class or interface, not throw exception

icedfish:


barryvdh:
Isn't the exception caught anyways?

icedfish:
it was catched and show red warning on user console.

But I don't think it need to show, every project has abstract class and interface, so every time I run the command will see the same warnings, but I can do nothing.

I think the warning is meanless, so just skip it will be more appropriate.

Issue #274, opened 2015-10-22
Auth::user resolves to wrong type


tkuijer:
In my project I have moved the normal `App\User` to `Domain\Models\User` (Residing in the app\Models folder). I also modified my auth configuration file to reflect this change. However the `ide-helper:generate` command doesn't pick up on this change and instead points the `Auth::user()` to the standard App\User model

shehi:
See readme file in repo, read it completely. You are wasting people's time.

barryvdh:
@shehi Relax, nothing wrong with asking a question.

See the config file here: 
https://github.com/barryvdh/laravel-ide-helper/blob/master/config/ide-helper.php#L86

It should replace `\Illuminate\Contracts\Auth\Authenticatable` with `config('auth.model', 'App\User'),`.
If `auth.model` is set, it should use that. But perhaps you have an old published config?

You are talking about the _ide_helper.php file, right?

shehi:
@barryvdh : I never needed all that - just give ```--dir=``` parameter and provide model's FQN, and any model should be mapped. I have Cartalyst's Sentinel running the show in my app, and it needs changing nothing at all, to get doc generation.

tkuijer:
I did find the option in the ide-helper configuration file (after posting
this issue, my fault..) however this should default to looking at the auth
config file, which doesn't appear to be working, so it uses the default
(which I now modified to reflect my project)
Op 22 okt. 2015 12:38 schreef "Şəhriyar İmanov" <notifications@github.com>:

barryvdh:
Could be due to the order of loading the config (or service provider) and/or caching the configs.

tkuijer:
This seems to match my findings while debugging. Looks like the `auth.model` configuration key isn't set at the time it gets requested by the ide-helper config file.

barryvdh:
I guess an option would be to leave our config 'null' and request the Auth model on runtime, then we're sure it's available.

tkuijer:
Sounds like that would be a good solution. Rather than configuring the model in the `ide-helper` configuration file, request it when we're sure it's available.

barryvdh:
Did that in above commit. Doesn't work for people who published their configs though.

rickselby:
For 5.2, this should be auth.providers.users.model. Not sure you can check the version in the config file though?

PR #324, opened 2016-02-28
Fix default user model for Laravel 5.2

rickselby:
Mentioned in #274. Not sure how this would work for lumen though.


barryvdh:
Simplified by checking both config values, using old one as fallback. See above commit. thanks!

Issue #312, opened 2016-01-25
Support camel case Model properties


valorin:
There are some useful packages, such as https://github.com/kirkbushell/eloquence, that allow you to access Model properties using camelCase. This allows you to avoid snake_case properties and generally keep your code cleaner.

It would be awesome if there was a way to tell the `ide-helper:models` command to write properties in camelCase instead of snake_case. It'd be cool if it could be automatic, but I'm happy to toggle a config option - since it's generally an application-wide option.

For example, instead of a docblock looking like this:

```
/**
 * @property int $id
 * @property int $team_id
 * @property string $name
 * @property string $slug
 * @property string $colour
 * @property \Carbon\Carbon $created_at
 * @property \Carbon\Carbon $updated_at
 */
```

It would look like this:

```
/**
 * @property int $id
 * @property int $teamId
 * @property string $name
 * @property string $slug
 * @property string $colour
 * @property \Carbon\Carbon $createdAt
 * @property \Carbon\Carbon $updatedAt
 */
```

PR #326, opened 2016-03-03
Allow for camel case model properties.

valorin:
Added support for camel case model properties, as I had previously requested in #312.

There are some Laravel packages (such as Eloquence) that allow for accessing Eloquent model properties via camel case, instead of snake case.

This PR adds in an option (default disabled), to support these packages by saving all model properties as camel case, instead of snake case.

For example, normally you would see this:

```
 * @property \Carbon\Carbon $created_at
 * @property \Carbon\Carbon $updated_at
```

With this enabled, the properties will be this:

```
 * @property \Carbon\Carbon $createdAt
 * @property \Carbon\Carbon $updatedAt
```

It is currently an all-or-nothing option, as that was the simplest to implement and it suits my needs.

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -116,4 +116,30 @@ return array(
 
     ),
 
+    /*
+     |--------------------------------------------------------------------------
+     | Support for camel cased models
+     |--------------------------------------------------------------------------
+     |
+     | There are some Laravel packages (such as Eloquence) that allow for accessing
+     | Eloquent model properties via camel case, instead of snake case.
+     |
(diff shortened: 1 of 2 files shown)


barryvdh:
Thanks!

HenkPoley:
Slight helpful comment for who it concerns, you might need to run `php artisan config:clear` after copying `vendor/barryvdh/laravel-ide-helper/config/ide-helper.php` to `./config/ide-helper.php` and editing `'model_camel_case_properties'` to true.

PR #338, opened 2016-03-29
Look for additional facade aliases in app config array

cnanney:
In my Lumen app, I load my providers and aliases from a `config/app.php` file like in Laravel. Maybe that is uncommon, but this simple PR will include those aliases as well if any are present.


kt81:
👍 
I really hope this feature will be merged.

Issue #352, opened 2016-06-07
Error when installing


showFocus:
I'm using a frsh install of Laravel 5.2.36. When I run this command: 

> composer require barryvdh/laravel-ide-helper

I get the following error:

> Your requirements could not be resolved to an installable set of packages.
>   Problem 1
>     - Installation request for phpdocumentor/reflection-docblock (locked at 3.0.2) -> satisfiable by phpdocumentor/reflection-docblock[3.0.2].
>     - barryvdh/laravel-ide-helper v2.1.0 requires phpdocumentor/reflection-docblock 2.0.4 -> satisfiable by phpdocumentor/reflection-docblock[2.0.4].
>     - barryvdh/laravel-ide-helper v2.1.1 requires phpdocumentor/reflection-docblock 2.0.4 -> satisfiable by phpdocumentor/reflection-docblock[2.0.4].
>     - barryvdh/laravel-ide-helper v2.1.2 requires phpdocumentor/reflection-docblock 2.0.4 -> satisfiable by phpdocumentor/reflection-docblock[2.0.4].
>     - barryvdh/laravel-ide-helper v2.1.3 requires phpdocumentor/reflection-docblock ^2.0.4 -> satisfiable by phpdocumentor/reflection-docblock[2.0.4].
>     - barryvdh/laravel-ide-helper v2.1.4 requires phpdocumentor/reflection-docblock ^2.0.4 -> satisfiable by phpdocumentor/reflection-docblock[2.0.4].
>     - Conclusion: don't install phpdocumentor/reflection-docblock 2.0.4
>     - Installation request for barryvdh/laravel-ide-helper ^2.1 -> satisfiable by barryvdh/laravel-ide-helper[v2.1.0, v2.1.1, v2.1.2, v2.1.3, v2.1.4].

nueko:
I have no issue when installing fresh 5.2, but 5.1 yes.
i think the package composer.json should be 
```json
"phpdocumentor/reflection-docblock": "^2.0|^3.0"
```

if you currious, just hit ```composer why phpdocumentor/reflection-docblock```

to fix the issue, i first remove phpdocumentor/reflection-docblock
```
composer rem phpdocumentor/reflection-docblock:^3.0
```
then install the phpdocumentor/reflection-docblock:^2.0
```
composer req phpdocumentor/reflection-docblock:^2.0
```

then install this package

```
composer req barryvdh/laravel-ide-helper
```
hope it helps.

[...]

PR #353, opened 2016-06-07
fix reflection-docblock issue #352

nueko:


barryvdh:
Did you check that the generated file is still correctly parsed?

barryvdh:
I added some tests an triggered a build with close/open. But as you can see, it fails the tests.

Some methods are not working in v3. You can see that the lower version is still working.

barryvdh:
Used my fork of v2 for now, to avoid conflicts.

PR #361, opened 2016-06-29
Use FQN for app() methods

### LINKED ISSUE #340: .phpstorm.meta.php has no effect in PhpStorm 2016.1
@barryvdh , after a month or so, when I return to my project, for some weird reason, the generated ```.phpstorm.meta.php``` file is missing a lot of lines in it (such as lines for ```app('log')```, ```app('validator')``` etc). My current file looks like this:

https://gist.github.com/shehi/1aeb5d96c5c1b338360474efbd7806f0

Has something changed with package?

cnanney:
This fixes completion in PhpStorm 2016.2+, and also works in PhpStorm 10. Note: despite needing the FQN in the .meta definition, you do not need to use it in your project to get autocomplete.

FQN is [now required for function references](https://youtrack.jetbrains.com/issue/WI-31102), closes #340.

Unfortunately, this does not fix the problem in 2016.1, as you still need to use the FQN `\app('foo')->` in your project to get autocomplete, but there's nothing we can do about that - it's a PhpStorm bug, and is now fixed in 2016.2 EAP.

### DIFF
--- a/src/Console/MetaCommand.php
+++ b/src/Console/MetaCommand.php
@@ -47,7 +47,7 @@ class MetaCommand extends Command
       'new \Illuminate\Contracts\Container\Container',
       '\Illuminate\Contracts\Container\Container::make(\'\')',
       '\App::make(\'\')',
-      'app(\'\')',
+      '\app(\'\')',
     ];
 
     /**


shehi:
Are you 100% sure and tested this @cnanney ? Don't really want to re-open this issue.

shehi:
@barryvdh , let's merge and close the related issue only after final release please. We shouldn't be basing our resolutions on EAPs, I think.

barryvdh:
@shehi did you verify this fixes it?

shehi:
@barryvdh , no, but I never test EAPs, will do it the moment final version is out.

barryvdh:
Reading the comments on the issue tracker,  this seems to be correct. Thanks!

cnanney:
I've tested the different .meta variants on PS 10, 2016.1, and 2016.2 EAP. This PR has no effect on 10, 2016.1 will remain broken, but will fix completion from 2016.2 forward is it now requires FQN.

PR #377, opened 2016-08-19
Allow Type Overrides

ellisio:
This PR is to allow a developer to define the following in their `ide-helper.php` configuration:

```php
    /*
    |--------------------------------------------------------------------------
    | Property Casts
    |--------------------------------------------------------------------------
    |
    | Cast the given "real type" to the given "type".
    |
    */
   'type_overrides' => array(
       'integer' => 'int',
       'boolean' => 'bool',
   ),
```

This will replace `@property integer` with `@property int` when running `ide-helper:models`. This will satisfy the [phpdoc_scalar](https://styleci.readme.io/docs/fixers#phpdoc_scalar) fixer when using the Laravel preset on StyleCI.

This is also the scalar convention for Laravel and I would actually honestly consider changing `integer` to `int` and `boolean` to `bool` in the following locations:

- https://github.com/barryvdh/laravel-ide-helper/blob/master/src/Console/ModelsCommand.php#L260
- https://github.com/barryvdh/laravel-ide-helper/blob/master/src/Console/ModelsCommand.php#L332
- https://github.com/barryvdh/laravel-ide-helper/blob/master/src/Console/ModelsCommand.php#L245
- https://github.com/barryvdh/laravel-ide-helper/blob/master/src/Console/ModelsCommand.php#L339

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -153,6 +153,5 @@ return array(
    'type_overrides' => array(
        // 'integer' => 'int',
        // 'boolean' => 'bool',
-   )
-
+   ),
 );
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -1,4 +1,4 @@
-<?php 
+<?php
 
 return array(
 
@@ -71,7 +71,7 @@ return array(
             'emergency' => 'Monolog\Logger::addEmergency',
         )
     ),
-    
+
     /*
     |--------------------------------------------------------------------------
     | Interface implementations
@@ -81,7 +81,7 @@ return array(
     | are detected by the helpers, others can be listed below.
     |
     */
-    
+
     'interfaces' => array(
 
     ),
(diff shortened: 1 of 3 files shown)


michaeljhopkins:
+1+1+1

Would relieve a lot of my headachest

barryvdh:
Do we need this a a config? Why not just always change them?

ellisio:
That's why I have the recommendation for the change spots in the OP. Giving the author the option on which one they want to use.

barryvdh:
Can you set them by default so that they make sense with setting the config?

config/ide-helper.php
```diff
@@ -142,4 +142,16 @@
...
+    |--------------------------------------------------------------------------
+    | Property Casts
+    |--------------------------------------------------------------------------
+    |
+    | Cast the given "real type" to the given "type".
+    |
+    */
+   'type_overrides' => array(
+       // 'integer' => 'int',
+       // 'boolean' => 'bool',
```
barryvdh: Uncomment and add good defaults please :)

ellisio:
Uncommented the defaults as requested.

barryvdh:
Thanks!

PR #383, opened 2016-08-26
Laravel 5.3  - Update composer.json

fravel77:
bump symfony/class-loader to at least 3.1 in order to use this package with Laravel 5.3


barryvdh:
Hmm this doesn't seem right. 3.1 is already supported with ^3.0

Issue #170, opened 2015-02-18
Mutator not showing up


Anahkiasen:
Take the following model:

``` php
class MyModel extends Model
{
  /**
   * @return string[]
   */
  public function getWordsAttribute()
  {
    return json_decode($this->attributes['words']);
  }
}
```

The IDE helper will currently generate `@property string $words` because the database column is a `text`.
Other mutators get their documentation added properly, but not the ones that have a column of the same name in the database it seems

barryvdh:
Are you sure other mutators are correct? I think I should add a reflection method to the attribute mutator to get it. Don't do that now I think:
https://github.com/barryvdh/laravel-ide-helper/blob/master/src/Console/ModelsCommand.php#L286

Anahkiasen:
They're more or less correct, most just display `mixed` as type instead of what the mutator actually returns, but they're there

Kurre:
Just wanted to bump this issue, since I've been "struggling" with same problems.

I have mutator which return Money object like this:
(slightly simplified example)
```php
/**
 * Ticket price
 *
 * @param $value
 * @return Money
 */
public function getPriceAttribute($value)
{
    if ($value !== null) {
        return Money::EUR($value);
    }

    return Money::EUR(0);
}
```
But since ModelsCommand doesn't look the docblocks for a return value, it's using the column type (like it should be in normal situation):
```php
/**
 * @property int $price
 */
```

My quick'n'dirty solution is this:
(https://github.com/barryvdh/laravel-ide-helper/blob/master/src/Console/ModelsCommand.php#L386-L390)
```php
//Magic get<name>Attribute
$name = Str::snake(substr($method, 3, -9));
if (!empty($name)) {
    $reflection = new \ReflectionMethod($model, $method);
    $phpdoc = new DocBlock($reflection);
    $type = null;

    if ($phpdoc->hasTag('return')) {
        $returnTags = $phpdoc->getTagsByName('return');
        $type = $returnTags[0]->getContent();
    }
    
    $this->setProperty($name, $type, true, null);
}
```

Not very elegant, but I works :)
Now I get:
```php
/**
 * @property Money $price
 */
```

So, i'm hoping that you, @barryvdh could take a another look on this issue when you have time.
This isn't really a big problem, but I would help to get these properties sorted out :)

pinepain:
@Kurre, I guess you mean something like this - https://github.com/barryvdh/laravel-ide-helper/pull/389.

[...]

PR #389, opened 2016-09-06
Use accessor docblock return type to guess property type

pinepain:
This change allow to use return type from accessor's dockblock to specify parameter type in generated class-level docblock.

For example, for this class
```php
class Test extends Model
{
    /**
     * Test getter
     *
     * @return string
     */
    public function getTestAttribute()
    {
        return 'test';
    }
}
```
generated docblock will look like 

```php
/**
 * Test
 *
 * @property-readonly string $test
 * @mixin \Eloquent
 */
```

This small change will lead to better code completion and should improve developer's experience for those who have plenty of custom getters/setters.

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -386,7 +386,9 @@ class ModelsCommand extends Command
                     //Magic get<name>Attribute
                     $name = Str::snake(substr($method, 3, -9));
                     if (!empty($name)) {
-                        $this->setProperty($name, null, true, null);
+                        $reflection = new \ReflectionMethod($model, $method);
+                        $type = $this->getReturnTypeFromDocBlock($reflection);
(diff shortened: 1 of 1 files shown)


pinepain:
I have in my branch some ugly fix for laravel-4.2, so if one need that, I can PR too.

Kurre:
This is great, been waiting for this :)

Some time ago I bumped this old issue https://github.com/barryvdh/laravel-ide-helper/issues/170 which can be closed if/when this PR is merged.

Thank you for putting time on this one :+1:

barryvdh:
THanks!

miklcct:
Can you also try to guess the return type using PHP 7.0 return type declaration?

barryvdh:
Not sure, perhaps with an upgraded reflector library

PR #394, opened 2016-10-06
add autocomplete for Facade classes

leo108:


barryvdh:
This will cause double implementations, right?

leo108:
Yes, there will be a warning about class multiple definition in IDE

leo108:
@barryvdh Do you have a better solution for this ?

zedentox:
Multiple class definitions warnings are not a big problem because _ide_helper_models.php already generate this kind of warning. So, if someone use php artisan ide-helper:models he already have to remove these warnings. (I use PHPStorm and I can remove them in the Inspections Menu)

In counterpart that allows to use aliases AND full paths of facades which is good for integrating ide-helper in an existing project. You dont need to make a big replace of all full paths everywhere and maybe making mistakes...

barryvdh:
I am kind of interested in this, but wouldn't it make more sense to generate something like this:

```php
namespace {
    exit("This file should not be included, only analyzed by your IDE");

    class App extends \Illuminate\Support\Facades\App {}
}


namespace Illuminate\Support\Facades {

    class App {
        ...
    }
}
```

So that we add the functions in the namespaced facade, and for the shortname, just extend it.


Also, Taylor did accept a PR for adding mixing to the actual class: https://github.com/laravel/framework/pull/16341

Wouldn't that be better?

zedentox:
I dont know about `@mixin`. Is it not too PHPStorm specific ?

I think your way is cleaner too.  I tried to modify my `_ide-helper.php` this way. We should not forget that some facades are not in namespace `Illuminate\Support\Facades` so we need multiple namespace declarations.

zedentox:
After some tests I found that if you use `@mixin` PHPStorm uses classes from the framework instead of classes in `_ide-helper.php`... So return type of `Auth::user()` is not your User model but Authenticatable.

@barryvdh I think I have a good generator. Do you think I should make another PR ? Inside another branch maybe ? I dont use Github usually... :)

zedentox:
You should add a contribution guide 😃

barryvdh:
The other PR is now tagged, so closing this. Let me know if something is not fully working!
https://github.com/barryvdh/laravel-ide-helper/releases/tag/v2.3.0

Issue #43, opened 2013-11-17
[Request] Support overriding methods


hailwood:
I want to be able to specify what model in particular something like Auth::user() returns, to do this I was hoping I could just add a couple of lines to the config file:

```
'Auth' => array(
    'user' => 'User::__construct'
)
```

This kind of works, in that the method is generated successfully, however this means we now have two user methods in the helper under auth instead of just the more explicit one.

gonzalom:
Hi,

I've been researching...

For **Laravel version 5.2** and **barryvdh/laravel-ide-helper version 2.2.2**

The **Auth Facade** is defined here:
https://laravel.com/api/5.2/Illuminate/Auth/AuthServiceProvider.html
https://github.com/illuminate/auth/blob/5.2/AuthServiceProvider.php

But it also uses the guard

In the generator, the **Auth driver ide helper** is defined here:
https://github.com/barryvdh/laravel-ide-helper/blob/v2.2.2/src/Generator.php#L123

As I have more than one guard, I don't have a default one.

In my generator, by default, it only include the class "\Illuminate\Auth\AuthManager" methods.

So I had to implement all of them, and this is how I did it:

For the **interfaces**, we have to do it with the methods, one by one:

```php
<?php
return [
    'magic' => array(
        ),
        /**
         * @see https://laravel.com/docs/5.2/authentication
         * @see https://laravel.com/api/5.2/Illuminate/Auth.html
         * @see \Illuminate\Auth\AuthServiceProvider
         * @see \Illuminate\Auth\AuthManager::__call
         * @see \Illuminate\Contracts\Auth\Guard
         * @see \Illuminate\Contracts\Auth\StatefulGuard
         */
        'Auth' => [
            'check'     => 'Illuminate\Contracts\Auth\Guard::check',
            'guest'     => 'Illuminate\Contracts\Auth\Guard::guest',
            'user'      => 'Illuminate\Contracts\Auth\Guard::user',
            'id'        => 'Illuminate\Contracts\Auth\Guard::id',
            'validate'  => 'Illuminate\Contracts\Auth\Guard::validate',
            'setUser'   => 'Illuminate\Contracts\Auth\Guard::setUser',
            // 'attempt'       => 'Illuminate\Contracts\Auth\StatefulGuard::attempt',
            // 'once'          => 'Illuminate\Contracts\Auth\StatefulGuard::once',
            // 'login'         => 'Illuminate\Contracts\Auth\StatefulGuard::login',
[...]

PR #418, opened 2016-12-17
Multi auth guards and multi auth providers compatibility

gonzalom:
As I had complex web apps where website (session auth driver) and API (token auth driver) live together, and also could have users which could login in more than one table (so more than one model), I felt the need for this commit. I hope you like it.

This commit is a consequence of my comment:
https://github.com/barryvdh/laravel-ide-helper/issues/43#issuecomment-267744350

It could be improved making it more flexible, auto generating the interfaces and managing methods match conflicts somehow with reflection classes or functions as 'class_implements' and 'get_class_methods', so as ''Illuminate\Contracts\Auth\StatefulGuard' interface is an extension of 'Illuminate\Contracts\Auth\Guard' interface, then the method should be the origin, so 'Illuminate\Contracts\Auth\Guard'. But I don't think it worth it, as the supported auth drivers are just 'session' or 'token'.

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -115,10 +115,56 @@ class Generator
         ), $flags);
     }
 
+    /**
+     * The auth settings are defined in config/auth.php
+     *
+     * \Illuminate\Auth\AuthManager already added by default
+     * @see \Illuminate\Auth\AuthServiceProvider::registerAuthenticator()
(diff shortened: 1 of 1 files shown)


gonzalom:
I'll send it again with the format fixes...

PR #419, opened 2016-12-17
Multi auth guards and multi auth providers compatibility

gonzalom:
As I had complex web apps where website (session auth driver) and API (token auth driver) live together, and also could have users which could login in more than one table (so more than one model), I felt the need for this commit. I hope you like it.

This commit is a consequence of my comment:
#43 (comment)

It could be improved making it more flexible, auto generating the interfaces and managing methods match conflicts somehow with reflection classes or functions as 'class_implements' and 'get_class_methods', so as ''Illuminate\Contracts\Auth\StatefulGuard' interface is an extension of 'Illuminate\Contracts\Auth\Guard' interface, then the method should be the origin, so 'Illuminate\Contracts\Auth\Guard'. But I don't think it worth it, as the supported auth drivers are just 'session' or 'token'.

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -115,10 +115,56 @@ class Generator
         ), $flags);
     }
 
+    /**
+     * The auth settings are defined in config/auth.php
+     *
+     * \Illuminate\Auth\AuthManager already added by default
+     * @see \Illuminate\Auth\AuthServiceProvider::registerAuthenticator()
+     * @see \Illuminate\Auth\AuthManager
+     *
+     * \Illuminate\Auth\RequestGuard is not a supported guard driver.
+     * @see \Illuminate\Auth\AuthManager::viaRequest()
+     * @see \Illuminate\Auth\RequestGuard
+     *
+     * Supported: "session", "token"
+     * @see \Illuminate\Auth\AuthManager::resolve()
+     * @see \Illuminate\Auth\AuthManager::createSessionDriver()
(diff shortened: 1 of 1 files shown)


gonzalom:
@barryvdh , did you reviewed this PR?

mfn:
@gonzalom do you still need this PR?

gonzalom:
Hi @mfn,

Thanks for your reply.

No, I don't need it anymore. It could be good to have for others though, I made it very standard, but I let you evaluate the benefits.

Regards.

mfn:
The reason I asked is that the implementation is hard to read, requires time to properly test and has zero tests (granted: back then there were almost no tests), so since OP suggested he doesn't need it I would err to drop this PR @barryvdh ; anyone can pick this up and make a new one anyway!

gonzalom:
Yeah, you are right, my bad about the tests, and it is hard to read indeed.

And on top of it, it was designed for a Larave\Lumen version that it's outdated already.

I am working in another company now, and our app needs are different, so I am not familiar with the latest Laravel auth particularities anymore.

I guess that you can close this, and if someone needs something similar to be implemented in the future, who knows, it may be useful as a reference/inspiration.

Thanks though for your follow up, really appreciated :)

Either way, let me know if you want some help in any particular development or update, I would be happy to help.

mfn:
Closing as per last comment, thank you!

PR #437, opened 2017-01-31
Add facades fullpath autocomplete

zedentox:
After PR #394 discussion I decided to make my own version in another way.

Facades with there methods are now declared in the same namespace as in Laravel. So you can use fullname of facades to import them : `use Illuminate\Support\Facades\Auth;` This is better for using ide-helper on an old project made without ide-helper. You don't need to replace fullnames anymore.

I also noted that autocompletion is improved for some librairies like Intervention : `Image::make()` result is now well autocompleted. It's surely due to the fact that now PHPStorm use Laravel docBlocks in association with the docBlocks in `_ide-helper.php`. For example, `Auth::user()` dont only return your user model but also `Authenticatable` like the original method.

To allow usage of short names, facades are also declared in the global namespace and extends the fullname of the facade. (\App extends \Illuminate\Support\Facades\App)

### DIFF
--- a/resources/views/helper.php
+++ b/resources/views/helper.php
@@ -9,14 +9,10 @@
  */
 
 <?php foreach($namespaces as $namespace => $aliases): ?>
-namespace <?= $namespace == '__root' ? '' : $namespace ?>{
-<?php if($namespace == '__root'): ?>
-    exit("This file should not be included, only analyzed by your IDE");
-<?= $helpers ?>
-<?php endif; ?>
+namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?>{
 <?php foreach($aliases as $alias): ?>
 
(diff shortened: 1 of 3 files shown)


barryvdh:
Thanks will give it a try :)

zedentox:
Ooops ! I found something broken with some aliases where alias name is not equal to the name of the class really used. (ex with alias Eloquent which use class Model)

Actually methods find, findMany, etc are not found in autocompletion due to this bug.

I have a fix. I pushed it to the same branch in my repo. Do I need to do another PR ?

Thanks!

barryvdh:
Yes create a new PR please

barryvdh:
Tagged! https://github.com/barryvdh/laravel-ide-helper/releases/tag/v2.3.0

PR #438, opened 2017-02-02
Readme [Lumen]: Simpler solution for Additional Facades and note on getting Eloquent mixin to work.

MitchellMcKenna:
- Currently docs suggest to create `config/app.php` just for aliases (but to renaming it something else) and having to rename it back every time you need to re-generate your ide-helper files. Instead suggest copying the config/app.php from vendor folder (just like you would do to override any other values in this config) and adding your aliases to it and leaving the file there.
- Adds note on adding `Eloquent` to `aliases` array so that `@mixin \Eloquent` works for autocompletion.
- Updates code sample to show adding aliases array to bottom of copied config/app.php instead.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -207,21 +207,30 @@ if ($app->environment() !== 'production') {
 
 After that, Laravel IDE Helper should work correctly. During the generation process, the script may throw exceptions saying that some Class(s) doesn't exist or there are some undefined indexes. This is normal, as Lumen has some default packages stripped away, like Cookies, Storage and Session. If you plan to add these packages, you will have to add them manually and create additional Facades if needed.
 
-#### Adding Additional Facades
+#### Adding the App Config
 
-Currently Lumen IDE Helper doesn't take into account additional Facades created under `bootstrap/app.php` using `create_alias()`, so you need to create a `config/app.php` file and add your custom aliases under an `aliases` array again, like so:
+In order to get autocompletion for additional Facades and for certain magic Model methods you need to copy `/vendor/laravel/lumen-framework/config/app.php` to `config/app.php`.
+
+##### Adding Additional Facades
+
+Currently Lumen IDE Helper doesn't take into account additional Facades created under `bootstrap/app.php` using `create_alias()`, so you'll have to add your custom aliases under an `aliases` array in `config/app.php` (see example below).
+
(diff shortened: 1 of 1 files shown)


MitchellMcKenna:
We might even be able to provide a `vendor:publish` command to copy the config/app.php from the vendor directory automatically, let me know if you'd like me to update the pull request with that.

DarkGhostHunter:
Currently, Lumen's `config/app.php` doesn't include any kind of array of Facades, so even by copying that file you won't get any Facade not already _aliased_.

I updated my pull for that particular reason also.

mfn:
@MitchellMcKenna do you feel this PR is still relevant?

MitchellMcKenna:
@mfn Yes, just tested out again with latest lumen version. 

I was only able to get `\Eloquent` autocomplete to work, plus I had to update the call `$app->withFacades()` call in `bootstrap/app.php` with the alias as well:
```
$app->withFacades(true, [
    Illuminate\Database\Eloquent\Model::class => 'Eloquent',
]);
```
Not sure why that step is necessary as well.

Like @DarkGhostHunter said, Lumen doesn't use illuminate AliasLoader so adding aliases in `config/app.php` won't create the aliases for the app, just allow ide-helper to know about them when ide-helper pulls them here: https://github.com/barryvdh/laravel-ide-helper/blob/6c8e8474305fc99acdb3c704e1f7ec5ff3ff665c/src/Generator.php#L273. This means you have to add your aliases in both places; in $app->withFacades() call for lumen to register them and in `config/app.php` for ide-helper to set them up for autocompletion. 

I will update the pull request with $app->withFacades() change and make the readme clearer as well.

kiding:
After struggling for hours looking through https://github.com/barryvdh/laravel-ide-helper/pull/438#issuecomment-571353601 and https://github.com/barryvdh/laravel-ide-helper/issues/611#issuecomment-361163745 and so many other issues, the cleanest way to tell `Generator` to include `\Eloquent`—even without creating `config/app.php`—I came up with is this:

```php
class_alias(Illuminate\Database\Eloquent\Model::class, 'Eloquent');
$app->config->set('app.aliases', ['Eloquent' => Illuminate\Database\Eloquent\Model::class]);
$app->register(IdeHelperServiceProvider::class);
```

That was rough.

PR #441, opened 2017-02-07
Fix facades fullpath autocomplete

zedentox:
Fix a bug concerning PR #437 

The wrong classes were used for some facades like Eloquent.

Before :
````php
namespace Illuminate\Database\Eloquent {
    class Eloquent {

namespace {
    class Eloquent extends \Illuminate\Database\Eloquent\Model {}
````

After :
````php
namespace Illuminate\Database\Eloquent {
    class Model { // <= Good

namespace {
    class Eloquent extends \Illuminate\Database\Eloquent\Model {}
````

### DIFF
--- a/resources/views/helper.php
+++ b/resources/views/helper.php
@@ -15,7 +15,7 @@ namespace  {
 namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> {
 <?php foreach($aliases as $alias): ?>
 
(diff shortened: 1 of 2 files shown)


Kurre:
Tested and seems to work as it should be 👍

zedentox:
@Kurre Thanks for your feedback !

@barryvdh Thanks for making my IDE happy :)

PR #450, opened 2017-02-18
Fix autocompletion of methods declared in facades

zedentox:
This PR fix a problem with methods declared directly in facades like `Auth::routes()` which are not autocompleted since 2.3.

I tested it on one of my projects and it works for `Auth` but I dont use all facades...

### DIFF
--- a/resources/views/helper.php
+++ b/resources/views/helper.php
@@ -14,7 +14,10 @@ namespace  {
 <?php foreach($namespaces as $namespace => $aliases): ?>
 namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> {
 <?php foreach($aliases as $alias): ?>
-
+    
+    /**
+     * @mixin <?= $namespace ?>\<?= $alias->getExtendsCLass() ?> 
+     */
     <?= $alias->getClassType() ?> <?= $alias->getExtendsCLass() ?> {
         <?php foreach($alias->getMethods() as $method): ?>
 


barryvdh:
Isn't this double? It already extends the class.

zedentox:
When extending the class PHPStorm uses the one declared in `_ide_helper.php`.
When using `@mixin` PHPStorm uses the original facade...

I dont know if its the best way to fix this problem.

resources/views/helper.php
```diff
@@ -14,8 +14,11 @@
 <?php if ($namespace == '\Illuminate\Database\Eloquent'): continue; endif; ?>
 namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> { 
 <?php foreach($aliases as $alias): ?>
-
-    <?= $alias->getClassType() ?> <?= $alias->getExtendsClass() ?> {
+    
+    /**
+     * @mixin <?= $namespace ?>\<?= $alias->getExtendsCLass() ?> 
+     */
```
mfn, line 20: This can't be done like this anymore, because in the meantime the phpdoc of the original class is injected already => https://github.com/barryvdh/laravel-ide-helper/pull/618

@zedentox did you check never versions if they solve your problem?

mfn (changes requested):

mfn:
Ok, re-reading the title and code again, this is exactly what the other PR solved and everyone already enjoys having for years.

@barryvdh pretty confident we can close this PR: we do have autocompletion for Facades (at least as good as it gets currently), and this is obsolete.

Issue #455, opened 2017-02-22
Wrong facades namespace


pinepain:
In laravel-ide-helper v2.3.1 (actually, after v2.2.3) when namespaced facade alias defined, in generated helper it still imported into root namespace (though, in advance to the right namespace).

pinepain:
Closed because of misconfiguration, though we faces with facades duplication =(

barryvdh:
Yes, facades are duplicated, but otherwise we can't provide any autocompletion.

pinepain:
It should be a separate issue, though I haven't yet figured out the cause, but in brief, in service provider I have

```php
public function boot() {
        $loader = AliasLoader::getInstance();

        $loader->alias('My\Remove', My\Awesome\Pacakge\Has\Remote\Facade::class);
}
```

which later doesn't aliased in project app.php, but in ide helper it get imported to root namesapce as Remote, which is not right and which conflict with Laravel's Remote facade which also in a root namespace. And I see not Remote in ide helper under My namespace which is also a bit disappointing. At this time I stay on 2.3.1, but will try to find some time to dig what went wrong in latest versions (haven't checked the diffs yet). If you have something in mind - please suggest.

barryvdh:
Ping @zedentox ideas?

pinepain:
At this time can confirm that issue introduced in 525733122ee14eceabcc60815fcf2b16a7f53d85.

pinepain:
The issue (obviously?) came from https://github.com/barryvdh/laravel-ide-helper/blob/e82de98cef0d6597b1b686be0b5813a3a4bb53c5/resources/views/helper.php#L40-L63 where we assume that all aliases come from root namespace, which is in my case is not.

zedentox:
@pinepain Can you give us the code actually generated in `_ide_helper.php` and your aliases in `app.php` ? It's the first time I see someone using a namespaced alias so I never tested this. 😄 
It's a problem for sure. I will fix it.

pinepain:
@zedentox Not sure about original ide helper as it a private project, but I'll create some small reproducible demo now.

zedentox:
@pinepain Yes of course I only want the wrong parts and what is attended. You can obfuscate all other things.

pinepain:
[...]

PR #456, opened 2017-02-24
Add namespaced aliases

zedentox:
Fix issue #455

### DIFF
--- a/resources/views/helper.php
+++ b/resources/views/helper.php
@@ -11,12 +11,7 @@ namespace  {
     exit("This file should not be included, only analyzed by your IDE");
 }
 
-<?php
-$aliasesGroupedByNamespaceOfExtendedClasses = $valid_aliases->groupBy(function ($alias) {
-    return $alias->getExtendsNamespace();
-});
-?>
-<?php foreach($aliasesGroupedByNamespaceOfExtendedClasses as $namespace => $aliases): ?>
+<?php foreach($namespaces_by_extends_ns as $namespace => $aliases): ?>
 <?php if ($namespace == '\Illuminate\Database\Eloquent'): continue; endif; ?>
 namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> { 
 <?php foreach($aliases as $alias): ?>
@@ -40,12 +35,7 @@ namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> {
 
 <?php endforeach; ?>
 
-<?php
-$aliasesGroupedByNamespace = $valid_aliases->groupBy(function ($alias) {
-    return $alias->getNamespace();
-});
-?>
-<?php foreach($aliasesGroupedByNamespace as $namespace => $aliases): ?>
+<?php foreach($namespaces_by_alias_ns as $namespace => $aliases): ?>
 namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> { 
 <?php foreach($aliases as $alias): ?>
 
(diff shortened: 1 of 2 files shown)


barryvdh:
Thanks! I'd rather not put the logic inside the template, can you move that to Generator class?

zedentox:
I had a big problem with that... How to name theses variables ? 🤣

pinepain:
Thanks for the fix! Is there anything I can help with to make this be good enough to appear in master and be tagged?

zedentox:
@pinepain Can you give a try to this newest version ?

resources/views/helper.php
```diff
@@ -11,16 +11,15 @@
...
-namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> {
+namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> { 
 <?php foreach($aliases as $alias): ?>
 
     <?= $alias->getClassType() ?> <?= $alias->getExtendsClass() ?> {
         <?php foreach($alias->getMethods() as $method): ?>
 
-        <?= trim($method->getDocComment('        ')) ?>
-
+        <?= trim($method->getDocComment('        ')) ?>
```
pinepain: Here is an extra whitespace after closing tag - `?> `. Minor but creates unnecessary diff.

resources/views/helper.php
```diff
@@ -11,16 +11,15 @@
     exit("This file should not be included, only analyzed by your IDE");
 }
 
-<?php foreach($namespaces as $namespace => $aliases): ?>
+<?php foreach($namespaces_by_extends_ns as $namespace => $aliases): ?>
 <?php if ($namespace == '\Illuminate\Database\Eloquent'): continue; endif; ?>
-namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> {
+namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> {
```
pinepain: Extra whitespace at the end of line

resources/views/helper.php
pinepain: Same below. Won't comment on every line. Unless it's intentional - give it a look, maybe it's due to some local ide settings. Apart of that - it works! Yay!

pinepain:
@zedentox Tried a new version and it seems to be working. There is minor comment on trailing white spaces in `views/helper.php`, but apart of that - 👍. Thank you for your time and efforts!

resources/views/helper.php
zedentox: They are intentional. If you remove some trailing whitespaces you will see that sometimes line breaks are ignored... Do you know why ?

barryvdh:
Okay merged, let me know if it's good or not

pinepain:
Thank you! Will do more intensive testing during this week, but the main issue looks solved.

watercayman:
Hi folks,

I'm not at a coding level to know if I'm on the right page or not, but I wanted to send you a little more info.   I've had a similar problem to Pinepain in #455 : I can not click through to methods in my namespaced aliases -- IDE doesn't recognize them, instead clicking points into the ide_helper file.  v2.2.3 **does** successfully work, latest version, including  b611cb5 **does not**.

I register these aliases in config.app only through the aliases array like so:

```
'aliases' => [
    'CacheHelper'=> \App\Helpers\CacheHelper::class,
]
```
Code functions fine, and always has (though I'm sure I'm screwing something up), and IDE Helper has always worked perfectly in identifying these until after 2.2.3.

watercayman:
@zedentox my apologies.  I thought this was related to this PR.  I thought the issue was namespaced aliases were not being picked up, and this PR was meant to fix that.  I was reporting that it does not for those aliases that I have registered.  Again, my apologies, I may not understand enough of this to know the difference between this and the issue looking to be fixed, though I thought they were the same.

PR #459, opened 2017-02-28
removed static declaration

Chrissi2812:
Because the function of Laravel are not static and phpstorms complains about it.

Before this:
image

And after:
image

I'm not sure if this is a proper fix, or why the function are static..
If there is a better way of getting rid of those errors im open for them 😀

### DIFF
--- a/resources/views/helper.php
+++ b/resources/views/helper.php
@@ -21,7 +21,7 @@ namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> {
 
         <?= trim($method->getDocComment('        ')) ?>
 
-        public static function <?= $method->getName() ?>(<?= $method->getParamsWithDefault() ?>)
+        public function <?= $method->getName() ?>(<?= $method->getParamsWithDefault() ?>)
         {<?php if($method->getDeclaringClass() !== $method->getRoot()): ?>
 
             //Method inherited from <?= $method->getDeclaringClass() ?>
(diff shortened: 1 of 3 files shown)


barryvdh:
The functions are not static, but we're calling them static in facades.

Chrissi2812:
So can you explain to me how I can call the method dynamic without phpstorm complaining? 

Is there a special reason that their are called static?

barryvdh:
Are you extending \Eloquent or Model in your models? You can play with that. Otherwise not really sure.

Chrissi2812:
My models look like this:
```php
namespace App;

 /**
 * @mixin \Eloquent
 */
class Article extends Model
{
}
```

if I change the `@mixin` to `Model`, i get the error that phpstorm cant find the `get` Method in **\Illuminate\Database\Eloquent\Builder**. 

With `@mixin \Eloquent` it just used the `get` Method of the `Eloquent` Class from the the _ide_helper.php

So changing it doesn't help much :slightly_frowning_face:

miklcct:
It is laravel brainf**king design which forward static magic calls to non-static magic calls with the same name which native PHP does not support, so there is no fix to this issue.

Artistan:
quick fix here...
https://github.com/barryvdh/laravel-ide-helper/issues/541#issuecomment-317411918

also have a pull request that will automate adding that mixin to the Eloquent/Model until phpstorm and others can handle static mixins.

PR #466, opened 2017-03-15
Add disable write magic methods on models

iboldurev:


vladimir-yuldashev:
+1 for this feature

barryvdh:
What is your use-case to disable this?

vladimir-yuldashev:
@barryvdh I do not like to use magical `where` functions in my project and do not want to encourage usage of them. Instead, only attributes and relations will be in PhpDoc.

vladimir-yuldashev:
@barryvdh After each `php artisan ide-helper:models` I need to remove these lines manually :(

iboldurev:
@barryvdh agrees with the answer @vladimir-yuldashev

barryvdh:
Okay, makes sense. I think that 'magic_where' might be a bit to broad, as you do want the scopes etc to be applied, right? Just not the magic where* methods. So perhaps call it `model_write_magic_where`?

vladimir-yuldashev:
@barryvdh Agree.

iboldurev:
@barryvdh @vladimir-yuldashev I'll fix it now

iboldurev:
@barryvdh done

barryvdh:
Only now you're skipping the scopes also. Is that intended?

Otherwise add the check to the getPropertiesFromTable method, (see the setMethod part)

iboldurev:
@barryvdh please check code.

src/Console/ModelsCommand.php
```diff
@@ -81,6 +82,7 @@ public function fire()
         $model = $this->argument('model');
         $ignore = $this->option('ignore');
         $this->reset = $this->option('reset');
+        $this->write_model_magic_where = $this->laravel['config']->get('ide-helper.write_model_magic_where');
```
barryvdh: Can you change it here to set the default to true? Instead of in the property? In case the config isn't set. ` $this->laravel['config']->get('ide-helper.write_model_magic_where', true);`

iboldurev: ok, just a minute

iboldurev:
@barryvdh you can merge this pr?

PR #471, opened 2017-03-28
add aliases to meta

Ellrion:
When I register some service like 

```
$this->app->singleton([FlashMessageSender::class => 'flash']);
```

for inject by classname (in other services) and using like `app('flash')` (in controller)
I dont have autocomplete in second usecase.

To fix this a suggest add aliases to meta file.


mfn (changes requested):
To me the changes look good!

Generally, variable names should be `$camelCase`

@Ellrion are you still interested in moving forward and resolving the merge conflict? I can help out with a test then.

PR #472, opened 2017-03-29
Updated installation on fresh Lumen 5.4

DarkGhostHunter:
Instructions should be more clear now for how to install, and what to do to add additional Facades.

It also gets rid of the "watch out of the `config/app.php` file" warning, as it loads only if Laravel IDE Helper is called under console/terminal (as it should be).


mfn:
IMHO this can be dropped, this package only supports Laravel 5.5+ so Lumen 5.4 is out of the box anyway?

DarkGhostHunter:
Should.

PR #475, opened 2017-03-30
Decoupled from main app View Factory

theMadness:
This should safeguard from app-dedicated overrides to the View Factory used in-app.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,8 @@
         "symfony/class-loader": "^2.3|^3.0"
     },
     "require-dev": {
+        "illuminate/config": "^5.0,<5.5",
+        "illuminate/view": "^5.0,<5.5",
         "phpunit/phpunit" : "4.*",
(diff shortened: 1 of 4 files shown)


barryvdh:
Interesting, you were using a different engine and removed the default one?

theMadness:
Well the default one doesn't parse PHP (we use twig) so it was failing.

barryvdh:
Okay I also use Twig but usually leave the php engine on.

apollopy:
1. This code is invalid, it should be deleted
https://github.com/theMadness/laravel-ide-helper/blob/405cb6baa259eb0672986c7c592b744fc1a4b17b/src/IdeHelperServiceProvider.php#L39-L40

2. Hope to provide in the configuration file to set the view path
https://github.com/theMadness/laravel-ide-helper/blob/405cb6baa259eb0672986c7c592b744fc1a4b17b/src/IdeHelperServiceProvider.php#L105

PR #481, opened 2017-04-04
Fix #480 where tinyint is casts as bool

### LINKED ISSUE #480: TINYINT database column identified as bool by Automatic phpDocs for models
Using `php artisan ide-helper:models`, a `TINYINT` column is documented as `bool`.

While `TINYINT` is commonly used to represent `bool`, I am using it to map to a small amount of model states.

And while I do not know what's the best way to differentiate the two, the length of the column can be used to at least rule out `bool` if it's `>1`. i.e. `TINYINT(2)` can never be `bool`.

komirad:
For issue https://github.com/barryvdh/laravel-ide-helper/issues/480

Explanation of changes:
`tinyint` is classified as `bool` by doctrine/dbal, but it should not be so in our use case. `tinyint` is better PHPDoc'ed as `int` (as it can be used for many reasons). **If the user wants `bool`, he should use Eloquent's `$casts` property to do that.**

I have therefore, moved doctrine/dbal's `boolean` to set the property type as `int` in PHPDoc.

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -347,14 +347,15 @@ class ModelsCommand extends Command
                         case 'integer':
                         case 'bigint':
                         case 'smallint':
+                        /**
+                         * @see https://github.com/barryvdh/laravel-ide-helper/issues/480
+                         */
+                        case 'boolean':
                             $type = 'integer';
                             break;
                         case 'decimal':
                         case 'float':
                             $type = 'float';
-                            break;
-                        case 'boolean':
-                            $type = 'boolean';
                             break;
                         default:
                             $type = 'mixed';
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -347,10 +347,7 @@ class ModelsCommand extends Command
                         case 'integer':
                         case 'bigint':
                         case 'smallint':
-                        /**
-                         * @see https://github.com/barryvdh/laravel-ide-helper/issues/480
-                         */
-                        case 'boolean':
+                        case 'boolean': // See issue #480
                             $type = 'integer';
                             break;
                         case 'decimal':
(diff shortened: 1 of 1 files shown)


barryvdh:
So when user adds cast to bool, it will be correct, right?

Can you remove the comment to this issue from the code?

komirad:
@barryvdh 

Yes, when user set casts, the generated type would be overridden. i.e. Cast to bool, will become bool.

barryvdh:
Thanks!

Belphemur:
Because of this change, PostgreSQL boolean are considered now as TinyInt which doesn't make any sense since it doesn't exists.

In Postgres you only have boolean to represent boolean value. This PR should be rollback, MySQL is the issue here. 

I would rather have the user having to set in their model the cast from Boolean -> int than having to make the user do the opposite.

komirad:
I would suggest that we detect the db type and handle it  accordingly: http://stackoverflow.com/questions/17908309/drupal-how-can-i-know-if-the-db-is-mysql-or-postgres

Belphemur:
I'm sorry but tinyint is a boolean: https://dev.mysql.com/doc/refman/5.7/en/integer-types.html
It's a bit that can only be 1 or 0. The definition of a boolean.

It has no reason to be casted as an integer. Doctrine defines it correctly, this PR doesn't make any sense.

Give me an example where a tinyint is not used either a boolean or a flag. I'm curious of your use case.

barryvdh:
Isn't it 1 byte instead of 1 bit, as stated between -128 / +127

Belphemur:
Sorry for my last comment.
Yes, your right, I misread and I apologise. 

I would think this PR would be better implemented as a setting in the database. To not break the current way it works and stay backwards compatible.

komirad:
We can use Laravel's `config('database.default')` to determine the DB type. Added PR.

PR #487, opened 2017-04-07
Implements Nullable

### LINKED ISSUE #407: generated phpDoc for models does not include |null on nullable columns
When using ide-helper:models --write, the generated phpDocBlock does not contain |null annotation on nullable properties. The |null annotation should be added on nullable columns for user to know that the column can be nullable.

Belphemur:
Using Doctrine notNull on column, this pull request implement the detection of Nullable.

It does it for the properties of the model and the relation that are using the property.

Fixes #407


barryvdh:
Thanks!

PR #489, opened 2017-04-10
Add support for pgsql for boolean types

komirad:
As discussed in #481 

Due to the changes, `pgsql`'s `boolean` became `int` too. This change uses Laravel's `config('database.default')` to determine the db type. Where it's the case that it is `pgsql`, it will use `boolean`.

[sqlite also uses int as boolean](https://sqlite.org/datatype3.html#boolean_datatype)


barryvdh:
Thanks

PR #493, opened 2017-04-17
Use short array syntax in configuration file

vyuldashev:
As long as this library required PHP >= 5.4 we can use short array syntax in config.php

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -48,9 +48,9 @@ return array(
 
     'include_helpers' => false,
 
-    'helper_files' => array(
+    'helper_files' => [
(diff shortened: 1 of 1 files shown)


mfn:
The current codebase in generall still uses `array()` by and large.

I suggest instead of just changing one file, if @barryvdh wants to change this, we should employ e.g. [php cs fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) to a) convert the whole codebase and b) ensure it stayts that way.

@barryvdh what do you think? Short array syntax is very common already and old PHP versions aren't a concern anymore.

barryvdh:
Sure

PR #495, opened 2017-04-23
Reversing order of default facade list

dcarbone:
Hi,

First off thanks for creating this lib, makes developing Lumen much less of a headache :)

Secondly, I've run into a small issue with alias name-clashing in the generated _ide_helper file that this PR hopes to rectify.

For one particular project, we are using [Event](http://docs.php.net/event) for use with React.  This extension exposes a class called `Event`, which obviously collides with the default Laravel `Event` facade class alias.

To that end, I have flipped the array in `Generate::getAliases`, making the class itself the key.  This allows me to squash the default definition for `Illuminate\Support\Facades\Event` with the alias that is actually used in my code, `LumenEvent`.

It also allows Lumen users to take advantage of this lib without having to call `array_flip` before providing their aliases to the Lumen application, as the `Application::withAliases` call expects this array format.   See [Application::withAliases](https://github.com/laravel/lumen-framework/blob/5.4/src/Application.php#L630)

This change should not pose any breaking changes for current users, and allows Lumen users to modify the aliases as they see fit.

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -248,32 +248,47 @@ class Generator
             return AliasLoader::getInstance()->getAliases();
         }
 
+        // Lumen
         $facades = [
-          'App' => 'Illuminate\Support\Facades\App',
-          'Auth' => 'Illuminate\Support\Facades\Auth',
-          'Bus' => 'Illuminate\Support\Facades\Bus',
-          'DB' => 'Illuminate\Support\Facades\DB',
-          'Cache' => 'Illuminate\Support\Facades\Cache',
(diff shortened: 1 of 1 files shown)


mfn:
To be frank, this sounds like an edge case? Can't just use a different name?

Besides, the PR contains a LOT of code style change noise.

I suggest to drop this.

dcarbone:
@mfn that is exactly what this PR was attempting to do.  I've since stopped using the extension altogether for this and other reasons, so close away.

mfn:
@dcarbone sorry to hear

PR #500, opened 2017-05-04
Adds support for macros

mjerwin:
Generates stubs for marcos if the class uses the *Marcroable* trait.

### DIFF
--- a/src/Alias.php
+++ b/src/Alias.php
@@ -315,11 +315,10 @@ class Alias
 
             // Check if the class is macroable
             $traits = collect($reflection->getTraitNames());
-            if($traits->contains('Illuminate\Support\Traits\Macroable')) {
+            if ($traits->contains('Illuminate\Support\Traits\Macroable')) {
                 $properties = $reflection->getStaticProperties();
                 $macros = isset($properties['macros']) ? $properties['macros'] : [];
-                foreach($macros as $macro_name => $macro_func)
-                {
+                foreach ($macros as $macro_name => $macro_func) {
                     $function = new \ReflectionFunction($macro_func);
                     // Add macros
                     $this->methods[] = new Macro(
--- a/src/Alias.php
+++ b/src/Alias.php
@@ -312,6 +312,25 @@ class Alias
                     }
                 }
             }
+
+            // Check if the class is macroable
+            $traits = collect($reflection->getTraitNames());
+            if($traits->contains('Illuminate\Support\Traits\Macroable')) {
+                $properties = $reflection->getStaticProperties();
+                $macros = isset($properties['macros']) ? $properties['macros'] : [];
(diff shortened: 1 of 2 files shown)


barryvdh:
Cool thanks

jonnywilliamson:
@barryvdh Could we get a new tag for a release please?

@mjerwin Fantastic. Thanks so much for this.

barryvdh:
Did you test this?

barryvdh:
You can get it with `2.3@dev`

jonnywilliamson:
I could only pull it in if I used `composer --dev require barryvdh/laravel-ide-helper:dev-master` but I don't like having dev-master in my composer file. When I tried 2.3.x@dev I only got ver 2.3.2 not this latest commit.

Anyway, I assumed because you have merged it was working so I'm just after testing it. 

Mainly I was trying to see if I could get autocomplete when using https://github.com/spatie/laravel-collection-macros 

However, after updating to this new commit, and generating a new `_ide-helper.php` file, I still don't get methods like `->ifAny()` on a collection.

I see the PR says that "Generates stubs for macros if the class uses the Macroable trait." but it looks like thats not going to help me in this case.

mjerwin:
Hey @jonnywilliamson,

Sorry, should have made it clearer. This only works on Facades so it currently wouldn't work for the `Collection` class.

aledelgo:
First, thanks for your contribution. this was a really needed feature! 

Unfortunately on my environment it doesn't works fine well... i'm on PhpStorm with laravel 5.3

i have this code in a custom registered provider (my code is working fine)

```
class ResponseMacroServiceProvider extends ServiceProvider {

    /**
     * Perform post-registration booting of services.
     *
     * @return void
     */
    public function boot()
    {



        /**
         * Create a response()->apiResponse(...) function
         */
        Response::macro('apiResponse', function($apiResultObj, $serializazionGroups=null, $includeDefaultGroups=true, $jsonPCallback=null)
        {

            /* custom code here...*/

            return $resp;


        });
    }

}
```

I see ide-helper correctly generate this in __ide_helper.php_ file


```
/**
         * Create a response()->apiResponse(.
         * 
         * ..) function
         *
         * @static 
         */ 
        public static function apiResponse($apiResultObj, $serializazionGroups = null, $includeDefaultGroups = true, $jsonPCallback = null)
        {
            return \Illuminate\Routing\ResponseFactory::apiResponse($apiResultObj, $serializazionGroups, $includeDefaultGroups, $jsonPCallback);
        }
```

but the IDE don't recognize it in the autocomplete list.
when i write:
` return response()->apiResp`
the method is not listed in the autocomplete list, and if i write it completely like 
`return response()->apiResponse($result, [], false);`
the method is underlined and marked with a tooltip sayng the method is not found 
image

Do you have any idea?

PR #505, opened 2017-05-09
Added check for Variadic

Daviex:
Row 264-266, added a check for the parameter if it's variadic.
If it's variadic, is useless to check for the optional. Also, it adds the dots before the variable name on the parameters declaration.

PHP.Net new features 5.6:
http://php.net/manual/en/migration56.new-features.php#migration56.new-features.variadics

### DIFF
--- a/src/Method.php
+++ b/src/Method.php
@@ -261,7 +261,10 @@ class Method
         foreach ($method->getParameters() as $param) {
             $paramStr = '$' . $param->getName();
             $params[] = $paramStr;
-            if ($param->isOptional()) {
+            //Check if the parameter is variadic, then, is useless to check if optional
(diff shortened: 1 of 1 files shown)


barryvdh:
This doesn't work for all supported php versions

Daviex:
Last Laravel works only with 5.6>=, maybe could add a branch forma new PHP versione... Without this will have errors on some functions

barryvdh:
I'm okay with bumping to php 5.6

mfn:
This is already implemented, see https://github.com/barryvdh/laravel-ide-helper/blob/d9a731daba7324502fb37f055960b9e1b57b814e/src/Method.php#L317-L319

This PR can be closed

PR #508, opened 2017-05-23
Don't use Facade alias but actual class

mfn:
There may be projects who don't have/need this particular alias.

---
IMHO it's good practice to only use facade alias locally in your project but not as 3rd party add-on: you can never now if it's really there.


Artistan:
`@mixin \Eloquent` is built by this package. It provides references to magic static method calls on non-static methods.

mfn:
I can't remember why I had problems with it (special/edge case probably), closing, thx!

PR #518, opened 2017-06-07
generate methods for the SoftDeletes trait

max-matiukhin:
this patch implements auto-completion  for the SoftDeletes trait methods


miklcct:
The query classes returned has also to be changed if #521 is merged

PR #523, opened 2017-06-13
Use new PhpStorm advanced metadata format

stidges:
This commit adds support for the new [PhpStorm Advanced Metadata](https://confluence.jetbrains.com/display/PhpStorm/PhpStorm+Advanced+Metadata) format.

What this allows us to do is to add autocompletion for the several (array) helper methods that Laravel offers:

array-helper-autocompletion

I have also added support for the new `makeWith` and `resolve` functions.

The only thing that I think could be improved here is not having to hardcode the helper functions, but I couldn't think of a better way to handle it.

### DIFF
--- a/resources/views/meta.php
+++ b/resources/views/meta.php
@@ -9,13 +9,42 @@ namespace PHPSTORM_META {
     * @author Barry vd. Heuvel <barryvdh@gmail.com>
     * @see https://github.com/barryvdh/laravel-ide-helper
     */
-    $STATIC_METHOD_TYPES = [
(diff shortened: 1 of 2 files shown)


barryvdh:
Cool. Does this work for all recent versions? Do we need to keep the old format also?

stidges:
This should work on 2016.2 and up, not sure if people are using old versions of Phpstorm. We could also throw this behind a config flag just in case?

pinepain:
@stidges Thanks a lot! Decent change! Didn't know jetbrains improved metadata so greatly.

PR #528, opened 2017-06-25
Add configurable Formatter Control tags around model docblocks

JackWH:
Firstly, thanks for such an invaluable package! I have only one minor irritation with it - that this pull request fixes - so I hope it's useful to others too. 😄 

I regularly use `php artisan ide-helper:models --write --reset` to keep model documentation up-to-date and accurate. IDE Helper writes docblocks like so:

```
/**
 * App\MyModel
 *
 * @property int $id Model identifier
 * @property string $name Name of model entry
 * @property Carbon $created_at Date of creation
 */
```

However, my PhpStorm auto-reformatting is configured to align docblock tags, like so:

```
/**
 * App\MyModel
 *
 * @property int    $id         Model identifier
 * @property string $name       Name of model entry
 * @property Carbon $created_at Date of creation
 */
```

So if I do an auto-reformat of my file, all tags that IDE Helper generated are neatly aligned. However, the next time I run `php artisan ide-helper:models --write --reset`, all that auto-formatting is undone. Unless I manually re-format everything again, my git history gets filled up with every model's docblock being rewritten in many commits.

To get around this, I've added configuration options to IDE Helper that allow you to set [formatter control tags](https://www.jetbrains.com/help/phpstorm/code-style.html#d366576e184) that surround the docblocks that are written to models. With this enabled, you can auto-reformat as much as you like, whilst still automatically keeping your docblocks up-to-date, without filling your git history with unnecessary reformatting commits.

With the option enabled, by default, docblocks are written like this:

```
// @formatter:off
/**
 * App\MyModel
 *
 * @property int $id Model identifier
 * @property string $name Name of model entry
 * @property Carbon $created_at Date of creation
 */
// @formatter:on

```

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -176,4 +176,23 @@ return array(
         'integer' => 'int',
         'boolean' => 'bool',
    ),
+    
+    /*
+    |--------------------------------------------------------------------------
+    | Formatter Control
+    |--------------------------------------------------------------------------
+    |
+    | Writes Formatter Control tags around model docblocks, to prevent automatic
+    | IDE reformatting changing any of the docblocks written by the IDE helper.
+    | This is useful when using the --write flag, to help keep version control
+    | histories clean between re-formats and IDE Helper re-writes.
+    |
+    */
+
+    'write_formatter_controls' => false,
+    
+    'formatter_control_tags'   => array(
+        'formatter_off' => "// @formatter:off\n",
+        'formatter_on'  => "\n// @formatter:on\n",
+    ),
 );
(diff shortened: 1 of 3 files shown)


barryvdh:
Hmm, can't we make the indentation configurable?

JackWH:
By indentation, do you mean using `/*` vs `//`?

If so, in my testing with PhpStorm, formatter controls *only* work when prefixed with `//` - putting them within the main `/** ... */` docblock prevents them from having any effect.

JackWH:
Any thoughts on this? I know it would be very useful to our team.

Thanks in advance! 😄

JackWH:
I'm going to close this PR, as it's quite outdated. For anyone else looking for this, the simplest solution is to simply manually add the Formatter Control tags yourself to any new models you create. 

I think they used to be removed when the docblocks were recreated by IDE Helper, but it looks like this is no longer the case in recent versions.

PR #532, opened 2017-06-27
Add back leading slash on meta class names

stidges:
After my last change to support the meta metadata format, I accidentally nuked the leading slash on the class names :(. This PR adds that slash back to fix autocompletion.


jonnywilliamson:
I have to say I didn't notice any issue with autocomplete when the backslash wasn't there! And I WAS using a version of this repository that had your new meta format in it.

Hmmm....

PR #536, opened 2017-07-10
Camel case setting mutators

MaartenStaa:
This adds a setting to control whether mutators get converted to snake case or preserve their camelCase format instead.

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -165,6 +165,18 @@ return array(
     'model_camel_case_properties' => false,
 
     /*
+     |--------------------------------------------------------------------------
+     | Snake cased mutators
+     |--------------------------------------------------------------------------
+     |
+     | When this option is true (default), properties such as from a
+     | getFooBarAttribute mutator get turned into foo_bar. Change this to false
+     | to use camel case instead, resulting in fooBar.
+     |
+     */
+    'model_snake_case_mutators' => true,
+
+    /*
     |--------------------------------------------------------------------------
     | Property Casts
     |--------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


barryvdh:
Hmm, I think this isn't a best practice in Laravel.

MaartenStaa:
I don't see how this is fundamentally different from the setting to convert database columns to camel case.

For better or worse, I'm working on a project where both database columns and, therefore naturally, also mutators are camel case.

mfn:
@MaartenStaa do you still need this or can it be closed?

MaartenStaa:
Yes, I do. At the moment if I regenerate annotations on my models there’s several cases I have to adjust by hand to the correct casing for that project.

mfn:
I thought about it a few times, but I'm with https://github.com/barryvdh/laravel-ide-helper/pull/536#issuecomment-314542069 : this really isn't the "Laravel way".

I think it's of course fine that people change how they use it, but that doesn't mean that they can also expect every Laravel-ecosystem tool works for them.

This would add complexity to the codebase for a very special case not many people are going to have, yet the library will have to support this for the future.

I'm on 👎 for this feature.

----
> I have to adjust by hand to the correct casing for that project

What I can imagine is that we make such things possible by making sure the ide helper library has a _stable and easy to use API_ for that, so that it's possible create eg. a custom ModelsCommand so they can extend / adapt it to their workflow.

However at the moment we are not there. The current implementation would need refactor but that it's impossible right now as the coverage is clearly too low to guarantee no breaks.

mfn:
@barryvdh I vote for closing the PR, IMHO it's too special-case

PR #543, opened 2017-07-24
readme update

TiddlyWin:


mfn:
The current readme already contains this:
> Note: bootstrap/compiled.php has to be cleared first, so run php artisan clear-compiled before generating.

This PR can be closed

PR #544, opened 2017-07-25
Add an eloquent command

Artistan:
for adding the `@mixin \Eloquent` to the Illuminate\Database\Eloquent\Model

### DIFF
--- a/src/Console/EloquentCommand.php
+++ b/src/Console/EloquentCommand.php
@@ -10,10 +10,7 @@
 
 namespace Barryvdh\LaravelIdeHelper\Console;
 
-use Barryvdh\Reflection\DocBlock;
-use Barryvdh\Reflection\DocBlock\Context;
-use Barryvdh\Reflection\DocBlock\Serializer as DocBlockSerializer;
-use Barryvdh\Reflection\DocBlock\Tag;
+use Barryvdh\LaravelIdeHelper\Eloquent;
 use Illuminate\Console\Command;
 use Illuminate\Filesystem\Filesystem;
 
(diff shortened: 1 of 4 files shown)


Artistan:
This may be called by itself, or possibly when the ide-helper is run?

barryvdh:
Why does it need to be a separate command?

Artistan:
No need really, I just did not want to presume where you thought it should be added.  Is it part of the ide-helper, or the model writer, or a seperate command. As it's own command it could be run by itself for but I guess it is useless without the writing of the \Eloquent mixin.

Artistan:
I will add it to the Generate class, since that creates the \Eloquent helper.

Artistan:
@barryvdh , I updated the Generator so it will add the mixin to the Eloquent\Model after the helper is created. (since it is irrelevant without the helper) 

1. Eloquent class as a helper, similar to Generator (separation of logic)
2. Add helper to write mixin to Model class after creating helper
3. keep command to allow running mixin writer in case of updating framework

Artistan:
mixin written to file...
image

mixin already exists...
image

Artistan:
bump.

artem328:
When this estimated to release?

KerryJones:
+1

GTCrais:
+1

ErSulba:
this resolves so many headaches.... i'm waiting for this, since laravel and phpstorm last update makes me to manually config all over again every project i have

ivansammartino:
Still I get PHPStorm warning when using code like `User::truncate();` in `DatabaseSeeder.php` (everything else seems to work fine)

kmark:
This only works for me if the added tag is the before the others already in my `Model.php`. Since, at least for me, it is not, this doesn't work (due to only appending the tag).

Issue #469, opened 2017-03-25
phpstorm.meta.php can now be a folder-name containing multiple files!


jonnywilliamson:
@barryvdh 

Over the last 2 weeks I came across an article online that described a new feature of phpstorm with regards processing of the `.phpstorm.meta.php` file.

Alas I can not for the life of me find that article, but it explained the following:

`.phpstorm.meta.php` can now be the name of a FOLDER in your project. Inside the folder you can now have MULTIPLE `php` files that contain custom data.

I have just tested this out and it seems to be working really well. The only thing you MUST do after creating the `.phpstorm.meta.php` **FOLDER** and moving any metadata **FILES** into it, is _RESTART_ the IDE. It won't pick up those files until the IDE has been restarted.

I was wondering if you might be interested in either
1) provide a config option to create the folder `.phpstorm.meta.php` instead of only the file `.phpstorm.meta.php`
2) make the folder the new default method when generated with this package?

I would attempt to send a PR if I knew what option you leaned more towards.

This then allows anyone to add their own files into the project to be used AS WELL as the one generated by this repository. Currently, because only 1 file is being read by the IDE at the moment it's hard to customise or add to this package's `.phpstorm.meta.php` file because it keeps getting overwritten relatively frequently and destroying any manual additions...

Here's an example of a previous issue: https://github.com/barryvdh/laravel-ide-helper/issues/337#issue-143846792

jonnywilliamson:
Of course as soon as I posted this I found the article online. 

https://confluence.jetbrains.com/display/PhpStorm/PhpStorm+Advanced+Metadata#PhpStormAdvancedMetadata-Deprecated:Legacymetadataformat(2016.1andearlier)

![image](https://cloud.githubusercontent.com/assets/2513663/24322747/44c96346-1162-11e7-9c8c-0a0b9f3ce717.png)

jonnywilliamson:
Do you have any thoughts on this @barryvdh ? Positive or negative! :)

barryvdh:
> By using example metadata below It can reside anywhere, but we recommend using .phpstorm.meta.php file in project root.

So should already work?

jonnywilliamson:
@barryvdh 

Evening, so I did some testing this evening and these were the results.

- If there is a `.phpstorm.meta.php` file, phpstorm will only read meta data from that file. ie, if you create `another.meta.php` file in the root folder, and put valid data in it, it does not get parsed.

[...]

PR #553, opened 2017-08-10
Allow for configurable meta filename

inxilpro:
PhpStorm supports putting multiple files in a `.phpstorm.meta.php/` directory. If you want to maintain multiple meta files, having to use the filename argument each time is a pain, so this makes that configurable.

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -13,6 +13,8 @@ return array(
 
     'filename'  => '_ide_helper',
     'format'    => 'php',
+    
+    'meta_filename' => '.phpstorm.meta.php',
 
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jonnywilliamson:
This would also be a good solution for https://github.com/barryvdh/laravel-ide-helper/pull/482

I'd be happy to close that PR for this one 👍

inxilpro:
Oops. Didn't see your PR. I think just configuring the file name would probably be a simpler option.

jonnywilliamson:
@inxilpro Yes I like your one much better.

jonnywilliamson:
As an aside I've been doing this for the past few months and it's working great also.

https://github.com/barryvdh/laravel-ide-helper/issues/469#issuecomment-310352549

PR #563, opened 2017-08-28
Fix magic method generation & helper class aliases

pantherdd:
Main fixes:
- Fix magic method generation: Before this, methods added via the 'magic' configuration option were calling a method in the root class with the 'magic' name instead of the 'real' name.
- Fix helper class aliases: Aliases for helper classes were mistreated before: the generator would write empty classes that hid the real helper implementation from the IDE. After the change, classes with the real namespace and class name are only generated for facades.

Miscellaneous:
- Fix `Generator` error output: The `OutputInterface` given to the constructor was never saved.
- Some cleanup of `Alias`, `Method` and `Macro`

### DIFF
--- a/resources/views/helper.php
+++ b/resources/views/helper.php
@@ -12,8 +12,8 @@ namespace  {
 }
 
 <?php foreach($namespaces_by_extends_ns as $namespace => $aliases): ?>
-<?php if ($namespace == '\Illuminate\Database\Eloquent'): continue; endif; ?>
-namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> { 
+<?php if($namespace == '\Illuminate\Database\Eloquent'): continue; endif; ?>
+namespace <?= ltrim($namespace, '\\') ?> { 
 <?php foreach($aliases as $alias): ?>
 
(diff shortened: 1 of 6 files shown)


PantherDD:
@barryvdh This is my first time contributing here, so I'm not sure what the process is usually. Please let me know if there's anything else I need to do before this PR can be considered for a merge. Thanks!

Artistan:
@PantherDD - that is a lot of code to review. You may want to chunk it out into multiple pull requests if possible.

mfn:
Thanks for your contribution @PantherDD , but I agree: there are too much changes and it's mixed with clean up. Best is to submit separate PRs for each feature and clean up etc.

@barryvdh I vote for closing the PR!

pantherdd:
@Artistan, @mfn: Sorry, I never got time to clean this PR up when you asked. Unfortunately since then I'm not working with Laravel anymore, and didn't follow all the changes to this project, so I don't think I'll be able to get this into the desired state. Feel free to take the ideas / code from this PR and apply them as you see fit, if they still make sense; the PR itself can be closed I think.

Cheers, and keep up the good work! :)

PR #577, opened 2017-09-21
Follow @mixin tags for Facade roots

inxilpro:
If a Facade's root class is annotated with a `@mixin` docblock tag, this will include that class when searching for methods to alias.

For example, if a facade's root was `B::class`:

```php
class A {
  public function foo() {}
}

/**
 * @mixin A
 */
class B {}
```

Before, the ide-helper alias would have no methods, but after this change it would add the `public static function foo() {}`;

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 language: php
 
 php:
-  - 5.5
   - 5.6
   - 7.0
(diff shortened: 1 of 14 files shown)


inxilpro:
Is there anything I can do to help get this merged?

PR #583, opened 2017-10-13
Support for ::fake() calls

inxilpro:
The add support for [fakes](https://laravel.com/docs/5.5/mocking). Before, if you called `Mail::fake()`/etc, all the assertion methods would not be auto-completed. This PR fixes that.


xakzona:
Usefull feature

PR #596, opened 2017-12-04
Ignore models by default

bozhkos:
This PR adds ability to define models, which should be ignored by default in config file. Not very big deal, but can be useful in some cases. 

For now, models should be defined in same way as on `ide-helper:models`, example:
```php   
 'ignored_models' => [
        'Models\\Stock\\Products\\Brand'
    ],
```
If this feature required by community, i will try to take care about more simple and convenient way.


bozhkos:
seems like community action abandoned here 😬

PR #600, opened 2017-12-27
Use Illuminate\Support\Carbon instead of Carbon\Carbon

mike-grinspan:
Laravel has a wrapper class for Carbon that it uses for model attributes in the `$dates` array and the `created_at` and `updated_at` attributes. This pull request changes all model docblocks to use the wrapper class instead of the base class.

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -153,13 +153,13 @@ return array(
      |
      | For example, normally you would see this:
      |
-     |  * @property \Carbon\Carbon $created_at
-     |  * @property \Carbon\Carbon $updated_at
+     |  * @property \Illuminate\Support\Carbon $created_at
+     |  * @property \Illuminate\Support\Carbon $updated_at
      |
      | With this enabled, the properties will be this:
      |
-     |  * @property \Carbon\Carbon $createdAt
-     |  * @property \Carbon\Carbon $updatedAt
+     |  * @property \Illuminate\Support\Carbon $createdAt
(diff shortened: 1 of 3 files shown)


svenluijten:
Any update on this? 🙂 This was included in Laravel since https://github.com/laravel/framework/pull/19771, would be useful to have in ide-helper as well 😄

mrextreme:
It still mixes \Illuminate\Support\Carbon and Carbon\Carbon when generating phpdoc blocks for models. Laravel 5.x, 6.x, 7.x, ide helper 2.6.7

mfn:
> mixes \Illuminate\Support\Carbon and Carbon\Carbon

But your example only has `\Carbon`? that seems totally of… ?

mrextreme:
I copy-pasted that from the actual code, while type in the text above it from head memory.  Just deleted that code block and ran ide-helper:models again, and the output is as copied it in yesterday, no change. \Carbon doesn't even exists, it should be \Carbon\Carbon. Just searched the source code at ...maypath/vendor/barryvdh/laravel-ide-helper, and everywhere i see \Illuminate\Support\Carbon, yet the output randomly produces \Carbon for some fields. Both start and stop are DATE fields, yet they end up having different classes.

svenluijten:
This discussion doesn't seem to be related to this issue? Maybe create a new issue so it can be tracked better? :)

mrextreme:
ok, sure

PR #618, opened 2018-02-01
Add DocBlocks for classes too

### LINKED ISSUE #606: Route::middleware is missing from _ide_helper.php
I've noticed Laravel 5.5 calls `Route::middleware()` in `app/Providers/RouteServiceProvider.php:54`
and although class Route is present in `_ide_helper.php` it is missing `middleware()`

Please advise.

glamorous:
Because the IDE-helper is missing some crucial facade hinting. I added the original docblocks of the Facades (provided by Laravel) to the created classes in this file. I used the already logic that I found in Method.php

This will close #606 and a lot of other requests of missing documentation for the classes.

### DIFF
--- a/resources/views/helper.php
+++ b/resources/views/helper.php
@@ -15,6 +15,7 @@
 namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> { 
 <?php foreach($aliases as $alias): ?>
 
+    <?= trim($alias->getDocComment('    ')) ?> 
     <?= $alias->getClassType() ?> <?= $alias->getExtendsClass() ?> {
         <?php foreach($alias->getMethods() as $method): ?>
 
--- a/src/Alias.php
+++ b/src/Alias.php
@@ -10,6 +10,11 @@
 
 namespace Barryvdh\LaravelIdeHelper;
 
+use Barryvdh\Reflection\DocBlock;
+use Barryvdh\Reflection\DocBlock\Context;
+use Barryvdh\Reflection\DocBlock\Serializer as DocBlockSerializer;
+use ReflectionClass;
+
 class Alias
 {
     protected $alias;
@@ -27,6 +32,7 @@ class Alias
     protected $valid = false;
     protected $magicMethods = array();
     protected $interfaces = array();
+    protected $phpdoc = null;
 
     /**
      * @param string $alias
@@ -56,7 +62,13 @@ class Alias
         $this->detectNamespace();
         $this->detectClassType();
         $this->detectExtendsNamespace();
-        
+
+        if(!empty($this->namespace)) {
+            //Create a DocBlock and serializer instance
+            $this->phpdoc = new DocBlock(new ReflectionClass($alias), new Context($this->namespace));
+        }
+
+
         if ($facade === '\Illuminate\Database\Eloquent\Model') {
             $this->usedMethods = array('decrement', 'increment');
         }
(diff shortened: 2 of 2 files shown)


szepeviktor:
@barryvdh What do you think?

szepeviktor:
@glamorous Thank you for your improvement.

It made phpstan understand Laravel code!

The only thing left in vanilla laravel is:

```
 ------ --------------------------------------------------------
  Line   app/Http/Controllers/Auth/RegisterController.php
 ------ --------------------------------------------------------
  65     Call to an undefined static method App\User::create().
 ------ --------------------------------------------------------
```

How could this Eloquent method be recognized?

```php
class RegisterController extends Controller
{
    /**
     * Create a new user instance after a valid registration.
     *
     * @param  array  $data
     * @return \App\User
     */
    protected function create(array $data)
    {
        return User::create([
            'name' => $data['name'],
            'email' => $data['email'],
            'password' => bcrypt($data['password']),
        ]);
    }
}
```

glamorous:
@szepeviktor No problem, I made the PR for our PHPStan too. For those Eloquent code hinting, we've extended the default Model class and added the needed PHP Docblocks ourself. You could add this in your user-model too, but the better way is to have a default model and add

```
/**
 * Abstract Model class, used for all our own models.
 *
 * @method static Model|Builder create($attributes = array())
 * @method static Builder distinct()
 * @method static Model|Collection|static[]|static|null find($id, $columns = array())
 * @method static Model|Builder  forceCreate($attributes)
 * @method static Builder inRandomOrder($seed = '')
 * @method static bool insert($query, $bindings = array())
 * @method static Builder orderBy($column, $direction = 'asc')
 * @method static Collection pluck($column, $key = null)
 * @method static Builder select($columns = array())
 * @method static Model updateOrCreate($attributes, $values = array())
 * @method static Builder when($value, $callback, $default = null)
 * @method static Builder where($column, $operator = null, $value = null, $boolean = 'and')
 * @method static Builder whereIn($column, $values, $boolean = 'and', $not = false)
 * @method static Builder|static whereNotExists($callback, $boolean = 'and')
 * @method static Builder|static whereNotNull($column, $boolean = 'and')
 * @method static Builder whereNull($column, $boolean = 'and', $not = false)
 * @method static Builder join($table, $first, $operator = null, $second = null, $type = 'inner', $where = false)
 *
 */
abstract class Model extends \Illuminate\Database\Eloquent\Model
{}
```

mfn:
> we've extended the default Model class and added the needed PHP Docblocks ourself

Anyone tried if this workaround would also work with a trait (which wouldn't "do" anything except having this phpdoc)?

pablorsk:
Great PR, but, how can we solve a problem like this?

```php
        $document = Document::findOrFail($some_id);
        public static function do_something(Document $document) {
              // doing something...
        }
```
PHPStan
```
Parameter #1 $document of static method App\Service::do_something() expects App\Document,                                                               
         Illuminate\Database\Eloquent\Collection|Illuminate\Database\Eloquent\Model given.
```

I try adding to Document model
```
 * @method static \App\Document findOrFail($id)
```
But the problem persists.

Only way to solve problem is adding to model (very ugly :-/)

```php
    /**
    * @return self
    */
    public static function findOrFail($id, $columns = array()) {
        return parent::findOrFail($id);
    }
```

glamorous:
@pablorsk Did you already extend your default Eloquent model like I suggested? What kind of level do you have in PHP Stan? You second "ugly" solution is not that bad because your IDE will give you better support then because he knows it's a Document-model.

mfn:
> You second "ugly" solution is not that bad because your IDE will give you better support then because he knows it's a Document-model.

Our code-base is basically littered with them to achieve this.

pablorsk:
Ok @glamorous @mfn, is not so uggly. But developers needs add `/** @var Document */` on every use of `find()`. I says uggly because developers may make a mistake in declaring the variable.

I think is a problem caused by the ease of laravel. `find()` can be return a collection or a model, and on PHP don't have generics like typescript, and is not possible set return type via #1 parameter of find() function.

Then, the best solution for now is set type on every case like `/** @var Document */`

@glamorous I'm ussing level 7 ;)

jamesaspence:
Hi guys, great job on this fix! When can us nubs expect a release with these fixes? I'd like my IDE to understand what `Route::prefix` is :)

PR #621, opened 2018-02-02
Support php's array callable format for macro's.

Zae:
Hi Barry,

We've been defining our macro's as callables of a class instead of using closures, sadly this breaks ide-helper because it expects only closures as macros.

This patch allows you to define methods like this as well: Str::macro('name', [$this, 'method']);

(This patch was based on the latest stable tag [2.4.1] because dev-master was broken for me, i'm not sure what branch you want me to push this to, if you want it on another branch so it's easier to tag please let me know)


Zae:
Hey @barryvdh,

Have you had time to look at this yet?

Greetings!

PR #629, opened 2018-02-11
"smart-reset": Reset properties/methods, but keep the text

Blizzke:
I really like to regenerate the complete set of properties and methods to keep everything sorted, but when I do that the text is removed. This adds an option somewhere in between (`-r` or `--smart-reset`) that transfers the text from the original docBlock to the new version, but discards all the rest.

### DIFF
--- a/src/Alias.php
+++ b/src/Alias.php
@@ -63,7 +63,7 @@ class Alias
         $this->detectClassType();
         $this->detectExtendsNamespace();
 
-        if(!empty($this->namespace)) {
+        if (!empty($this->namespace)) {
             //Create a DocBlock and serializer instance
             $this->phpdoc = new DocBlock(new ReflectionClass($alias), new Context($this->namespace));
         }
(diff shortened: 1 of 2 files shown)


barryvdh:
Makes sense, not sure why it would have the shortcut `r` though? Instead of `s` or something.

Should we make this default behaviour?

Blizzke:
I took `-r` because it is similar to `-R`, the regular reset, wanted to keep it the similar. 

Might be nice to have it as default behavior, as I think people would like to keep their existing descriptions.

MatanYadaev:
@Blizzke @barryvdh Any idea how to make it ignore Swagger's annotations? https://github.com/zircote/swagger-php#usage
```php
/**
 * App\Models\User.
 *
 * @OA\Schema(
 *   title="User Model",
 *   required={"id","email"},
 *   @OA\Property(property="id",type="integer",format="int32"),
 *   ......
 * )
 * @property int $id
 * ...
 */
class User
{
    ....
}
```

PR #663, opened 2018-05-15
Prevent to generate classes without methods

xakzona:
Some classes generates without methods like this
class Format extends \App\Helpers\Format {}
namespace App\Helpers { 
    class Format {
     
    }
}

But App\Helpers\Format contains methods

### DIFF
--- a/resources/views/helper.php
+++ b/resources/views/helper.php
@@ -14,7 +14,7 @@
 <?php if ($namespace == '\Illuminate\Database\Eloquent'): continue; endif; ?>
 namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> { 
 <?php foreach($aliases as $alias): ?>
-
+    <?php $methods = $alias->getMethods(); if(!count($methods)) continue; ?>
     <?= trim($alias->getDocComment('    ')) ?> 
     <?= $alias->getClassType() ?> <?= $alias->getExtendsClass() ?> {
         <?php foreach($alias->getMethods() as $method): ?>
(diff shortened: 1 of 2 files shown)


barryvdh:
So does it fail, or what is the problem?

xakzona:
I made example repo with readme how to reproduce it
https://github.com/xakzona/laravel-ide-helper-example

xakzona:
Where are any questions? I will repeat trouble here

> Check config/app.php

'Format' => \App\Helpers\Format::class

Class \App\Helpers\Format contains real method "exampleMethod"

I have commited _ide_helper.php where you can see class without any methods (but it should contain exampleMethod)

That's why than you are making your code like this \Format::exampleMethod(); IDE doesn't know about this method because class is empty (in _ide_helper.php)

xakzona:
No reasons to generate dublicate classes without any methods

msiemens:
I have a similar problem on a project that uses [Laravel Horizon](https://laravel.com/docs/5.6/horizon)

BadSantos:
I have a similar problem, please fix it ASAP

pablorsk:
> I have a similar problem on a project that uses [Laravel Horizon](https://laravel.com/docs/5.6/horizon)

I have similar problem with Horizon, Any one can fix it? I update ide-helper to 2.6.x, but the problem persist. Generate a empty class like:

```
namespace Laravel\Horizon { 

    /**
     * 
     *
     */ 
    class Horizon {
         
    }
 
}
```
I dont know the reason.. :/

PR #672, opened 2018-06-13
suppress formatter in helper file

eidng8:


eidng8:
May I ask when will these be put on packagist? It seems the last version there was a few months old.

PR #675, opened 2018-06-19
bugfix - Space/Enter between starting php tag and comment for PHPStan

glamorous:
The added comment with `//@formatter:off` isn't valid PHP code. There must be a space/enter between the php opening tag and the comment. PHPStan crashes/cries at this. This fix will make PHPStan happy again!


barryvdh:
Always nice to make PHPStan happy.

PR #691, opened 2018-08-21
Prevent loading views if `view` class not in container

vyuldashev:
If views are disabled in application an error is encountered:

```
In Container.php line 767:

  Class view does not exist
```

This pull requires will prevent loading views in `IdeHelperServiceProvider` if there is no `view` binding in application container.

### DIFF
--- a/src/IdeHelperServiceProvider.php
+++ b/src/IdeHelperServiceProvider.php
@@ -37,8 +37,10 @@ class IdeHelperServiceProvider extends ServiceProvider
      */
     public function boot()
     {
-        $viewPath = __DIR__.'/../resources/views';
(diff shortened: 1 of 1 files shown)


barryvdh:
I think the view is required, what is your use-case to have it not enabled?

vyuldashev:
I disable ViewServiceProvider in some of my projects. For example, in my bot application it is disabled (as it is not needed): https://github.com/fondbot/fondbot/blob/master/config/app.php#L154

barryvdh:
Yeah but this package actually needs it. So perhaps it should register/boot it instead then.

Hunternnm:
@barryvdh If the application only api then this change is necessary

PR #701, opened 2018-09-07
Add ::query support

### LINKED ISSUE #692: `::query()` is not aware of its model
When, for example, using custom scopes, this will break autocompletion:

```php
Number::query()
    ->whereType(NumberType::OWNER)
    ->whereResidence($owner->residence)
    ->first();
```

PHPStorm is unaware of which model `::query()` returns. It's probably fixable? I'm more than willing to send a PR if you can point me in the right direction.

brendt:
This PR adds support for `::query` and comparable methods. It fixes #692.

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -409,7 +409,6 @@ class ModelsCommand extends Command
     protected function getPropertiesFromMethods($model)
     {
         $methods = get_class_methods($model);
-
         if ($methods) {
             sort($methods);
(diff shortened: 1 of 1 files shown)


barryvdh:
Should this include `static`?

barryvdh:
I see we don't that in the other places, so guess it's fine :)

brendt:
It is generated by default with `static`, I assumed it was the default behaviour. Thanks for merging, let me know if anything still has to change!

PR #702, opened 2018-09-07
 Remove DocBlocks for classes. It will Cause conflict in IDE.

mitoop:
The DocBlocks in _ide_helper.php will cause conflict and the generated IDE helper classes will not work.


pahan35:
@barryvdh could you please merge this? ATM I need to do this changes in every my project to fix the problem #704

MaXal:
We've fixed the issue on our side. The fix will be available in the next week EAP but duplicated entries are still better to be eliminated.

mitoop:
In my local env, I remvoe the `    <?= trim($alias->getDocComment('    ')) ?> 	`  in the source code for temporarily solve.

Issue #703, opened 2018-09-09
Method with same name already defined in this class


realmarkcross:
L5.7.2, LIDEH:2.5.1 
_ide_helper.php produces error in PHPStorm :  'Method with same name already defined in this class'
on many methods.  Never had an issue with 5.6, but this is the first time I'm installing both 5.7 Laravel and 2.5.1 LIDEH.
Screenshot shows the first part of the file with underlined methods that reflect this error.
![image](https://user-images.githubusercontent.com/10559427/45260596-a4ccb180-b3b9-11e8-82be-528566f8496b.png)

guice:
+1 on this. I can't here with this same issue.

yutarux:
Me, three.

Laravel 5.5
Intellij IDEA Ultimate 2018.2.3
ide-helper 2.5.0

yutarux:
Same as #697 ?

guice:
@yutarux No. Our files create successfully. We don't get an error during compose update. Our error is within PhpStorm, complaining about pre-existing methods.

Evrimedont:
+1

Laravel 5.7
PHPStorm 2018.2.3
ide-helper 2.5.1

SeagleDub:
+1

Laravel 5.7
PHPStorm 2018.2.3
ide-helper 2.5.1

tnorthcutt:
Another one here; confirmed that reverting to `2.4.3` fixed it. Not sure what the issue is yet, but [here](https://github.com/barryvdh/laravel-ide-helper/compare/v2.4.3...master) are the commits since `2.4.3`.

tnorthcutt:
Update: the bug appears to be introduced in this commit: https://github.com/barryvdh/laravel-ide-helper/commit/b644e48801edba4ac56f1dee561b72d67f731e06

/cc @barryvdh

guyaloni:
+1

Laravel 5.7.6
PHPStorm 2017.3.6

hcancelik:
+1

Laravel 5.7.9
PHPStorm 2018.2.5

iAxel:
+1

Laravel 5.7.6 
PHPStorm 2017.3.6

JeppeKnockaert:
I've proposed a fix for this issue in #720.

iAxel:
> I've proposed a fix for this issue in #720.

Please explain step by step how to solve this problem...
Grateful...

JeppeKnockaert:
[...]

PR #720, opened 2018-10-23
Prevent duplicate function declarations in docblocks

JeppeKnockaert:
When a method is declared both in the phpdoc and in the class itself,
PhpStorm (2018.3) complains about duplicate function declarations, preventing
correct code hinting.

To solve this, I remove the already declared functions from the generated dockblocks.

This solves issue #703 and is an improvement on PR #618.
The solution does require a 

Examples:

In DB all `@method` tags have been deleted.
screenshot 2018-10-23 at 15 59 10

In Route, some `@method` tags remain.
screenshot 2018-10-23 at 15 59 27


composer.json
```diff
@@ -14,7 +14,7 @@
         "illuminate/support": "^5.5,<5.8",
         "illuminate/console": "^5.5,<5.8",
         "illuminate/filesystem": "^5.5,<5.8",
-        "barryvdh/reflection-docblock": "^2.0.4",
+        "barryvdh/reflection-docblock": "dev-master#9b08619",
```
JeppeKnockaert: This needs to become a new version of reflection-doclock

rjpounder:
I get this when I try this commit - 

In Alias.php line 421:
                                                                      
  Call to undefined method Barryvdh\Reflection\DocBlock::deleteTag()

composer.json
barryvdh: Is tagged as 2.0.5

JeppeKnockaert: Thanks, I updated it!

PR #731, opened 2018-12-04
fix: Helper has syntax error when real method is closure.

HikaruYasuda:
Thanks for the helpful package!

I'd like to fix syntax errors that occur in _ide_helper when real method is closure.

current: (syntax error)
```php
return \Illuminate\Http\Request::Illuminate\Foundation\Providers\{closure}($rules, $params);
```

want this: (undefined method warning)
```php
return \Illuminate\Http\Request::closure($rules, $params);
```

### DIFF
--- a/src/Method.php
+++ b/src/Method.php
@@ -48,6 +48,10 @@ class Method
         $this->interfaces = $interfaces;
         $this->name = $methodName ?: $method->name;
         $this->real_name = $method->name;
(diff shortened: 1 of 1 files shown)


barryvdh:
Is this already fixed with https://github.com/barryvdh/laravel-ide-helper/pull/736 ?

HikaruYasuda:
@barryvdh 
I thought that # 736 is more reliable than this PR.
Thank you for the confirmation!

PR #733, opened 2018-12-04
Resolve how to call the original methods.

HikaruYasuda:
Thanks for the useful package and maintenance.

When calling the original method, I would like to use static call or instance method call, as appropriate.
Also, want to resolve the return type when it was $this.

current:
("Non-static method 'withGlobalScope' should not be called statically" warning in PHPStorm)
("Return value is expected to be 'Eloquent', 'Illuminate\Database\Eloquent\Builder' returned" warning in PHPStorm)
```php
/**
 * Register a new global scope.
 *
 * @param string $identifier
 * @param \Illuminate\Database\Eloquent\Scope|\Closure $scope
 * @return $this
 * @static 
 */ 
public static function withGlobalScope($identifier, $scope)
{
    return \Illuminate\Database\Eloquent\Builder::withGlobalScope($identifier, $scope);
}
```

Idea
```php
/**
 * Register a new global scope.
 *
 * @param string $identifier
 * @param \Illuminate\Database\Eloquent\Scope|\Closure $scope
 * @return \Illuminate\Database\Eloquent\Builder 
 * @static 
 */ 
public static function withGlobalScope($identifier, $scope)
{
    return $instance->withGlobalScope($identifier, $scope);
}
```

### DIFF
--- a/resources/views/helper.php
+++ b/resources/views/helper.php
@@ -36,7 +36,7 @@ namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> {
             //Method inherited from <?= $method->getDeclaringClass() ?>
             <?php endif; ?>
 
-            <?= $method->shouldReturn() ? 'return ': '' ?><?= $method->getRoot() ?>::<?= $method->getRealName() ?>(<?= $method->getParams() ?>);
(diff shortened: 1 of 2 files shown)


HikaruYasuda:
goto #735

Issue #734, opened 2018-12-06
Invalid _ide_helper.php generated after upgrade to v2.5.2


mikepsinn:
```
        /**
         * 
         *
         * @static 
         */ 
        public static function validate($rules, $params = null)
        {
            return \Illuminate\Http\Request::Illuminate\Foundation\Providers\{closure}($rules, $params);
        }
        
        /**
         * 
         *
         * @static 
         */ 
        public static function hasValidSignature($absolute = true)
        {
            return \Illuminate\Http\Request::Illuminate\Foundation\Providers\{closure}($absolute);
        }
         
```

I use:
"barryvdh/laravel-ide-helper": "2.5.2",
"laravel/framework": "5.7.16"

I can verify that this bug was introduced in https://github.com/barryvdh/laravel-ide-helper/commit/981ff45b43e0cf808af0a5a5f40f6369e0e29499 as 2.5.1 works.  

(However, 2.5.1 has the "already defined" warnings in PHPStorm.)

xlstefan:
@mikepsinn can you test this with the v2.5.3?
I think this issue is solved.

mikepsinn:
@xlstefan Yes, it works great now!  Thanks so much!  :D

Issue #726, opened 2018-11-15
version 2.5.2 + latest laravel 5.7.13 generates code which is not valid for phpstan


victor-ivanov:
I've created new project with
`composer create-project laravel/laravel blog`

added latest `ide-helper`
```
cd blog
composer require barryvdh/laravel-ide-helper
```
getting following code generated for `Request` facade `validate` and `hasValidSignature` methods
```
/**
 * 
 *
 * @static 
 */ 
public static function validate($rules, $params = null)
{
    return \Illuminate\Http\Request::Illuminate\Foundation\Providers\{closure}($rules, $params);
}

/**
 * 
 *
 * @static 
 */ 
public static function hasValidSignature($absolute = true)
{
    return \Illuminate\Http\Request::Illuminate\Foundation\Providers\{closure}($absolute);
}
```

phpstan fails to validate such code with error 
```
PHP Parse error:  syntax error, unexpected '\' (T_NS_SEPARATOR), expecting ';' in .../_ide_helper.php
```

**Workaround**
It generates proper code with previous version of `laravel-ide-helper` `2.5.1`

```
public static function validate($rules, $params = null)
{
    return \Illuminate\Http\Request::validate($rules, $params);
}
public static function hasValidSignature($absolute = true)
{
    return \Illuminate\Http\Request::hasValidSignature($absolute);
}
```

STayinloves:
Shared the same issue using ide-helper on dev-master

adampatterson:
I just noticed that I can't click through on route class names to methods, or to the view file.

Also using IDE Helper version 2.5.2, latest laravel 5.7.13

I had previously tried IDE Helper 2.4.* and updated because it was not working.

ondrejmirtes:
I recommend switching to Larastan.

xlstefan:
@victor-ivanov can you test this with the v2.5.3?
I think this issue could be solved with new release.

victor-ivanov:
Thanks guys! version `2.5.3` works for me

PR #736, opened 2018-12-10
bugfix - Do not use the real_name for closures

JeppeKnockaert:
In https://github.com/barryvdh/laravel-ide-helper/commit/5d8d93273f67797f61b92cc32690d058b6c6b4d9, an issue was introduced causing syntax errors to appear for methods defined in Macro's.

See issues: https://github.com/barryvdh/laravel-ide-helper/issues/726, https://github.com/barryvdh/laravel-ide-helper/issues/734

This is caused by trying to retrieve the real method name of an anonymous function (a closure).  Closures do not have a method name that makes sense to use in the ide_helper
file. e.g. `Illuminate\Foundation\Providers\{closure}` would be the real_name instead of
just `validate`. So for the closure methods, we should just use the provided
methodName instead.

### DIFF
--- a/src/Method.php
+++ b/src/Method.php
@@ -47,7 +47,7 @@ class Method
         $this->method = $method;
         $this->interfaces = $interfaces;
         $this->name = $methodName ?: $method->name;
-        $this->real_name = $method->name;
+        $this->real_name = $method->isClosure() ? $this->name : $method->name;
         $this->initClassDefinedProperties($method, $class);
 
         //Create a DocBlock and serializer instance


xlstefan:
Can this solve an issue with Laravel 5.7 version and generation of bad ide_helper.php?

JeppeKnockaert:
Possibly, we're using Laravel 5.7 and this solves the syntax errors for us.

xlstefan:
This solved most of the syntax errors!
I created new issue with more details above.

Thanks @JeppeKnockaert !

Issue #739, opened 2018-12-11
Syntax errors in _ide_helper.php


xlstefan:
This pull request (https://github.com/barryvdh/laravel-ide-helper/pull/736) fixed most of the syntax errors but not all.
There are still syntax errors bellow you can see details.
Commit hash: https://github.com/barryvdh/laravel-ide-helper/commit/99226c92b72b02bf60d9b339fdf9f446d5d9921d
Possible related issues: #734 , #726 

Error message: **Method with this name already defined in this class**
Only in these 4 methods it's shown.

Some more details about packages which I am using:

>     "require": {
>         "php": "^7.1.3",
>         "backpack/crud": "^3.5",
>         "bensampo/laravel-enum": "^1.16",
>         "beyondcode/laravel-self-diagnosis": "1.0.2",
>         "doctrine/dbal": "^2.5",
>         "fideloper/proxy": "^4.0",
>         "laravel/framework": "5.7.*",
>         "laravel/horizon": "^1.4",
>         "laravel/telescope": "^1.0",
>         "laravel/tinker": "^1.0",
>         "laravelcollective/html": "^5.4.0",
>         "predis/predis": "^1.1",
>         "ext-json": "*"
>     },
>     "require-dev": {
>         "backpack/generators": "^1.2",
>         "barryvdh/laravel-debugbar": "^3.2",
>         "barryvdh/laravel-ide-helper": "dev-master#99226c9",
>         "filp/whoops": "^2.0",
>         "fzaninotto/faker": "^1.4",
>         "laracasts/generators": "dev-master",
>         "laravel/homestead": "v7.12.0",
>         "mockery/mockery": "^1.0",
>         "nunomaduro/collision": "^2.0",
>         "phpunit/phpunit": "^7.0"
>     }

```
         /**
         * Get an array of all files in a directory.
         *
         * @param string $directory
         * @param bool $hidden
         * @return \Symfony\Component\Finder\SplFileInfo[] 
         * @static 
         */ 
        public static function files($directory, $hidden = false)
        {
            return \Illuminate\Filesystem\Filesystem::files($directory, $hidden);
        }
        
        /**
         * Get all of the files from the given directory (recursive).
         *
         * @param string $directory
         * @param bool $hidden
         * @return \Symfony\Component\Finder\SplFileInfo[] 
         * @static 
         */ 
        public static function allFiles($directory, $hidden = false)
        {
[...]

PR #740, opened 2018-12-12
bugfix - Prevent redeclaring methods inside Aliases

JeppeKnockaert:
This pull request fixes issue https://github.com/barryvdh/laravel-ide-helper/issues/739.

The getMethods function gets called both in the removeDuplicateMethodsFromPhpDoc
function and in helper.php, which causes the method detection to be performed twice.

Parts of the method detection includes a check for duplicate methods, but Macro
methods for example, aren't checked for duplicates.

Instead of checking everywhere if we're not adding duplicates, I've included a
check in the getMethods call to just use the already generated methods
if they have been generated already.

### DIFF
--- a/src/Alias.php
+++ b/src/Alias.php
@@ -176,6 +176,10 @@ class Alias
      */
     public function getMethods()
     {
+        if (count($this->methods) > 0) {
+            return $this->methods;
+        }
+
         $this->addMagicMethods();
         $this->detectMethods();
         return $this->methods;
@@ -411,10 +415,9 @@ class Alias
      */
     protected function removeDuplicateMethodsFromPhpDoc()
     {
-        $methods = count($this->methods) > 0 ? $this->methods : $this->getMethods();
         $methodNames = array_map(function (Method $method) {
             return $method->getName();
-        }, $methods);
+        }, $this->getMethods());
 
         foreach ($this->phpdoc->getTags() as $tag) {
             if ($tag instanceof MethodTag && in_array($tag->getMethodName(), $methodNames)) {


xlstefan:
Hi @JeppeKnockaert,

I just checked, this PR solved syntax errors for methods 3 and 4 (from #739 ).
But syntax errors for methods 1 and 2 are still there:

        `/**
         * Get an array of all files in a directory.
         *
         * @param string $directory
         * @param bool $hidden
         * @return \Symfony\Component\Finder\SplFileInfo[] 
         * @static 
         */ 
        public static function files($directory, $hidden = false)
        {
            return \Illuminate\Filesystem\Filesystem::files($directory, $hidden);
        }
        
        /**
         * Get all of the files from the given directory (recursive).
         *
         * @param string $directory
         * @param bool $hidden
         * @return \Symfony\Component\Finder\SplFileInfo[] 
         * @static 
         */ 
        public static function allFiles($directory, $hidden = false)
        {
            return \Illuminate\Filesystem\Filesystem::allFiles($directory, $hidden);
        }`

JeppeKnockaert:
Just for anyone who passes here, the issue mentioned by @xlstefan is something that needs to be addressed in https://github.com/barryvdh/ReflectionDocBlock (see #739 for more info).

JeppeKnockaert:
@barryvdh If you make a version of ReflectionDocBlock, I'll include a commit to bump it here so all fixes for #739 are solved in one PR.

barryvdh:
Tagged v2.0.6

PR #745, opened 2018-12-21
Include DocBloc’s from classes to allow deeper code inspection

mrtimp:
Currently if a class has magic methods and the class include type hints they aren't resolved by the ide-helper package.

For example before:

screenshot_258

and after:

screenshot_259

My proposed PR simply inspects classes, includes their DocBlocks to add in any comments and PhpDoc including @method's to allow type hinting.

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -193,4 +193,16 @@ return array(
         'integer' => 'int',
         'boolean' => 'bool',
    ),
+
+    /*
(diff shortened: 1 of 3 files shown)


mrtimp:
I should add this enabling this is optional via the config file, I figure if people feel this should be the default behaviour we could enable it by default in a future version of the ide-helper.

Issue #738, opened 2018-12-11
PhpStorm, IDE helper does not recognize model methods


Dunrar:
The IDE helper does not recognize method calls on models. For example, if I get a user via find or fail, like 
`$employee = User::findOrFail($id);` and then call a Method, like `$expiringQualifications = $employee->expiringQualifications()->pluck('name', 'id');`, the IDE helper does not recognize the expiringQualifications() Method. It tells me:

> Method 'expiringQualifications' not found in \Illuminate\Database\Eloquent\Collection|\Illuminate\Database\Eloquent\Model

MarkWisselink:
had the same issue; fixed it by replacing the class definition of 'Eloquent extends \Illuminate\Database\Eloquent\Model' with 'class Eloquent extends \Illuminate\Database\Eloquent\Model {}'
into it's own namespace, and then moving the old definition inside a new namespace block like:
`namespace Illuminate\Database\Eloquent {
    class Model {
        //the class definition
    }
}`
obviously this is a static fix; and should be fixed properly in the parser

mfn:
Works for me in PhpStorm with the latest version:

![image](https://user-images.githubusercontent.com/87493/71750230-19addf80-2e78-11ea-9cb7-fc382cf42305.png)

`findOrFail` is generated as:
```php
            /**
             * Find a model by its primary key or throw an exception.
             *
             * @param mixed $id
             * @param array $columns
             * @return \Illuminate\Database\Eloquent\Model|\Illuminate\Database\Eloquent\Collection|static|static[] 
             * @throws \Illuminate\Database\Eloquent\ModelNotFoundException
             * @static 
             */ 
            public static function findOrFail($id, $columns = [])
```

The `static` makes this work.

PR #751, opened 2019-01-04
Update helper.php

Numynum:
should fix https://github.com/barryvdh/laravel-ide-helper/issues/738
(tested on netbeans)


mfn:
I don't understand how this fixes #738, see my comment there ( https://github.com/barryvdh/laravel-ide-helper/issues/738#issuecomment-570705659 ) as it works for me.

Numynum:
if I remember correctly (it's been a year :)), it's because my models don't extend \Eloquent (in root namespace); and the IDE helper as it is now, only works if you do
this PR moves the model back into the namespace, and if I remember correctly, makes \Eloquent extend the namespaced model class
and in that way, whether your actual model extends \Eloquent, or \Illuminate\Database\Eloquent\Model doesn't matter, both will work

for sure correct me if I'm wrong, as I'm not a laravel expert at all yet, so I might be missing something obvious

mfn:
> it's been a year :)

😄 



> and the IDE helper as it is now, only works if you do

This does not strike me as the current behaviour.

None of my models extend `\Eloquent`, all extend `\Illuminate\Database\Eloquent\Model`

The phpdoc for everything gets generated correctly in each model.

See also the tests recently added, they all work with the models directly.

Do you happen to use a specific set of configurations to trigger your problem?

I really suggest to try this again with the current version. I would rather understand the nature of the fix better before proceeding here.

Numynum:
I'll check it out after the weekend, I don't have a good setup to test with where I am now :)

keep in mind I've only been using the IDE helper file, and not any docblock generation for the models themselves
do the generated docblocks `@mixin  \Eloquent` maybe? because it doesn't make sense to me otherwise

when you check out just the (pre-)generated IDE helper file, there is no `\Illuminate\Database\Eloquent\Model` class defined there, only `\Eloquent` :thinking:

mfn:
@Numynum what exact steps are you using, which commands you run with which arguments and which files does generate for you?
Thanks

Numynum:
all I do is `php artisan ide-helper:generate`
which generates a file called `_ide_helper.php`

after some more testing, it seems to be the case that netbeans 8.2 doesn't support `@mixin` docblocks; which means that even if I used `php artisan ide-helper:generate --write_mixins` or similar to add `@mixin \Eloquent` to all my models, it would not work for netbeans users

this PR changes the FQCN for models (!only in the _ide_helper file!) to `\Illuminate\Database\Eloquent\Model` instead of `\Eloquent`
which means that mixing in \Eloquent is no longer needed

mfn:
>  netbeans 8.2

That version was released in 2017 and is already superseded…
…also Netbeans doesn't strike as a "popular" choice and this change could negatively affect the more popular ones in a negative way.

I vote for not pursuing this PR and rather would close it, sorry.

Numynum:
I know there's a newer version, I haven't checked if it supports mixin now

just curious, what negative impact would there be?
there's an exception in the current code for eloquent models, so there must've been some reason for it :thinking: 
feel free to close the PR in any case (didn't know about the mixin when I was making it), at this point I'm just curious :)

mfn:
@barryvdh I vote for dropping the PR, we already decided it to close the referenced issue :}

PR #758, opened 2019-01-25
Add support for factory()->create() return types

inxilpro:
Right now, PhpStorm does not know what the return type of `factory(\App\User::class)->create()` is. To solve this, this PR adds two generators (when `include_factory_builders` is set to `true`):

First, the `generate` command will create "faux" FactoryBuilder classes, like:

```php
namespace App {
    /**
     * @method \Illuminate\Database\Eloquent\Collection|User[]|User create($attributes = [])
     * @method \Illuminate\Database\Eloquent\Collection|User[]|User make($attributes = [])
     */
    class UserFactoryBuilder extends \Illuminate\Database\Eloquent\FactoryBuilder {}
}
```

Second, the `meta` command will override the `factory()` method to return the associated builder class:

```php
override(\factory(0), map([
    '' => '@FactoryBuilder',
    'App\User' => \App\UserFactoryBuilder::class,
]));
```

Now, PhpStorm will think that the return type of `factory(\App\User::class)` is  `\App\UserFactoryBuilder`. Calls to `create()` or `make()` will correctly autocomplete all the methods/properties on the `\App\User` class.

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -1,6 +1,6 @@
 <?php
 
-return array(
+return [
 
     /*
     |--------------------------------------------------------------------------
@@ -11,9 +11,9 @@ return array(
     |
     */
 
-    'filename'  => '_ide_helper',
-    'format'    => 'php',
-    
+    'filename' => '_ide_helper',
+    'format'   => 'php',
+
     'meta_filename' => '.phpstorm.meta.php',
 
     /*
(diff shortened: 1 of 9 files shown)


satoved:
Would be great to have!

mikerogne:
👍 Would love this!

inxilpro:
I have this functionality in [laravel-faux-generics](https://github.com/glhd/laravel-faux-generics) — the package isn’t fully tested or documented, but works.

barryvdh:
Why was the detectFake function removed?

splinter89:
@inxilpro Please make this happen :pray:

inxilpro:
Alright, fixed. @barryvdh I submitted the `detectFake` PR and this one around the same time, and that functionality got merged but this one didn't. I must've explicitly removed that code from the PR at the time so that the diff was minimal, but that meant that the change happened after the detect fake code was added. We should be all good now.

inxilpro:
Hrm. There's a PHPCS error on a line I didn't touch. I'm going to fix it…

PR #761, opened 2019-01-29
Fixed #760

### LINKED ISSUE #760: Relations with query not work
We have this class:

```
<?php

namespace App;

use App\Comment;
use Illuminate\Database\Eloquent\Relations\HasMany;

class Post extends Model
{
    public function comments(): HasMany
    {
        return $this->hasMany(Comment::class);
    }
    
    public function commentsUnread(): HasMany
    {
        return $this->comments()->where('unread', true);
    }
}
```

After run ```php artisan ide-helper:models``` we got: 
```
/**
 * @property-read \Illuminate\Database\Eloquent\Collection|\App\Comment[] $comments
 */
```

But expected:
```
/**
 * @property-read \Illuminate\Database\Eloquent\Collection|\App\Comment[] $comments
 * @property-read \Illuminate\Database\Eloquent\Collection|\App\Comment[] $commentsUnread
 */
```

Seems like because ide-helper just search functions with relations [names](https://github.com/barryvdh/laravel-ide-helper/blob/master/src/Console/ModelsCommand.php#L468) and ignore method return type.

Anyone have workaround for this?

UksusoFF:
Because you [drop support](https://github.com/barryvdh/laravel-ide-helper/commit/09db8c9a76711e98c61af0795934fb15955223fb#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780) php < 7 we can use [ReflectionFunctionAbstract::getReturnType](http://php.net/manual/en/reflectionfunctionabstract.getreturntype.php).

This PR tested for relations with return type and without.

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -450,81 +450,73 @@ class ModelsCommand extends Command
                 } elseif (!method_exists('Illuminate\Database\Eloquent\Model', $method)
                     && !Str::startsWith($method, 'get')
                 ) {
-                    //Use reflection to inspect the code, based on Illuminate/Support/SerializableClosure.php
                     $reflection = new \ReflectionMethod($model, $method);
 
-                    $file = new \SplFileObject($reflection->getFileName());
-                    $file->seek($reflection->getStartLine() - 1);
-
-                    $code = '';
-                    while ($file->key() < $reflection->getEndLine()) {
-                        $code .= $file->current();
-                        $file->next();
+                    if ($reflection->getNumberOfParameters()) {
+                        continue;
                     }
-                    $code = trim(preg_replace('/\s\s+/', '', $code));
-                    $begin = strpos($code, 'function(');
-                    $code = substr($code, $begin, strrpos($code, '}') - $begin + 1);
 
-                    foreach (array(
-                               'hasMany',
-                               'hasManyThrough',
-                               'belongsToMany',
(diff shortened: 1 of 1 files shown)


src/Console/ModelsCommand.php
```diff
@@ -450,81 +450,73 @@ protected function getPropertiesFromMethods($model)
...
+                    }
 
-                    $code = '';
-                    while ($file->key() < $reflection->getEndLine()) {
-                        $code .= $file->current();
-                        $file->next();
+                    $returnType = $reflection->getReturnType();
+
+                    if ($returnType === null) {
+                        $relationObj = $model->$method();
```
edcoreweb: You might trigger some unwanted behaviour here. You are basically calling every method on the model. (Imagine you have a method which sends a mail or something.)

edcoreweb:
> Yeah perhaps use the return type if possible, otherwise old behavior.

I did a minimal implementation of this issue as well: https://github.com/barryvdh/laravel-ide-helper/pull/763

barryvdh:
Merged the other one, thanks!

UksusoFF:
> You might trigger some unwanted behaviour here. You are basically calling every method on the model. (Imagine you have a method which sends a mail or something.)
@edcoreweb I don't really worried about this on development environment (laravel-ide-helper should be used only on dev right?).
Seems like PR #763 maybe more safety for any other.

PR #763, opened 2019-02-03
Nested eloquent relations detection.

edcoreweb:
In my applications, sometimes I have the following scenario:

```php
/**
 * @return \Illuminate\Database\Eloquent\Relations\HasMany
 */
public function tags()
{
    return $this->hasMany(Tag::class);
}

/**
 * @return \Illuminate\Database\Eloquent\Relations\HasMany
 */
public function redTags()
{
    return $this->tags()->where('color', 'red');
}
```

Currently `redTags` is not included in the generated output.

This pull request addresses that by checking the return type of the functions when scanning for relations on a model. (both native return type and fallback in the `@return` docblock)

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -452,6 +452,8 @@ class ModelsCommand extends Command
                 ) {
                     //Use reflection to inspect the code, based on Illuminate/Support/SerializableClosure.php
                     $reflection = new \ReflectionMethod($model, $method);
+                    // php 7.x type or fallback to docblock
+                    $type = (string) $reflection->getReturnType() ?: $this->getReturnTypeFromDocBlock($reflection);
 
                     $file = new \SplFileObject($reflection->getFileName());
                     $file->seek($reflection->getStartLine() - 1);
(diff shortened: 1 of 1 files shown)


emielmolenaar:
Doesn't this break stuff like

```
public function relation(): HasOne
{
  return $this->hasOne(SomeModel::class, 'foreign_key', 'key');
}

public function anotherRelation(): HasOne
{
  return $this->relation->anotherHasOne();
}
```

`$this->relation` is null when ide-helper is trying to figure out the nested relation.

edcoreweb:
@emielmolenaar 

The way you do it is not recommended, and I don't think it should be covered.

Notice the difference between:
```php
return $this->relation()->anotherHasOne();
```
And:
```php
return $this->relation->anotherHasOne();
```

The second one actually fetches the relation from the database. (If you eager load 5000 entities with this relation it will do at least 5000 queries)

IMHO: The idea of a relation function is to actually return a relation and not fetch data.

Just out of curiosity why do you need to do it like that ?

emielmolenaar:
@edcoreweb Thanks for respondig. I understand the difference between `relation` and `relation()` and I agree with you for a large part. Your first example will not work by the way, the method `anotherHasOne` doest not exist one the `HasOne` class :smiley: 

We have some convenience methods that return a relation based on some conditions. For example:

```
public function getRelation(): HasOne
{
   if($this->anotherRelation->isSomeValue()) {
      return $this->someHasOneRelation()->where('some_field', 'some_value')->where('another_field', 'another_value');
   }

   return $this->someHasOneRelation()->where('yet_another_field', 'some_value');
}
```

In this case `anotherRelation` has already been loaded (we are very aware of the N + 1 query problem). We sometimes use this kind of methods to prevent duplicate lines of code.

edcoreweb:
@emielmolenaar 

I see, I usually use scopes for this. The code as it is now needs to call the relation function using reflection (to get the related model), so as long as the relation function can run without context, you should be fine.

The easiest way to solve this in your case would be to check for the relation existence:

```php
if ($this->anotherRelation && $this->anotherRelation->isSomeValue()) {
    // do stuff
}
```

or use the **Null Object pattern** and make a default model, and use it without the extra check

```php
public function anotherRelation()
{
    return $this->belongsTo(SomeModel::class)->withDefault();
}
```

PR #764, opened 2019-02-03
New --write-mixin option.

edcoreweb:
For model auto-completion there are 2 options now.

+ Adding the doc blocks on each model
+ Adding the doc blocks in ` _ide_helper_models.php`

The main cons for these 2 are (a lot of comments for the first option, duplicate class notices for the second)

This pull request proposes a third combined option, where the ` _ide_helper_models.php` file is generated but with a prefixed class name.

For each model a simple `@mixin` tag is added, pointing to the helper file prefixed class.

This way the committed comments/changes are minimal and duplicate class notices disappear.


edcoreweb:
@barryvdh Is this a chance this will be merged ? I will delete it otherwise. Thanks.

mfn:
I've mixed feelings about this addition.

Ok, the good thing is: it's totally optional.

But it really diminishes the usefulness of what the package provides.

Two examples:

1. if you've the phpdoc in your model, it's automatically part of your VCS, which means any change (and so also: erroneous change) can be easily spotted.
  Example: you add a new `JSONB` column but forgot to add it do $casts. At least when you review your changes, it's noticable that the proper is documented as e.g. `string|null` when it should be `array|null`
2. static source code analyzers: they can't work with the `_ide_helper_models.php` file but with the model phpdoc

edcoreweb:
@mfn Yes, I see your point, usually I don't want my models filled with the types to the brink, but I like the auto-complete.

I commit my _ide_helper_models.php files and update them from time to time when the package gets updates. This way I still have the changes in my VCS but I know it's only a doc change and not a code change by looking at the file name.

As for static analyzers did not use those before, as I don't really write a lot of code in strict type mode. For me this package provides only one thing: auto-complete.

If this is not wanted in the package, that's not a problem at all. For my purposes I already forked the repo and use that. Thought that maybe this might be a middle ground solution for the duplicate class warnings as well, which is a problem with separate files.

mfn:
> and update them from time to time when the package gets updates

Hmm, that feels wrong. Your annotations should be updated every time you change them (migrations: i.e. add/change/delete columns, or models: add/change/delete casts, etc.), not just when you update the package 🙀 

> duplicate class warnings as well, which is a problem with separate files.

Yeah, that's also rather annoying.

I'm not sure why you so much insist on nothing having the phpdoc in the files. If you consider not having them and a (new?) team dev looks at your model: literally no idea what attributes are supposed to be there.

But: no need to argue about your workflow, absolutely respect that 👍 

My "IMHO" is, this isn't necessarily a great fit for _this- library and people in general are in favour of the docs. And since you found a working solution for yourself, I vote for not merging this.
Just my 2c.

edcoreweb:
> add/change/delete columns. 

I meant I update the fork, of course the doc is updated with my models :D 

Anyway feel free to close this if it will not be merged, or let me know if you can't and I'll do it. Thanks for the input!

src/Console/ModelsCommand.php
```diff
@@ -151,6 +154,7 @@ protected function getOptions()
           array('filename', 'F', InputOption::VALUE_OPTIONAL, 'The path to the helper file', $this->filename),
           array('dir', 'D', InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY, 'The model dir', array()),
           array('write', 'W', InputOption::VALUE_NONE, 'Write to Model file'),
+          array('write-mixin', 'M', InputOption::VALUE_NONE, 'Write mixin to Model file'),
```
mfn, line 157: (untested) but I think it makes sens to be clear how this works
```suggestion
          array('write-mixin', 'M', InputOption::VALUE_NONE, 'Write models to {$this->filename} and adds @mixin to each model, avoiding IDE duplicate declaration warnings'),
```

edcoreweb:
@mfn Hi, sorry been busy these days, if it's going to get merged, sure I can take some time to look into this.

mfn:
@decadence please wait for @barryvdh to give his ok if he wants to merge this :)

ParogDev:
Just chiming in to say this looks like an amazing solution. Looking forward to seeing this merged. Thank you so much for taking the time to solve this for those of us that don't like suppressing errors and like adhering to standards!

mfn:
ping @barryvdh

JoshStaff:
This would be very handy for me/us! Thanks @edcoreweb and contributors :D

Some communication from barryvdh would be great, given it has been radio silence for >a year from him.

barryvdh:
I'm not really sure why you wouldn't want to write it to the model files, but I guess this could work.
Can you add a test for this?

mfn:
> I'm not really sure why you wouldn't want to write it to the model files,

I'm in the same boat btw. If yet to see a reasonable for people to want to separate the annotation for the models from them. Some _then_ even want to version them. Separately. That's a recipe for things drifting apart.

Either way, this PR isn't about this and I think "the other camp" will truly welcome this.

src/Console/ModelsCommand.php
```diff
@@ -810,12 +814,13 @@ protected function createPhpDocs($class)
 
 
         if ($this->write) {
+            $modelDocComment = $this->write_mixin ? "/**\n * @mixin IdeHelper{$classname}\n */" : $docComment;
```
barryvdh, line 817: This will replace the existing attributes, right? Not sure if we want that?

edcoreweb: You are correct, I did not think about this, we should keep the exisiting ones, will look into a way to do that.

edcoreweb:
> If so, please:
> 
> * will it work if you've models with the same class name in different namespaces? e.g. `App\Models\Post` and `App\MoarModels\Post`, etc.?

It will work with the same name in different namespaces.  Considering your example the _ide_helper_models.php file will still have the original namespaces:

```php
namespace App\Models\Post {
    class IdeHelperPost extends \Eloquent {}
}

namespace App\MoarModels\Post {
    class IdeHelperPost extends \Eloquent {}
}
```

And in the models, if only the name of the class is set, it will first look into the current namespace, which is what we want.

edcoreweb:
@mfn @edcoreweb Please take a look, it's done from my part.

mfn:
Thanks for the ping 😏 I haven't forgotten, I'm just quite busy ATM

edcoreweb:
> Thanks for the ping 😏 I haven't forgotten, I'm just quite busy ATM

No problem, no rush from me.

emargareten:
@edcoreweb can we have this as a config option as well?

README.md
```diff
@@ -121,6 +121,9 @@ By default, you are asked to overwrite or write to a separate file (`_ide_helper
 You can write the comments directly to your Model file, using the `--write (-W)` option, or
 force to not write with `--nowrite (-N)`.
 
+Alternatively using the `--write-mixin (-M)` option will only add a mixin tag to your Model file,
+writing the rest in (`_ide_helper_models.php`).
+
```
mfn, line 126: I think the README fails to "sell" the feature so people understand what it does: the name different from the model, avoiding the IDE duplicate annoyance.

I fail to come up with a good description, but IMHO this super-pro should be written here 😄

src/Console/ModelsCommand.php
```diff
@@ -805,17 +812,34 @@ protected function createPhpDocs($class)
         }
 
         $serializer = new DocBlockSerializer();
-        $serializer->getDocComment($phpdoc);
```
mfn, line 808: lol, thx

mfn (approved):
Nice, IMHO looks good to me 👍 

Maybe some better "selling 😏" in the readme?

ping @barryvdh

src/Console/ModelsCommand.php
edcoreweb: Yes, initially thought it was some kind of hack, but it is just a forgotten line.

mfn (approved):
@edcoreweb I hope you don't mind but I updated your PR against master, fixed merge conflicts and adapted the recent idiomatic changes we did (codestyle fixes, test adaptions)

@barryvdh IMHO we're good to merge here!

PR #765, opened 2019-02-13
Add support for invokable classes as macro function

standaniels:
Macros can be used with invokable classes, like so.

```php
<?php

namespace App\Providers;

use Illuminate\Support\ServiceProvider;
use Illuminate\Support\Facades\Response;

class ResponseMacroServiceProvider extends ServiceProvider
{
    public function boot()
    {
        Response::macro('foo', new Foo());
    }
}

class Foo
{
    public function __invoke()
    {
        return 'foobar';
    }
}
```

When running `ide-helper:generate` the following fatal error was thrown.

>Symfony\Component\Debug\Exception\FatalThrowableError  : ReflectionFunction::__construct() expects parameter 1 to be string, object given

This commit fixes this error.

### DIFF
--- a/src/Alias.php
+++ b/src/Alias.php
@@ -395,6 +395,10 @@ class Alias
             return new \ReflectionMethod($macro_func[0], $macro_func[1]);
         }
 
+        if (is_object($macro_func) && is_callable($macro_func)) {
+            return new \ReflectionMethod($macro_func, '__invoke');
+        }
+
         return new \ReflectionFunction($macro_func);
     }
 


standaniels:
Is there anything I can do to succeed the check above?

edcoreweb:
@standaniels Nope, fails on my pull request as well.

francislavoie:
This is actually a breaking change for me. If I make a macro like this:

```php
        /**
         * @return TestResponse
         */
        TestResponse::macro('assertViewData', function($callback) {
            /** @var TestResponse $this */
            Assert::assertTrue($callback((object)$this->original->getData()));
            return $this;
        });
```

It now makes a `ReflectionMethod` instead of a `ReflectionFunction`, and it loses the doc comment. This means that the generated helpers are missing the return type, so I get warnings all over when trying to make fluent macros.

This is because `Closure` is also an object and callable, but `ReflectionFunction` should be used to pull out the doc comment.

PR #767, opened 2019-02-23
Remove register provider

ahmed-aliraqi:
You do not need to register service provider on non-production enviroment because it already registered using laravel package discovery


mfn:
Actually, I would say let's go the other way and remove the auto-discovery.

Because the auto discover means, it's always loaded. But you don't (shouldn't) need to load this in production.

As such, rather the readme should say but we should remove it from composer.json…

barryvdh:
With the --dev flag and auto Discovery, it's never loaded in production. So this looks good to me.

barryvdh:
Only it has conflicts so cannot merge this.

mfn:
> With the --dev flag and auto Discovery, it's never loaded in production. So this looks good to me.

Yeah, correct. But I still think it doesn't harm to keep it in the readme for those who don't / can't / won't use auto-discovery as additional hint what to watch out for.

PR #768, opened 2019-02-26
Add Author on models

fhferreira:
Update ModelsCommand.php to add Author through the files


mfn:
What's your use case?

mfn:
@barryvdh no feedback from author, I suggest to drop this PR 🤷‍♀️

barryvdh:
I guess this would make it easier to reset the phpdocs and keep tags, but I think this is a bit too specific.
Maybe we could reset only the docs that are added by us, not other tags.

PR #772, opened 2019-02-27
Fix php 7.3 bug

neoteknic:
fix error on line l 483: 
Exception: stripos(): Non-string needles will be interpreted as strings in the future. Use an explicit chr() call to preserve the current behavior
Could not analyze class App\Models\Client.


yupmin-ct:
👍 good!!

PR #773, opened 2019-02-28
Resolves view factory only at services' construction time

MatteoOreficeIT:
Signed-off-by: Matteo Orefice <matteo.orefice@bitia.it>

This change allows to register `Barryvdh\LaravelIdeHelper\IdeHelperServiceProvider` in any order in `app.php`, I discovered it cannot be registered before `Illuminate\Filesystem\FilesystemServiceProvider` so the documentation is not really in sync because it throws an error if we put `Barryvdh\LaravelIdeHelper\IdeHelperServiceProvider` at the top

*Based on my actual knowledge trying to resolve object via the container is not allowed before every provider's register() method has been completed*


barryvdh:
Do we even need to explicitly list the factories? We can just use DI now right?

mfn:
Probably not because of https://github.com/barryvdh/laravel-ide-helper/pull/475 
> This should safeguard from app-dedicated overrides to the View Factory used in-app.

People use dedicated view factories not support php, thus packages assuming php works (like this) break, thus a custom view is necessary.

MatteoOreficeIT (commented):
undo of createLocalViewFactory optimization that is not strictly required

MatteoOreficeIT:
> @MatteoOreficeIT do you still need the PR?

Yes, this code break the Laravel register/boot event sequence, my only remark is about the time `createLocalViewFactory()`  is called, we can build the independent VF as we need even after, but not in register event. It could also be defined with DI but I don't need to inject anything of different

MatteoOreficeIT:
Hi, I will try to fix in next days

barryvdh:
Closing due to inactivity

PR #779, opened 2019-03-22
use Throwable instead of Exception

leo108:
It may throw an error instead of an exception for some classes.

### DIFF
--- a/src/Console/MetaCommand.php
+++ b/src/Console/MetaCommand.php
@@ -90,7 +90,7 @@ class MetaCommand extends Command
                 if (is_object($concrete)) {
                     $bindings[$abstract] = get_class($concrete);
                 }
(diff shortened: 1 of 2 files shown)


mfn:
Do you have an example of what is thrown, instead?

leo108:
@mfn https://github.com/barryvdh/laravel-ide-helper/blob/master/src/Console/MetaCommand.php#L89

This line is trying to instantiate `$abstract`.

If the instance is a subclass of `Illuminate\Http\Request`, the `validator()` method will be called in its constructor method.

I may use `$this->user()->someMethod()` to get some info from current login user in `validator()` method, and `$this->user()` will return null in CLI mode, so it will throw a `Call to a member function someMethod() on null` error.

leo108:
@barryvdh Can you check this?

PR #782, opened 2019-03-26
Support for custom builder classes

brendt:
You can override the builder class used on a per model basis. This wasn't supported yet.

For reference: https://mobile.twitter.com/timacdonald87/status/1109964446766497793


brendt:
Thanks! Any idea when this will be tagged?

barryvdh:
Tagged it

PR #783, opened 2019-03-26
Chages to x2Many relationships detection & adds $relationship_countt

edvordo:
* x to Many relationships are detected via a `Many` keyword in the relationship object class name
   * reason: example: `belongsTo` was incorrectly recognized as `belongsToMany`
 * adds `@property-read int|null $x2Many_relationship_method_name_count` model class docBlock comment
   * when model makes use of `withCount` or `loadCount` there is a property on the resource `$instance->relationship_method_count` that states the count of related models
   * can be `NULL` when `withCount` or `loadCount` aren't used

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -503,13 +503,19 @@ class ModelsCommand extends Command
                                     'morphToMany',
                                     'morphedByMany',
                                 ];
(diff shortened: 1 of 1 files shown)


edvordo:
@barryvdh Anything I can do to get this merged? :)

PR #797, opened 2019-06-04
Update Model DocBlocks automatically after migrations

mortenscheel:
I've always wanted this feature myself, and with the recent addition of Migration Events in v5.8.16, it was relatively easy to accomplish.

Let me know if anything needs to be changed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,7 @@
         "illuminate/support": "^5.5,<5.9",
         "illuminate/console": "^5.5,<5.9",
         "illuminate/filesystem": "^5.5,<5.9",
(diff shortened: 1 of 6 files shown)


mfn:
This competes with https://github.com/barryvdh/laravel-ide-helper/pull/842

mfn:
FTR, this was closed in favour of https://github.com/barryvdh/laravel-ide-helper/pull/842 , see discussion there

barryvdh:
I like this approach, but would just run the command on all models probably. Seems a bit tricky for this, not sure if there is a need for that complexity.

PR #803, opened 2019-07-02
Fix phpcs

liquid207:
Fix job `phpcs -p --standard=PSR2 src/` fail in github Travis CI.

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -456,7 +456,9 @@ class ModelsCommand extends Command
                     //Use reflection to inspect the code, based on Illuminate/Support/SerializableClosure.php
                     $reflection = new \ReflectionMethod($model, $method);
                     // php 7.x type or fallback to docblock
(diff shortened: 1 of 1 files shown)


src/Console/ModelsCommand.php
```diff
@@ -456,7 +456,9 @@ protected function getPropertiesFromMethods($model)
                     //Use reflection to inspect the code, based on Illuminate/Support/SerializableClosure.php
                     $reflection = new \ReflectionMethod($model, $method);
                     // php 7.x type or fallback to docblock
-                    $type = (string) $reflection->getReturnType() ?: (string)$this->getReturnTypeFromDocBlock($reflection);
+                    $type = (string) $reflection->getReturnType()
+                        ? (string) $reflection->getReturnType()
+                        : (string)$this->getReturnTypeFromDocBlock($reflection);
```
mfn: Still looks slightly wrong:
`? (string) $re…`
vs
`: (string)$thi`

liquid207: I saw Travis CI passed, but not good enough do you have a better way?

PR #804, opened 2019-07-08
Model relation properties being overwritten

markokeeffe:
There was a bug introduced in the ModelsCommand here: https://github.com/barryvdh/laravel-ide-helper/commit/bcb881b20d57c3408bead36f1a836bb8137a13b7 in `src/Console/ModelsCommand.php:483`.

The loose string comparison of `stripos($impl, $type) !== false` means that several relation declarations are being overwritten with an incorrectly matched relation based on relation type.

**All `morphTo` relations will be overwritten using the `morphToMany` relation type, and then overwritten again using the `morphedByMany` relation type.** 

This results in PHPDoc block for those relations showing as (for example): 
```
@property-read \Illuminate\Database\Eloquent\Collection|\App\Models\Comms\MessageRecipient[] $owner
```
Instead of: 
```
@property-read \Illuminate\Database\Eloquent\Model|\Eloquent $owner
```

**All `hasMany` relations will be overwritten using the `hasManyThrough` relation type (looks like the logic for handling the two is the same, so no immediate harm here)**

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -483,7 +483,7 @@ class ModelsCommand extends Command
                                'morphedByMany' => '\Illuminate\Database\Eloquent\Relations\MorphToMany'
                              ) as $relation => $impl) {
                         $search = '$this->' . $relation . '(';
-                        if (stripos($code, $search) || stripos($impl, (string)$type) !== false) {
+                        if (stripos($code, $search) || $impl === (string)$type) {
                             //Resolve the relation's model to a Relation object.
                             $methodReflection = new \ReflectionMethod($model, $method);
                             if ($methodReflection->getNumberOfParameters()) {
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -409,133 +409,129 @@ class ModelsCommand extends Command
     protected function getPropertiesFromMethods($model)
     {
         $methods = get_class_methods($model);
-        if ($methods) {
-            sort($methods);
-            foreach ($methods as $method) {
-                if (Str::startsWith($method, 'get') && Str::endsWith(
-                    $method,
-                    'Attribute'
-                ) && $method !== 'getAttribute'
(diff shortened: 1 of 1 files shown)


src/Console/ModelsCommand.php
```diff
@@ -409,133 +409,129 @@ protected function getPropertiesFromTable($model)
...
+                        if ($methodReflection->getNumberOfParameters()) {
+                            continue;
+                        }
 
-                            $relationObj = $model->$method();
-
-                            if ($relationObj instanceof Relation) {
-                                $relatedModel = '\\' . get_class($relationObj->getRelated());
-
-                                $relations = [
```
barryvdh: Is this not required anymore?

mfn:
Can you please remove your whitespace-only changes? Makes it extremely hard to read.

Although [`?w=1`](https://github.com/barryvdh/laravel-ide-helper/pull/804/files?w=1) works, it shouldn't be necessary.

Thanks!

barryvdh:
Yeah I checked with ?w, but it's because of the indent change. Not a big problem.

mfn:
> because of the indent change

Oh, sorry for the noise, thanks!

barryvdh:
No worries, you are right that it's easier to review without whitespace changes, especially when they're not really required.

src/Console/ModelsCommand.php
markokeeffe: It's never used

mfn:
Change looks good but IMHO the whole refactoring turned everyone off from seeing the one-liner change properly, IDK.

I made https://github.com/barryvdh/laravel-ide-helper/pull/804 which cherry-picks the actual change, original author is kept (!) and test is adapted, too.

markokeeffe:
I made the whitespace changes as the CI build failed on the `phpcs` step for line lengths so I thought I'd clean up the formatting of the code a bit and remove an unused variable.

I'll close this PR if you aren't going to merge it.

PR #825, opened 2019-09-11
Update readme.md

s-patompong:
In version 2.6.5, this commit https://github.com/barryvdh/laravel-ide-helper/commit/f8a8cb18e0a42ae4131b7ed865fe792f9b34630c made it so library's consumer no longer need to pull in `doctrine/dbal` manually. I think we should mention that in the readme file as well. Although I'm not sure if it adds any value to the file itself or not.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -91,13 +91,6 @@ The `Illuminate/Support/helpers.php` is already set up, but you can add/remove y
 
 ### Automatic phpDocs for models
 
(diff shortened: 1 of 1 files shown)


barryvdh:
I think we can just remove the nite entirely.

s-patompong:
@barryvdh thanks for the suggestion. 

New commit that remove doctrine/dbal note is pushed to this PR.

PR #835, opened 2019-10-07
Update Generator.php

shopapps:
fixes php 7.3 error when a null item is inserted into  array_merge

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -50,9 +50,9 @@ class Generator
         // Find the drivers to add to the extra/interfaces
         $this->detectDrivers();
 
-        $this->extra = array_merge($this->extra, $this->config->get('ide-helper.extra'));
-        $this->magic = array_merge($this->magic, $this->config->get('ide-helper.magic'));
-        $this->interfaces = array_merge($this->interfaces, $this->config->get('ide-helper.interfaces'));
+        $this->extra = array_merge((array)$this->extra, (array)$this->config->get('ide-helper.extra'));
+        $this->magic = array_merge((array)$this->magic, (array)$this->config->get('ide-helper.magic'));
+        $this->interfaces = array_merge((array)$this->interfaces, (array)$this->config->get('ide-helper.interfaces'));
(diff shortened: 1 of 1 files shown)


barryvdh:
What are the values now?

shopapps:
I'm not sure what values it calculated to,  all I know is that one of them clearly was null which kicked up an Exception in php 7.3 (previously we had been using 7.1 which -presumably- accepted null values).    If i can get any spare time later today I can undo my changes and see what comes up.

Ideally there should be variable validation as part of the code but casting them both as (array) when inserting into array_merge is a generally accepted quick and dirty workaround :-D

barryvdh:
Yeah we should probably add the default value as empty array.

shopapps:
sounds like a better approach - lol

src/Generator.php
```diff
@@ -50,9 +50,9 @@ public function __construct(
         // Find the drivers to add to the extra/interfaces
         $this->detectDrivers();
 
-        $this->extra = array_merge($this->extra, $this->config->get('ide-helper.extra'));
-        $this->magic = array_merge($this->magic, $this->config->get('ide-helper.magic'));
-        $this->interfaces = array_merge($this->interfaces, $this->config->get('ide-helper.interfaces'));
+        $this->extra = array_merge((array)$this->extra, (array)$this->config->get('ide-helper.extra'));
+        $this->magic = array_merge((array)$this->magic, (array)$this->config->get('ide-helper.magic'));
+        $this->interfaces = array_merge((array)$this->interfaces, (array)$this->config->get('ide-helper.interfaces'));
```
mfn: > all I know is that one of them clearly was null 

The reason is likely that your config didn't have these settings at all. Maybe because you upgraded from an older version.

However, the correct solution is not to cast but to make the config lookups to default to an empty array.

Would you mind updating the PR?

mfn:
@barryvdh IMHO vote for closing with same reasoning as in https://github.com/barryvdh/laravel-ide-helper/pull/836#issuecomment-690710241

PR #836, opened 2019-10-07
Update ModelsCommand.php

shopapps:
Fix for PHP7.3 error:  array_merge(): Expected parameter 1 to be an array, null given

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -82,8 +82,8 @@ class ModelsCommand extends Command
         $filename = $this->option('filename');
         $this->write = $this->option('write');
         $this->dirs = array_merge(
-            $this->laravel['config']->get('ide-helper.model_locations'),
-            $this->option('dir')
+            (array) $this->laravel['config']->get('ide-helper.model_locations'),
+            (array) $this->option('dir')
         );
         $model = $this->argument('model');
         $ignore = $this->option('ignore');


src/Console/ModelsCommand.php
```diff
@@ -82,8 +82,8 @@ public function handle()
         $filename = $this->option('filename');
         $this->write = $this->option('write');
         $this->dirs = array_merge(
-            $this->laravel['config']->get('ide-helper.model_locations'),
-            $this->option('dir')
+            (array) $this->laravel['config']->get('ide-helper.model_locations'),
```
mfn: This is the same as https://github.com/barryvdh/laravel-ide-helper/pull/835#discussion_r363038544
- don't cast, but provide a default value

src/Console/ModelsCommand.php
```diff
@@ -82,8 +82,8 @@ public function handle()
...
+            (array) $this->option('dir')
```
mfn: Why is this necessary?

It's defined as array so it can't be not an array, see https://github.com/barryvdh/laravel-ide-helper/blob/c0bb9ee1fd623f90c2ba593ab151923eb66c933a/src/Console/ModelsCommand.php#L152

src/Console/ModelsCommand.php
DannyvdSluijs: The reason for this not being an array is because in config it was assigned with a different type. Casting to an array would only hide the fact there is a misconfiguration.
Perhaps it would be smart to introducing a config value object/DTO that is made available from the DI. On creation it should validate the config, this to avoid scattered configuration checks throughout the code.

src/Console/ModelsCommand.php
DannyvdSluijs: Probably the array casting is added to ensure it becomes an array. The signature `InteractsWithIO::option(string $key = null): string|array|bool|null` shows it potentially can return many types. I'm not sure if this is a good thing to add though it would break when changing the definition.

mfn:
@barryvdh I don't think casting the returned value from the config is the correct approach: it would simply hide bugs in the config that way or have other side effects even harder to debug => I vote for closing this PR!

PR #837, opened 2019-10-10
Make writing relation count properties optional

AegirLeet:
https://github.com/barryvdh/laravel-ide-helper/pull/783 introduced writing of DocBlock comments for `$relation_count` properties, populated by `withCount`/`loadCount`. This makes writing those comments optional and configurable (still on by default).

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -52,6 +52,22 @@ return array(
 
     /*
     |--------------------------------------------------------------------------
+    | Write Model relation count properties
+    |--------------------------------------------------------------------------
+    |
+    | Set to false to disable writing of relation count properties to model DocBlocks.
+    |
+    | If set to true, every "x to many" relation will generate an accompanying
+    | "$relation_count" property DocBlock comment.
+    |
+    | If set to false, no such DocBlock comments will be generated.
+    |
+    */
+
+    'write_model_relation_count_properties' => true,
+
+    /*
+    |--------------------------------------------------------------------------
     | Write Eloquent Model Mixins
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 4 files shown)


spawnia:
I am interested in this configuration too.

Same as `write_model_magic_where`, i think this should be optional.

config/ide-helper.php
```diff
@@ -50,6 +50,22 @@
...
+    |--------------------------------------------------------------------------
+    | Write Model relation count properties
+    |--------------------------------------------------------------------------
+    |
+    | Set to false to disable writing of relation count properties to model DocBlocks.
+    |
+    | If set to true, every "x to many" relation will generate an accompanying
+    | "$relation_count" property DocBlock comment.
+    |
+    | If set to false, no such DocBlock comments will be generated.
```
spawnia: This seems a bit redundant, i think the description is long and clear enough without this line.

AegirLeet:
@barryvdh Any chance to get this merged?

AegirLeet:
Is this repo still maintained?

config/ide-helper.php
mfn: I second this, please remove it

PaulRotter:
> I think this would be useful addition 👍
> 
> Even better if you could add a test!

Am I right in assuming that the missing test is the only thing, that prevents this PR from being merged?

mfn:
@PaulRotter I'm not the ultimate decision make here, but it's my believe that changes/added features should be accompanied with tests showing they work as intended.

I therefore made sure that, especially the ModelsCommand, can be tested and added an initial batch of tests so others can follow more easily; you should find enough examples to make a test and it shouldn't take you longer than the the actual implementation.

Personally I don't see anything (else) in the way of merging this.

PR #802, opened 2019-07-02
Fix issue 784 `Arr` and `Str` no autocomplete.

liquid207:

PR #841, opened 2019-10-28
Fix non-facade classes will result in no autocomplete

### LINKED ISSUE #784: Laravel added aliases for 'Arr' and 'Str' no autocomplete from helper
ide-helper version: 2.6.2

In Laravel 5.8.3 aliases were added for the `Arr` and `Str` classes by default: 

https://github.com/laravel/laravel/pull/4951

They are **not added as a `Facade`** though just `alias` of a `class`, **I assume this is the issue**:
```php
'Arr' => Illuminate\Support\Arr::class,
'Str' => Illuminate\Support\Str::class,
```

In `PHPStorm` I am not getting any autcomplete functions for these classes:
image

Do I need to create my own `Facade` for these classes?

---
**ide-helper.php**
The `ide-helper` outputs this in the `ide-helper.php` file:

```php
...
namespace Illuminate\Support { 

    class Arr {
         
    }

    class Str {
         
    }
 
}
...
namespace  { 
    ....

    class Arr extends \Illuminate\Support\Arr {}

    ....

    class Str extends \Illuminate\Support\Str {}

    ....
}
...
```

---

**Temporary solution**
I get autocomplete if I remove this section from the `ide-helper.php` but this file is regenerated often. 

```php
...
namespace Illuminate\Support { 

    class Arr {
         
    }

    class Str {
         
    }
 
}
```

netpok:
Currently non-facade classes are resulting non functioning autocomplete. The reason of this that an empty class will be defined in the ide helper class, which will be used for the auto-completion. This pull request prevents this behavior by only handling facades in the namespaced part of the helper file.

Unlike #802  this pull request will keep the alias registration.

Fixes #784

### DIFF
--- a/resources/views/helper.php
+++ b/resources/views/helper.php
@@ -21,15 +21,12 @@
  */
 
 <?php foreach ($namespaces_by_extends_ns as $namespace => $aliases) : ?>
-    <?php if ($namespace == '\Illuminate\Database\Eloquent') :
-        continue;
-    endif; ?>
-namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> { 
+namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> {
     <?php foreach ($aliases as $alias) : ?>
-        <?= trim($alias->getDocComment('    ')) ?> 
+        <?= trim($alias->getDocComment('    ')) ?>
         <?= $alias->getClassType() ?> <?= $alias->getExtendsClass() ?> {
         <?php foreach ($alias->getMethods() as $method) : ?>
-            <?= trim($method->getDocComment('        ')) ?> 
+            <?= trim($method->getDocComment('        ')) ?>
         public static function <?= $method->getName() ?>(<?= $method->getParamsWithDefault() ?>)
         {<?php if ($method->getDeclaringClass() !== $method->getRoot()) : ?>
             //Method inherited from <?= $method->getDeclaringClass() ?>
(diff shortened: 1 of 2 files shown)


mfn:
https://github.com/barryvdh/laravel-ide-helper/pull/841/files?w=1 :-)

netpok:
Hoops, I accidentally removed the spaces (not that they should be there).

mfn:
Seems to me this is the same as https://github.com/barryvdh/laravel-ide-helper/pull/802 but with less "noise"?

netpok:
Sorry for the late reply, the problem with #802 is that it will remove the Arr and Str classes from the
```php
namespace{
    // ...
}
```
section so the \Arr and \Str calls will lose autocomplete.

Also the import
```php
use Illuminate\Support\Facades;
```
should be
```php
use Illuminate\Support\Facades\Facade;
```
because currently it filters out all the facades. I think this was some newer change because originally it worked for me (I think).

---

This version does not have the mentioned issues. Also I rebased my pull request to the latest master.

nalatreb (commented):
I think you should update your branch, because there are changes in this method: `getAliasesByExtendsNamespace`

netpok:
Ok. I rebased it to current master.

resources/views/helper.php
```diff
@@ -21,15 +21,12 @@
  */
 
 <?php foreach ($namespaces_by_extends_ns as $namespace => $aliases) : ?>
-    <?php if ($namespace == '\Illuminate\Database\Eloquent') :
-        continue;
-    endif; ?>
```
mfn, line 26: Why is this removed?

netpok: It was originally filtered out because it's not a facade so it doesn't need to be there but all non facade classes are filtered out already.

mfn: Ah, nice, thank you 👍

mfn (approved):
Generally LGTM

There's one thing I'm not sure: AFAIK people sometimes expect non-Facades also to show up here, which would have happened before.

Also I hope at some day we can have some tests for this, but I see that this isn't easy here…

netpok:
Basically you don't want the non Facade classes show up there as they would be empty and break the aliased classes. They still show up in the alias section, so `Illuminate\Support\Arr` will not be here but `\Arr extends Illuminate\Support\Arr` will be here.

mfn:
Thank you for your clarification, makes sense to me!

netpok:
Does this PR needs any more attention from me or is it good for now?

mfn:
Good question, I think we wait for approval from @barryvdh 

image

netpok:
@barryvdh Any blocking reason here I should resolve?

KentarouTakeda:
@barryvdh @mfn 
Do we have any update?
With this nice fix, It will prevent phpstan>=1.7 from reporting many errors to empty helper class.

mfn:
Still has my 👍🏼 !

snmatsui:
ping

PR #842, opened 2019-10-28
Automatically generate helpers

netpok:
This pull request allows automatic helper generation. This is achieved by listening on the command finished event of `migrate`, `migrate:rollback` and `package:discover` commands.

I added this feature disabled by default, but it could be default on in the next major version.

Feel free to rename the config variable, this was my first idea.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,9 @@
         "scrutinizer/ocular": "~1.1",
         "squizlabs/php_codesniffer": "^3"
     },
+    "suggest": {
+        "illuminate/events": "Required for automatic helper generation (^5.5|^6)."
+    },
     "autoload": {
         "psr-4": {
             "Barryvdh\\LaravelIdeHelper\\": "src"
--- a/src/IdeHelperServiceProvider.php
+++ b/src/IdeHelperServiceProvider.php
@@ -14,6 +14,8 @@ use Barryvdh\LaravelIdeHelper\Console\EloquentCommand;
 use Barryvdh\LaravelIdeHelper\Console\GeneratorCommand;
 use Barryvdh\LaravelIdeHelper\Console\MetaCommand;
 use Barryvdh\LaravelIdeHelper\Console\ModelsCommand;
+use Barryvdh\LaravelIdeHelper\Listeners\GenerateHelpers;
+use Illuminate\Console\Events\CommandFinished;
 use Illuminate\Support\ServiceProvider;
 use Illuminate\View\Engines\EngineResolver;
 use Illuminate\View\Engines\PhpEngine;
@@ -37,6 +39,10 @@ class IdeHelperServiceProvider extends ServiceProvider
      */
     public function boot()
     {
+        if ($this->app['config']->get('ide-helper.listen_on_commands')) {
+            $this->app['events']->listen(CommandFinished::class, GenerateHelpers::class);
+        }
+
         if ($this->app->has('view')) {
             $viewPath = __DIR__ . '/../resources/views';
             $this->loadViewsFrom($viewPath, 'ide-helper');
--- a/src/Listeners/GenerateHelpers.php
+++ b/src/Listeners/GenerateHelpers.php
@@ -34,6 +34,7 @@ class GenerateHelpers
         switch ($event->command) {
             case "package:discover":
                 $this->generate('generate', $event->output);
+                $this->generate('meta', $event->output);
                 break;
             case "migrate":
             case "migrate:rollback":
@@ -51,7 +52,7 @@ class GenerateHelpers
     {
         $parameters = $this->config->get(
             'ide-helper.listen_'.$command.'_parameters',
-                array()
+            array()
         );
 
         if ($parameters === false) {
--- a/src/Listeners/GenerateHelpers.php
+++ b/src/Listeners/GenerateHelpers.php
@@ -36,7 +36,10 @@ class GenerateHelpers
             case "package:discover":
                 $this->artisan->call(
                     'ide-helper:generate',
-                    $this->config->get('ide-helper.listen_generate_parameters', []),
+                    $this->config->get(
+                        'ide-helper.listen_generate_parameters',
+                        array()
+                    ),
                     $event->output
                 );
                 break;
@@ -44,7 +47,10 @@ class GenerateHelpers
             case "migrate:rollback":
                 $this->artisan->call(
                     'ide-helper:models',
-                    $this->config->get('ide-helper.listen_model_parameters', []),
+                    $this->config->get(
+                        'ide-helper.listen_model_parameters',
+                        array()
+                    ),
                     $event->output
                 );
                 break;
--- a/src/Listeners/GenerateHelpers.php
+++ b/src/Listeners/GenerateHelpers.php
@@ -48,7 +48,7 @@ class GenerateHelpers
                 $this->artisan->call(
                     'ide-helper:models',
                     $this->config->get(
-                        'ide-helper.listen_model_parameters',
+                        'ide-helper.listen_models_parameters',
                         array()
                     ),
                     $event->output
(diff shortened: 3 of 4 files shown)


src/Listeners/GenerateHelpers.php
```diff
@@ -0,0 +1,28 @@
...
+     */
+    public function handle(CommandFinished $event)
+    {
+        switch ($event->command) {
+            case "package:discover":
+                Artisan::call('ide-helper:generate', [], $event->output);
+                break;
+            case "migrate":
+            case "migrate:rollback":
+                Artisan::call('ide-helper:models', ['--no-write'], $event->output);
```
mfn: Problem here is that this is project dependent.

E.g. I use `ide-helper:models --write --reset`

Maybe this too should be in the config what args to pass?

netpok: I haven't thought of that, done.

netpok:
I think this pull request is complete from my side. Of course suggestions are welcomed.

mfn:
This competes with https://github.com/barryvdh/laravel-ide-helper/pull/797

netpok:
After comparing the two solution my opinion is:
 - The other solution adds an extra event check to every single database query, it doesn't do anything as the recorder is not enabled but it's still an overhead (I know this is only a dev package). That issue can be simply rectified by registering the query listener only on the started event.
 - I'm not sure that it's necessary to only update the models that were affected by the migration, it is much easier to update all the models.
 - It only resolves the models under the app path so any external models like the cashier subscription model will be skipped.

Now the other side of the coin:
 - The config structure is probably better than mine.
 - The use of the MigrationsStarted event is a nice catch that I missed, since I made this pull request I noticed a problem where only a full match triggers the update, so for example `migrate:roll` not while it still rolls the migration back.

---

The best of the two would be in my opinion:
 - Adopt the other config syntax
 - Update the models simply on the MigrationsEnded event (Probably just trigger a flag on that event and still handle it in the CommandFinished event where we can get a valid output interface for the artisan call)
 - Update the helpers when the `package:discover` command is finished since this command is usually executed by the composer there is a low chance of missing characters. Sadly there is no PackagesDiscovered event as far as I know

I considered dropping the helper update part in favor of the post update section but it looks quite hacky to me. If the script runs in a non dev environment it fails with an error code to hide the error massage and make composer not fail one needs to use the following:
```
@php artisan ide-helper:generate > /dev/null || true
```
While this solves the problem I'm not sure that this command does anything usable under windows.

@mfn and anyone else: Please share your thoughts about this.

mfn:
I didn't share this information yet but when I found the two PRs doing the same, my initial gut feeling was (and still is) from a technical approach I prefer this PR.

But I also agree with your assessment regarding the config options 😄 

However before putting more work into this, IMHO @barryvdh should voice his preference here (this PR, other PR or no PR)

mortenscheel:
I'm the author of the other PR, and I agree with everything you've both said.

The reason I decided to only update the affected models was that I was having a weird issue with git at the time. After running `ide-helper:models`, every single model file (even unmodified ones) would show up as modified. It's not an issue any longer, and I suppose it was just a misconfiguration of git's `core.autocrlf` config.

So yeah, that complicated workaround wasn't necessary after all. Just update everything on `MigrationsEnded`.

mfn:
@mortenscheel thanks for your feedback!

Would you then mind closing your PR in favour of this one?

src/Listeners/GenerateHelpers.php
```diff
@@ -0,0 +1,64 @@
...
+     */
+    public function handle(CommandFinished $event)
+    {
+        switch ($event->command) {
+            case "package:discover":
+                $this->generate('generate', $event->output);
+                $this->generate('meta', $event->output);
+                break;
+            case "migrate":
+            case "migrate:rollback":
```
jpickwell:
```suggestion
            case "migrate:fresh":
            case "migrate:refresh":
            case "migrate:reset":
            case "migrate:rollback":
```

src/Listeners/GenerateHelpers.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * Handle the event.
+     *
+     * @param  CommandFinished  $event
+     */
+    public function handle(CommandFinished $event)
+    {
+        switch ($event->command) {
```
jpickwell:
```suggestion
        if ($this->eventIsInvalid($event)) {
            return;
        }

        switch ($event->command) {
```

src/Listeners/GenerateHelpers.php
```diff
@@ -0,0 +1,64 @@
...
+                $this->generate('generate', $event->output);
+                $this->generate('meta', $event->output);
+                break;
+            case "migrate":
+            case "migrate:rollback":
+                $this->generate('models', $event->output);
+                break;
+        }
+    }
+
```
jpickwell:
```suggestion

    /**
     * @param \Illuminate\Console\Events\CommandFinished $event
     *
     * @return bool
     */
    protected function eventIsInvalid(CommandFinished $event): bool
    {
        /*
         * The event is invalid if any of the following are true:
         * 1. exitCode is not zero (0)
         * 2. input does not have a "command" argument; this happens when the
         *    -V/--version flag is used
         * 3. input has "help" option (-h/--help)
         * 4. input has "pretend" option (--pretend)
         */
        return $event->exitCode !== 0 ||
            !$event->input->hasArgument("command") ||
            $this->hasOption($event, "help") ||
            $this->hasOption($event, "pretend");
    }

```

mortenscheel:
Looking forward to having this feature added. Any progress?

netpok:
I'm still waiting response from the project maintainers about this pull request.

src/Listeners/GenerateHelpers.php
netpok: You have a point, this should be mostly solved by using MigrationsEnded event as a trigger as mentioned above, nut it can be definitely used on the package discover command, because as far as I know there is no PackageDiscovered event. Thanks for the suggestions.

mfn:
:/

ping @barryvdh 🤗

netpok:
I will try to make a pull request to the framework with a PackagesDiscovered event so we could use that.

composer.json
```diff
@@ -26,6 +26,9 @@
         "scrutinizer/ocular": "~1.1",
         "squizlabs/php_codesniffer": "^3"
     },
+    "suggest": {
+        "illuminate/events": "Required for automatic helper generation (^5.5|^6)."
```
mfn, line 30: This should also have `|^7` in the meantime

src/IdeHelperServiceProvider.php
```diff
@@ -37,6 +39,10 @@ class IdeHelperServiceProvider extends ServiceProvider
      */
     public function boot()
     {
+        if ($this->app['config']->get('ide-helper.listen_on_commands')) {
```
mfn, line 42: the default for `get()` (2nd parameter) should be `false` for clarity sake

(it will work anyway because it's `null`, but…)

barryvdh:
I like the idea, but not sure if this is the best way. For Composer commands you can just add it to composer.json, but for migrations I see the benefit. I think it would be enough to just have a config option `'post-migrate' => ['ide-helper:generate -RW']` or something?

mfn:
I side with @barryvdh last comment, this would combine the "established way" (composer) but allows a way into the migration (with the flexibility of providing how)

netpok:
I will update the PR for the changes, I found out that:
`"test $COMPOSER_DEV_MODE -eq 0 || $PHP_BINARY artisan ide-helper:generate"`
can be used in the post dump section without displaying or generating errors in the production environment

netpok:
Sorry, I forgot this, I will have some time this week to update this

composer.json
mfn: Time has passed:
- `^5.5` can be removed
- `^7` and `^8` added 😄

MatanYadaev:
Is this PR dead?

netpok:
Sorry for not handling this, I'm closing it in favor of #1163 which is an updated version for this.

PR #844, opened 2019-11-06
Builder 从 model 中获取，万一有人修改了 model 自带的 builder 呢！

lee2son:


mfn:
What's the reason for this change, what does it fix?

netpok:
For me it looks like a case where the model is not an Eloquent one but instead something else (like a [MongoDB one](https://github.com/jenssegers/laravel-mongodb/blob/master/src/Jenssegers/Mongodb/Query/Builder.php))

mfn:
Ah, thanks for the hint!

Looking at it from a technical standpoint, this seems quite ok getting the class from the actual method.

However I still would like to see test for this (rebase with master, there are now Model).

Also:
- the permission of the ModelsCommand should not change (?)
- upstream repo/branch doesn't exist anymore 🤷‍♀️

mfn:
@barryvdh author never responded, the intention is unclear, merge conflicts => I suggest to drop the PR :}

mfn:
I'm closing this as the original author isn't responding, it lacks a clear description and also has no tests.

PR #859, opened 2019-12-21
Add support for custom date class via Date::use()

mfn:
Laravel 5.8 introduced a feature to support a custom date class via
`Date::use()`, see https://github.com/laravel/framework/pull/25320

When e.g. using `Date::use(CarbonImmutable)` in a project, it means
all date casts are not returning `\Illuminate\Support\Carbon` anymore
but `\Carbon\CarbonImmutable`, which means all the generated type hints
for dates are now wrong.

This change tries to be still backwards compatible with Laravel < 5.8
which do not have the Date facade.

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -64,6 +64,14 @@ class ModelsCommand extends Command
     protected $nullableColumns = [];
 
     /**
+     * During initializtion we use Laravels Date Facade to
+     * determine the actual date class and store it here.
+     *
+     * @var string
+     */
+    protected $dateClass;
+
+    /**
      * @param Filesystem $files
      */
     public function __construct(Filesystem $files)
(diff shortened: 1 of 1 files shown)


barryvdh:
Thanks

amcsi:
This unfortunately only is for the date type model properties. This does not make functions like `now()` show that it returns CarbonImmutable.

mfn:
> This unfortunately only is for the date type model properties. This does not make functions like `now()` show that it returns CarbonImmutable.

Indeed, this PR is only for the models.

I think you're referring to the Facade, right? I don't use them so 🤷‍♀️

Since the projects I needed for made a final decision to move to immutable, we use `CarbonImmutable::now()` directly in the code anyway.

amcsi:
I know a benefit to the facade is that it lets you easily swap out what the current time is considered to be.

I don't expect to you to fix this by the way. I rather just wanted to make sure people are aware of this fact.

PR #866, opened 2020-01-01
[CHORE] Remove unused/stale utilities

mfn:
### ~Remove Grumphp~
**Edit:** per request, did not remove it but adjusted the config to exclude `resources/`

###  Remove scrutinizer/ocular
Looks unused, also not used on travis

###  Remove StyleCI
Isn't enabled


barryvdh:
Grumphp probably shouldn't run on the views folder, like Travis. I like it for local development, because Travis runs after pushing, grumphp prevents it (a bit)

mfn:
Roger that!

I:
- reinstated grumphp
- fixed the config so `resources/` is excluded

barryvdh:
Thanks!

PR #867, opened 2020-01-01
[CHORE] Bump grumphp version

mfn:
The current version emits the following:
```
Deprecated: strpos(): Non-string needles will be interpreted as strings in the future. Use an explicit chr() call to preserve the current behavior in laravel-ide-helper/vendor/phpro/grumphp/src/Locator/ConfigurationFile.php on line 75

Deprecated: strpos(): Non-string needles will be interpreted as strings in the future. Use an explicit chr() call to preserve the current behavior in laravel-ide-helper/vendor/phpro/grumphp/src/Locator/ConfigurationFile.php on line 75
```

Manually tested it locally, worked as expected.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     "require-dev": {
         "illuminate/config": "^5.5|^6",
         "illuminate/view": "^5.5|^6",
-        "phpro/grumphp": "^0.14",
+        "phpro/grumphp": "^0.17.1",
         "squizlabs/php_codesniffer": "^3",
         "orchestra/testbench": "^3|^4",
         "mockery/mockery": "^1.3"


mfn:
Travis fails because the newer grumphp requires 7.2 but it's always installed as a dev dependency. Even it's not needed on travis.

I'll check something else first and will come back to make it 💚

mfn:
Let's wait for the outcome of https://github.com/barryvdh/laravel-ide-helper/pull/868 , because if that PR get's merged, all builds will automatically green here!

mfn:
@barryvdh all green here, [since 7.2 is now the minimum](https://github.com/barryvdh/laravel-ide-helper/pull/868) √

PR #868, opened 2020-01-01
Drop support for PHP < 7.2 and improve test matrix

mfn:
Based on the feedback in https://github.com/barryvdh/laravel-ide-helper/pull/865#issuecomment-570085281 I _tried_ to come up with a possible solution here:

This is mostly driven by two factors:
- Everything < PHP 7.2 is **not** supported anymore => https://www.php.net/supported-versions.php
- In light of expanding the test suite, many recent versions of tools have 7.2 as minimum requirement

As such, the changes summarized here are:
- drop support for < 7.1 and explicitly bump php verison in `composer.json `
- rework the travis test matrix **to be explicit**
  - Previously it was unclear what was really tested because only the PHP version was specified and all the dependencies aligned itself, but that didn't express what Laravel version was actually tested => this is now explicit
  - as such was we now can run phpunit for all supported versions 🎉
- this requires relaxing the version constraint for orchestra/testbench so composer can chose the best supported version
- ~still some special work was needed for Laravel 5.5:~
  - ~a shim for `assertNotStringContainsString`~
  - ~explicitly install mockery/mockery, as older orchestra/testbanch versions didn't require it~
  => **was removed**
- added mockery as root dependency
- added `--prefer-lowest` to test matrix

Does not approach make sense to everyone?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,34 +12,53 @@ env:
 matrix:
   include:
     - php: '7.2'
-      env: LARAVEL='5.5.*'
+      env: LARAVEL='5.5.*' COMPOSER_FLAGS='--prefer-lowest'
     - php: '7.2'
-      env: LARAVEL='5.8.*'
+      env: LARAVEL='5.5.*' COMPOSER_FLAGS='--prefer-stable'
     - php: '7.2'
-      env: LARAVEL='^6.0'
+      env: LARAVEL='5.8.*' COMPOSER_FLAGS='--prefer-lowest'
+    - php: '7.2'
+      env: LARAVEL='5.8.*' COMPOSER_FLAGS='--prefer-stable'
+    - php: '7.2'
+      env: LARAVEL='^6.0' COMPOSER_FLAGS='--prefer-lowest'
+    - php: '7.2'
+      env: LARAVEL='^6.0' COMPOSER_FLAGS='--prefer-stable'
     - php: '7.3'
-      env: LARAVEL='5.5.*'
+      env: LARAVEL='5.5.*' COMPOSER_FLAGS='--prefer-lowest'
     - php: '7.3'
-      env: LARAVEL='5.8.*'
+      env: LARAVEL='5.5.*' COMPOSER_FLAGS='--prefer-stable'
     - php: '7.3'
-      env: LARAVEL='^6.0'
+      env: LARAVEL='5.8.*' COMPOSER_FLAGS='--prefer-lowest'
+    - php: '7.3'
+      env: LARAVEL='5.8.*' COMPOSER_FLAGS='--prefer-stable'
+    - php: '7.3'
+      env: LARAVEL='^6.0' COMPOSER_FLAGS='--prefer-lowest'
+    - php: '7.3'
+      env: LARAVEL='^6.0' COMPOSER_FLAGS='--prefer-stable'
     - php: '7.4'
-      env: LARAVEL='5.5.*'
+      env: LARAVEL='5.5.*' COMPOSER_FLAGS='--prefer-lowest'
     - php: '7.4'
-      env: LARAVEL='5.8.*'
+      env: LARAVEL='5.5.*' COMPOSER_FLAGS='--prefer-stable'
     - php: '7.4'
-      env: LARAVEL='^6.0' RUN_PHPCS=1
+      env: LARAVEL='5.8.*' COMPOSER_FLAGS='--prefer-lowest'
+    - php: '7.4'
+      env: LARAVEL='5.8.*' COMPOSER_FLAGS='--prefer-stable'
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -7,33 +7,41 @@ cache:
...
   global:
-    - RUN_PHPUNIT=1
     - RUN_PHPCS=0
 
 matrix:
   include:
-    - php: 7.0
-      env: RUN_PHPUNIT=0
-    - php: 7.0
-      env: RUN_PHPUNIT=0 COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
```
mfn: `--prefer-lowest` isn't necessary anymore due to the explicit php/laravel version matrix

.travis.yml
```diff
@@ -7,33 +7,41 @@ cache:
...
 before_script:
   - phpenv config-rm xdebug.ini || true
 
 install:
-  - |
-    if [[ $RUN_PHPUNIT = 0 ]]; then
-      # We assume the older PHP/Laravel versions and thus restore a working dependency system for them
-      composer remove --dev orchestra/testbench --no-interaction --no-update
-    fi
-  - travis_wait 20 travis_retry composer update ${COMPOSER_FLAGS} --no-interaction --prefer-dist
```
mfn: `travis_wait` also not necessary anymore, the resolving of the version constraint without `--prefer-lowest` does not require it anymore

.travis.yml
```diff
@@ -7,33 +7,41 @@ cache:
...
+  - composer require "illuminate/filesystem:${LARAVEL}" --no-interaction --no-update
+  - composer require --dev "illuminate/config:${LARAVEL}" --no-interaction --no-update
+  - composer require --dev "illuminate/view:${LARAVEL}" --no-interaction --no-update
+  - if [[ $LARAVEL = '5.5.*' ]]; then composer require --dev mockery/mockery --no-interaction --no-update; fi
+  - composer update --no-interaction --prefer-dist
 
 script:
+  - vendor/bin/phpunit
   - if [[ $RUN_PHPCS = 1 ]]; then vendor/bin/phpcs --standard=psr2 src/; fi
-  - if [[ $RUN_PHPUNIT = 1 ]]; then vendor/bin/phpunit; fi
```
mfn: I decided to run phpunit tests first, i.e. make them more important to fail a build then the phpcs, hope that's good

tests/TestCase.php
```diff
@@ -33,4 +34,21 @@ protected function runCommand(Command $command, array $arguments = [], array $in
...
+     * @todo if support for Laravel 5.5 is dropped, this shim can be replaced with actual parent call
+     *
+     * @param string $needle
+     * @param string $haystack
+     * @param string $message
+     */
+    public static function assertStringNotContainsString(string $needle, string $haystack, string $message = ''): void
+    {
+        if (!method_exists(Assert::class, 'assertStringContainsString')) {
+            parent::assertNotContains($needle, $haystack, $message);
```
barryvdh: So is the shim the same as the other method? If so, why do we need the shim?

barryvdh:
Can't we just always require mockery?

In laravel-cors, I do it like this: https://github.com/fruitcake/laravel-cors/blob/master/.travis.yml
I don't think we need to list all versions explicitly, just the framework and the rest will follow. Possibly Testbench will just follow as well (as you do now)

tests/TestCase.php
mfn: Laravel 5.5 requires a version of orchestra/testbench which uses a PHPUnit version which does not feature `assertStringNotContainsString`

Our code calls this one and if the parent is not there, we know it's an older PHPUnit and we call the (now deprecated / old) way of doing it.

Otherwise we just use the actual parent with the code below ⬇️

mfn:
> Can't we just always require mockery?

Yeah, I thought about that too. Probably should be anyway, because we're explicitly using it!



> In laravel-cors, I do it like this: https://github.com/fruitcake/laravel-cors/blob/master/.travis.yml
> I don't think we need to list all versions explicitly, just the framework and the rest will follow. Possibly Testbench will just follow as well (as you do now)

Roger that, will give it another spin and ping you then!

.travis.yml
mfn: `--prefer-lowest` was actually re-introduced into the whole matrix!

tests/TestCase.php
mfn: …and the code is gone 😄 

Because we know added `--prefer-lowest` also to the test matrix, we've Laravel 5 versions in our tests without _and with_ the mixin markers, so I employed a different solution not requiring this method/shim which simply skips the test (if `@mixin`s are already present).

mfn:
@barryvdh all addresses, all green!

PR #870, opened 2020-01-03
Add more tests

mfn:
Summary of tests added:
- be more systematic when testing discrete column types
- relations of models
- model in another namespace gets correctly qualified
- softdeletes
- getters


mfn:
@barryvdh another round of tests!

PR #871, opened 2020-01-05
[CODESTYLE] Replace phpcs with php-cs-fixer and use pre-defined / Laravel compatible config

mfn:
## Intro
The initial idea was to simply ensure to automate (and keep it consistent) with the short array syntax => https://github.com/barryvdh/laravel-ide-helper/pull/493#issuecomment-570712736
> sure

The more I looked at it, it become clear that the general code style wasn't really coherent.

I've used https://github.com/matt-allan/laravel-code-style in the past, basically just a pre-set of php-cs-fixer config settings, most suitable for Laravel related packages.

Thus, this PR switched over to php-cs-fixer and uses that config with some "useful additions".

Since this PR could not be green _without actually including the style fixes_, the final diff of course 'sploded 💥 

[I therefore suggest to rather focus on the diff except the last 3 commits first](https://github.com/barryvdh/laravel-ide-helper/compare/9e567d538875ad87ab96b8dba9d8af6fc8e20dcd%E2%80%A6e09211e6d22dd6390fc7a25e9d7f07b6dd497ddc)

## Summary
- Replace phpcs (PHP_CodeSniffer) with php-cs-fixer and use https://github.com/matt-allan/laravel-code-style with some more config options
  - Still integrated into grumphp
- Run it on github actions instead of travis
- Fixed the whole codebase and adapted tests

I left some in-line comments to raise some eyes on selected parts.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,0 +1,18 @@
+<?php
+require __DIR__ . '/vendor/autoload.php';
+$finder = PhpCsFixer\Finder::create()
+    ->in(__DIR__);
+return (new MattAllan\LaravelCodeStyle\Config())
+        ->setFinder($finder)
+        ->setRules([
+            '@Laravel' => true,
+            '@Laravel:risky' => true,
+            // Project specific
+            'array_indentation' => true,
+            'is_null' => true,
+            'modernize_types_casting' => true,
+            'method_argument_space' => [
+                'on_multiline' => 'ensure_fully_multiline',
+            ],
+        ])
+        ->setRiskyAllowed(true);
--- a/resources/views/helper.php
+++ b/resources/views/helper.php
@@ -1,7 +1,7 @@
 <?= '<?php' ?>
 <?php
 /**
- * @var \Barryvdh\LaravelIdeHelper\Alias[][] $namespaces_by_alias_ns
+ * @var \Barryvdh\LaravelIdeHelper\Alias[][]
  * @var \Barryvdh\LaravelIdeHelper\Alias[][] $namespaces_by_extends_ns
  * @var bool $include_fluent
  * @var string $helpers
@@ -12,7 +12,7 @@
 
 /**
  * A helper file for Laravel 5, to provide autocomplete information to your IDE
- * Generated for Laravel <?= $version ?> on <?= date("Y-m-d H:i:s") ?>.
+ * Generated for Laravel <?= $version ?> on <?= date('Y-m-d H:i:s') ?>.
  *
  * This file should not be included in your code, only analyzed by your IDE!
  *
@@ -20,26 +20,26 @@
  * @see https://github.com/barryvdh/laravel-ide-helper
  */
 
-<?php foreach($namespaces_by_extends_ns as $namespace => $aliases): ?>
+<?php foreach ($namespaces_by_extends_ns as $namespace => $aliases): ?>
 <?php if ($namespace == '\Illuminate\Database\Eloquent'): continue; endif; ?>
 namespace <?= $namespace == '__root' ? '' : trim($namespace, '\\') ?> { 
-<?php foreach($aliases as $alias): ?>
+<?php foreach ($aliases as $alias): ?>
 
     <?= trim($alias->getDocComment('    ')) ?> 
     <?= $alias->getClassType() ?> <?= $alias->getExtendsClass() ?> {
-        <?php foreach($alias->getMethods() as $method): ?>
+        <?php foreach ($alias->getMethods() as $method): ?>
 
         <?= trim($method->getDocComment('        ')) ?> 
         public static function <?= $method->getName() ?>(<?= $method->getParamsWithDefault() ?>)
-        {<?php if($method->getDeclaringClass() !== $method->getRoot()): ?>
+        {<?php if ($method->getDeclaringClass() !== $method->getRoot()): ?>
 
             //Method inherited from <?= $method->getDeclaringClass() ?>
             <?php endif; ?>
 
-            <?php if($method->isInstanceCall()):?>
+            <?php if ($method->isInstanceCall()):?>
             /** @var <?=$method->getRoot()?> $instance */
             <?php endif?>
-            <?= $method->shouldReturn() ? 'return ': '' ?><?= $method->getRootMethodCall() ?>;
+            <?= $method->shouldReturn() ? 'return ' : '' ?><?= $method->getRootMethodCall() ?>;
         }
         <?php endforeach; ?> 
     }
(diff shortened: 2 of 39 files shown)


.php_cs.dist
```diff
@@ -0,0 +1,18 @@
...
+        ->setRules([
+            '@Laravel' => true,
+            '@Laravel:risky' => true,
+            // Project specific
+            'array_indentation' => true,
+            'is_null' => true,
+            'modernize_types_casting' => true,
+            'method_argument_space' => [
+                'on_multiline' => 'ensure_fully_multiline',
+            ],
```
mfn: I simple found those rules to be quite useful

resources/views/helper.php
```diff
@@ -1,7 +1,7 @@
 <?= '<?php' ?>
 <?php
 /**
- * @var \Barryvdh\LaravelIdeHelper\Alias[][] $namespaces_by_alias_ns
+ * @var \Barryvdh\LaravelIdeHelper\Alias[][]
```
mfn: That seems… unfortunate? No idea why it's removed 💥 

Doesn't affect code but IDE hinting within the helper isn't so nice for this varaible. Maybe we can leave with that for now and try to fix it (or report a php-cs-fixer bug) later 🤷‍♀️

resources/views/helper.php
```diff
@@ -12,34 +12,34 @@
 
 /**
  * A helper file for Laravel 5, to provide autocomplete information to your IDE
- * Generated for Laravel <?= $version ?> on <?= date("Y-m-d H:i:s") ?>.
+ * Generated for Laravel <?= $version ?> on <?= date('Y-m-d H:i:s') ?>.
```
mfn: Strings without inline vars are converted to '`'

src/Alias.php
```diff
@@ -10,13 +10,13 @@
...
-use Closure;
-use ReflectionClass;
 use Barryvdh\Reflection\DocBlock;
 use Barryvdh\Reflection\DocBlock\Context;
+use Barryvdh\Reflection\DocBlock\Serializer as DocBlockSerializer;
 use Barryvdh\Reflection\DocBlock\Tag\MethodTag;
+use Closure;
 use Illuminate\Config\Repository as ConfigRepository;
-use Barryvdh\Reflection\DocBlock\Serializer as DocBlockSerializer;
+use ReflectionClass;
```
mfn: Laravel (since 6?) finally switched to alphabetical imports, which is reflected in all the files.

src/Console/GeneratorCommand.php
```diff
@@ -50,17 +49,14 @@ class GeneratorCommand extends Command
...
-
     /**
-     *
      * @param \Illuminate\Config\Repository $config
      * @param \Illuminate\Filesystem\Filesystem $files
      * @param \Illuminate\View\Factory $view
      */
     public function __construct(
-        /*ConfigRepository */ $config,
+        $config,
```
mfn: I actually removed them, they cause discrepancy between running php-cs-fixer locally vs. on github.

Should not happen but TBH this is so unorthodox I simply removed it; the phpdoc above is correct(er) anyway.

mfn:
Yeah, so, reviewing code style PRs is never fun. To the best of my knowledge, nothing "broke" with it.

Not sure if we want it, but seeing all the inconsistencies fixed (which phpcs didn't report at all) somewhat gives a good overall feeling.

mfn:
Rebased with master, all green @barryvdh

barryvdh:
I'm not completely sold on the external package. I'm more fan of PSR-12 usually.. what is the big advantage of this versus psr-12?

mfn:
> what is the big advantage of this

Nothing really, but it's a Laravel package and thus I figured the code style should adhere to Laravel. Am I wrong?

Basically, _that_ package is the closest we've in terms of out-of-the-box and it's specifically for providing php-cs-fixer rules according to the "Laravel way".

> psr-12

To my knowledge, no OSS project yet exists which implements "all the rules". Certainly php-cs-fixer would be my first choice.

It's just when I increased my contributions here I realized the lack of coherent code style.

Personally I don't mind which one: but a project should have "one" which is well enforced (which currently doesn't seem so).

ps: feel free to close it, if you think it's not the right time => no hard feelings!

Thanks

mfn:
I guess this is going nowhere 😅

barryvdh:
Yeah sorry for the lack of response. I find these kind of PRs difficult. I understand the need for consistency, but it usually leads to a lot of merge conflicts with open pull requests, with little 'real' value for users.
And as stated, I believe PSR-12 is better suited.

mfn:
> Yeah sorry for the lack of response. I find these kind of PRs difficult. I understand the need for consistency, but it usually leads to a lot of merge conflicts with open pull requests, with little 'real' value for users.


@barryvdh in the hopes of being not too annoying, I'd like to ask one more time about this

I would like to propose to move to php-cs-fixer because IMHO its results are better/more useful/practical and we'll be able to fix a lot of the inconsistencies _automatically_ we still have

> And as stated, I believe PSR-12 is better suited.

I totally agree! But to the best of my knowledge, there's no "finished" PSR-12 out-of-the box config yet, but IMHO it's not such a big issue as I would propose to already not deviate from PSR-12 where we can align it (_some_ things are always up for debate, no PSR is going to specify each and every character).

Needless to say I've used this approach in multiple projects successfully: even more consistent code base and with GHA auto-commit, less friction for contributions anyway.

Yes, existing PRs are always going to suffer a bit but let's be honest: I don't see so many of them being merged anytime soon and am I'm happy to assist them where possible; I think I've proven that.

Thanks!

barryvdh:
So what do you want to do exactly? Reformat all a psr-12, or use the Laravel can fixer?

mfn:
> So what do you want to do exactly?

- switch to a tool (php-cs-fixer instead of phpcs) which I believe can do a better job at formatting
- psr-12; at least [as good as it gets](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4502) with php-cs-fixer currently
  - however I ask to have patience as no Code Style PSR12 will ever answer all code style question and there's always room for discussion in between

The gist is, it wouldn't matter to me if Laravel CS or PSR-12 (but yes, preferring PSR-12) but to also have a tool which IMHO does a better job and we can use to fix many more smaller issues (like automatically reforming unused `use` imports [which phpcs can't] and others).

Thanks!

barryvdh:
But we're already using PSR-12? 

PSR-12 is fine by me and if you think php-cs-fixer does a better job then that's also fine with me :)

mfn:
Sent in https://github.com/barryvdh/laravel-ide-helper/pull/1030

PR #873, opened 2020-01-07
Fixed bug in model command where loose string comparison would override properties for model relations

mfn:
This is the same as https://github.com/barryvdh/laravel-ide-helper/pull/804 **without** the refactoring (and adapting the tests).

I don't thin that refactoring so many lines in contrast for a single line change makes sense and could be the reason this wasn't merged already.

Note: I cherry-picked the original commit, which means @markokeeffe as the author **is kept**.


mfn:
Note: the failing test is unrelated to this change, it's because of phpcs complaining about the change from https://github.com/barryvdh/laravel-ide-helper/pull/862

PR #879, opened 2020-01-19
Restore original event dispatcher after Event::fake()

aurelien-roy:
When trying to retrieve Fake facades, a call to Event::fake() triggers a change of the event dispatcher for the Model class.
See:
https://github.com/laravel/framework/blob/5d650365e0905e9436c99134cce218207438d5bc/src/Illuminate/Support/Facades/Event.php#L40

When we swap again to the real facade, we should restore the original event dispatcher for Model.

This oversight cause the ide-helper::generate command to fail on some custom Laravel installations that rely on certain events triggering to boot properly (This is the case for OctoberCMS (http://github.com/octobercms/october/, a CMS based on Laravel)).


mfn:
Isn't this somehow a bug in the `Event` facade itself?

Seems to me that the event facade should override `::swap()` and call the two static methods currently called from `::fake`.

Since I presume you have a reproducible environment, I'm curious if you can change the Event facade temporarily and try if it also fixes the problem?
```php
    public static function fake($eventsToFake = [])
    {
        static::swap($fake = new EventFake(static::getFacadeRoot(), $eventsToFake));

        return $fake;
    }
    
    public static function swap($instance)
    {
        parent::swap($instance);
        
        Model::setEventDispatcher($fake);
        Cache::refreshEventDispatcher();        
    }
```

aurelien-roy:
Thanks for the suggestion @mfn !
Indeed your solution is working and seems more appropriate. (Though it won't fix the problem with October since it relies on a old Laravel version, but I guess I'll create my own fork rather than dirty patching this repo^^')

aurelien-roy:
Thanks for the suggestion @mfn !
Indeed your solution is working and seems more appropriate. (Though it won't fix the problem with October since it relies on a old Laravel version, but I guess I'll create my own fork rather than dirty patching this repo^^')

aurelien-roy:
Owh. In fact it seems they also have a 'fakeFor' method that is intented to be used when needing to use a fake facade temporarily. Shouldn't ide-helper uses this one?

mfn:
> Shouldn't ide-helper uses this one over fake() when available?

Good find 😄 "I think so" would be my first reaction.

src/Alias.php
```diff
@@ -213,6 +214,9 @@ protected function detectFake()
             }
         } finally {
             $facade::swap($real);
+            if(get_class($fake) == 'Illuminate\Support\Testing\Fakes\EventFake') {
```
mfn, line 217: Maybe also import?
```suggestion
            if(get_class($fake) === Illuminate\Support\Testing\Fakes\EventFake::class) {
```

mfn:
PS: I tested this in a private project, no issue (also made sure I've the `Event` Facade registered)

alesha2911:
Hi.
Sorry for inconvience solution.
These simple patch for restore dispatcher, based on 2.7.0 version https://github.com/alesha2911/laravel-ide-helper/commit/ba9e60d0a84eebd0f48a8ce990859496b7fbfe4d

And based on master branch: https://github.com/alesha2911/laravel-ide-helper/commit/ba9e60d0a84eebd0f48a8ce990859496b7fbfe4d

If you need use this version - define package in composer with pointed to these branch:
```
    "require-dev": {
        "barryvdh/laravel-ide-helper": "dev-from-2.7.0-issue-879"
    },
    "repositories": [
        {
            "type": "vcs",
            "url": "git@github.com:alesha2911/laravel-ide-helper.git"
        }
    ],
```

mfn:
@alesha2911 thanks for your contribution!

However your solution is not sufficient: the string-based matching will not work with custom dispatchers living in a different namespace, which packages like https://github.com/fntneves/laravel-transactional-events provide.

mfn (commented):
@aurelien-roy thanks for your PR. Since you didn't get back on my feedback and I figured this approach was valid, I took the liberty to finalize this myself.

@alesha2911 can you please test the recent changes if they work for your use-case?

barryvdh:
Closing due to inactivity

PR #882, opened 2020-01-20
[TESTS] Fix after recent merged "stepping on each others toes"

mfn:
Just keeping things in order here


GrahamCampbell:
Should be `int`. Not `integer`.

mfn:
> Should be `int`. Not `integer`.

But this is auto-generated, i.e. no function change, just adapted tests?

PR #889, opened 2020-02-11
change: typehint the collection type

mr-feek:
Now that phpstorm has support for generics, we can typehint that the collection returned will be of type model, rather than returning a generic collection|model[]

This should be very helpful for https://github.com/nunomaduro/larastan

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -531,10 +531,10 @@ class ModelsCommand extends Command
                                     'morphedByMany',
                                 ];
                                 if (strpos(get_class($relationObj), 'Many') !== false) {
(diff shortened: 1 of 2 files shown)


mfn:
Right after submitting my comment and re-reading your description I had an "insta-regret" feeling for a moment:
> Now that phpstorm has support for generics

However, I'm using PhpStorm 2020.1 EAP (Build #PS-201.4865.19) and I **cannot** confirm this, btw; in the below code with your changes, the `$comment->message` property is not known by PhpStorm to reference that defined attribute:
```php
 * @property-read \Illuminate\Database\Eloquent\Collection<\App\Comment> $comments
```
image

mr-feek:
I think you're right, i need to revisit this

Issue #832, opened 2019-09-26
error in  ide-helper:generate


behroozbc:
hi 
i install ide hleper in  laravel 6 and see this error 
when run 
php artisan ide-helper:generate

`> @php artisan ide-helper:generate

   ErrorException  : array_merge(): Argument #2 is not an array

  at C:\Users\behro\Documents\website\hamaiesh elmi\laravel-backend\vendor\barryvdh\laravel-ide-helper\src\Generator.php:53
    49|
    50|         // Find the drivers to add to the extra/interfaces
    51|         $this->detectDrivers();
    52|
  > 53|         $this->extra = array_merge($this->extra, $this->config->get('ide-helper.extra'));
    54|         $this->magic = array_merge($this->magic, $this->config->get('ide-helper.magic'));
    55|         $this->interfaces = array_merge($this->interfaces, $this->config->get('ide-helper.interfaces'));
    56|         // Make all interface classes absolute
    57|         foreach ($this->interfaces as &$interface) {

  Exception trace:

  1   array_merge([])
      C:\Users\behro\Documents\website\hamaiesh elmi\laravel-backend\vendor\barryvdh\laravel-ide-helper\src\Generator.php:53

  2   Barryvdh\LaravelIdeHelper\Generator::__construct(Object(Illuminate\Config\Repository), Object(Illuminate\View\Factory), Object(Illuminate\Console\OutputStyle), "")
      C:\Users\behro\Documents\website\hamaiesh elmi\laravel-backend\vendor\barryvdh\laravel-ide-helper\src\Console\GeneratorCommand.php:112
`

teeli:
Also getting this error recently. I have `extra`, `magic` and `interfaces` defined as arrays in the ide-helper config file. Still getting `ErrorException  : array_merge(): Expected parameter 2 to be an array, null given` in `vendor/barryvdh/laravel-ide-helper/src/Generator.php:53` (`$this->extra = array_merge($this->extra, $this->config->get('ide-helper.extra'));`)

Tried clearing all the caches etc, but no help.

ScriptingBeating:
I had the same error. This is how I solved it.
`php artisan vendor:publish --provider="Barryvdh\LaravelIdeHelper\IdeHelperServiceProvider" --tag=config

php artisan config:cache`

That's it

netpok:
Your config was cached, you need to clear it:

```bash
php artisan config:clear
```

Configs usually should only be cached in production.

[...]

PR #893, opened 2020-02-28
🐛 fixes ide-helper:generate array_merge error

acidjazz:
addresses issue #832

### DIFF
--- a/src/Console/GeneratorCommand.php
+++ b/src/Console/GeneratorCommand.php
@@ -83,6 +83,9 @@ class GeneratorCommand extends Command
                 'Error generating IDE Helper: first delete your compiled file (php artisan clear-compiled)'
             );
         } else {
+            // Clear our cache to avoid merging errors
+            $this->call('config:clear');
+
             $filename = $this->argument('filename');
             $format = $this->option('format');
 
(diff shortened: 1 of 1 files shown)


netpok:
I would not expect a helper library to change the application state. In my opinion if someone decides to cache the config he/she should be the one who clears those cached values.

mfn:
Ack, also 👎 on this PR

acidjazz:
@netpok @mfn yall think maybe instead a check on the array itself before the `array_merge()` happens?

mfn:
@acidjazz never encountered the error mentioned in the issue, but I see no harm if getting a config value to have a default of empty array in case it's not found (if that is what this is about?)

mazeripper:
👎 on this also—A library should change the setup of your app, but I did encounter the same issue. Maybe there should be a question of the README highlighting that cached values should be cleared?

mfn:
The readme was recently updated and now contains this in the [installation section](https://github.com/barryvdh/laravel-ide-helper#installation):
> Note: Avoid caching the configuration in your development environment, it may cause issues after installing this package; respectively clear the cache beforehand via php artisan cache:clear if you encounter problems when running the commands

I think the ought to be enough, caching development config is discouraged by all I ever read so far.

@barryvdh IMHO this PR should be dropped :}

PR #894, opened 2020-02-28
Fixed #892

### LINKED ISSUE #892: Relations with query not work again
After #873 we have missed compare all relations types.
Because  $reflection->getReturnType() return type without slash.

`\Illuminate\Database\Eloquent\Relations\MorphToMany` !== `Illuminate\Database\Eloquent\Relations\MorphToMany`

Steps to reproduce described at #760

UksusoFF:
Fixed #892


mfn:
@UksusoFF one feedback tough: the reason I broke it I think is easy: there was no test for it.

Now you changed it and no test needed adaption => I can already foresee in the future someone will break it again etc.

Suggestion: can you please add a test which, in case this would be changed (like without this PR) it would show it breaks?

thanks!

mfn:
(I doubled checked by reverting the code change and checking the failed tests)

PR #872, opened 2020-01-06
Stop calling a method static if its not static

Jamesking56:
Fixes #846

PR #897, opened 2020-03-06
newModelQuery and newQuery can't be called statically

miklcct:
as title


mfn:
This overlaps with https://github.com/barryvdh/laravel-ide-helper/pull/872/files#diff-9c76e55ce4ef570404e58c82d642c1e5R629 btw.

The other PR has another goal but uses a similar approach by modifying `setMethods` to reach this.

PR #901, opened 2020-03-25
Prevent generation of invalid code for certain parameter default values

loilo:
Certain default values for method parameters (strings containing single quotes or trailing backslashes) currently generate an invalid `_ide_helper.php` file.

This PR fixes that by replacing the current naïve code generation with a call to `var_export()`.

### DIFF
--- a/tests/MethodTest.php
+++ b/tests/MethodTest.php
@@ -49,9 +49,9 @@ class ExampleTest extends TestCase
     }
 
     /**
-     * Test special characters in method
+     * Test special characters in methods default values
      */
-    public function testSpecialChars()
+    public function testDefaultSpecialChars()
     {
         $reflectionClass = new \ReflectionClass(ExampleClass::class);
         $reflectionMethod = $reflectionClass->getMethod('setSpecialChars');
--- a/tests/MethodTest.php
+++ b/tests/MethodTest.php
@@ -47,6 +47,21 @@ class ExampleTest extends TestCase
         $this->assertEquals(['$last', '$first = \'Barry\'', '...$middle'], $method->getParamsWithDefault(false));
         $this->assertEquals(true, $method->shouldReturn());
     }
+
+    /**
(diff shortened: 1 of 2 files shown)


mfn:
Can you give an example how the code in practice looks like which triggers this?

I also think we should write tests for this

loilo:
I agree. I have to admit that I was a little bit tired when I opened the PR and didn't have the patience to grasp the test structure of the repo. 😁 

I encountered this problem when I made a facade for the [loilo/x-filesystem](https://github.com/loilo/x-filesystem) package, the problem occurs with [this method declaration](https://github.com/loilo/x-filesystem/blob/55446cd/src/XFilesystem.php#L383-L389):

```php
public function readCsvFile(
    string $filename,
    int $mode = self::PARSE_OBJECT,
    string $delimiter = ',',
    string $charset = 'UTF-8',
    string $enclosure = '"',   // <- If I change this to "'", it breaks as well
                               //    because it outputs '''
    string $escapeChar = '\\'  // <- This breaks because it outputs '\'
)
```

If you want me to, I can attach a minimal reproduction repo.

loilo:
FWIW, I've added a test for these special characters in parameter default values.

mfn:
Tanks, will take a look ASAP!

tests/MethodTest.php
```diff
@@ -47,6 +47,21 @@ public function testOutput()
         $this->assertEquals(['$last', '$first = \'Barry\'', '...$middle'], $method->getParamsWithDefault(false));
         $this->assertEquals(true, $method->shouldReturn());
     }
+
+    /**
+     * Test special characters in method
+     */
+    public function testSpecialChars()
```
mfn:
```suggestion
    /**
     * Test special characters in methods default values
     */
    public function testDefaultSpecialChars()
```

mfn:
ping @barryvdh

PR #905, opened 2020-03-31
Generate noinspections PHPStorm tags

mzglinski:
When using write option for models eq
```bash
 php artisan ide-helper:models --smart-reset --write 
```
Generated PHPDoc has fully qualified names, about which PHPStorm complains (as it should). However, this is problematic since later during automatic code cleanup tasks these names are replaced with `use` statements. This generates unnecessary changes in the files, especially if some developers forget to run code cleanup. Later on, all of this stays in the git history.

As a solution, this PR adds a new option to the model command. `--phpstorm-noinspections` eg
```bash
php artisan ide-helper:models --smart-reset --write --phpstorm-noinspections
```
This will include two PHPStorm inspections at the end of the PHPDoc comment
```
 * @noinspection PhpFullyQualifiedNameUsageInspection
 * @noinspection PhpUnnecessaryFullyQualifiedNameInspection
```

which will stop PHPStorm from replacing the autogenerated FQN with use statements.

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -58,6 +58,7 @@ class ModelsCommand extends Command
     protected $dirs = array();
     protected $reset;
     protected $keep_text;
+    protected $phpstorm_noinspections;
     /**
      * @var bool[string]
      */
@@ -96,6 +97,7 @@ class ModelsCommand extends Command
         $model = $this->argument('model');
         $ignore = $this->option('ignore');
         $this->reset = $this->option('reset');
+        $this->phpstorm_noinspections = $this->option('phpstorm-noinspections');
         if ($this->option('smart-reset')) {
             $this->keep_text = $this->reset = true;
         }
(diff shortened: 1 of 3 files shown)


mfn:
> ` * @noinspection PhpFullyQualifiedNameUsageInspection * @noinspection PhpUnnecessaryFullyQualifiedNameInspection`

Wait, why does it need both?

I'm on 2020.1 Beta and only `PhpFullyQualifiedNameUsageInspection` is recognized, the other seems not to have an impact?

mzglinski:
Sure, I will write a test tomorrow ;)

As for `PhpUnnecessaryFullyQualifiedNameInspection` please refer to https://www.jetbrains.com/help/phpstorm/php-unnecessary-fully-qualified-name.html 

This is for relations basically, they are also FQN but all models are usually in the same namespace so FQN is not necessary and PHPStorm removes their namespaces on code cleanup/code reformat.

mfn:
> As for `PhpUnnecessaryFullyQualifiedNameInspection` please refer to https://www.jetbrains.com/help/phpstorm/php-unnecessary-fully-qualified-name.html
> 
> This is for relations basically, they are also FQN but all models are usually in the same namespace so FQN is not necessary and PHPStorm removes their namespaces on code cleanup/code reformat.

Thanks for the details!

I think it would be great if you could add this as a code comment with the block `if ($this->phpstorm_noinspections) {` so it's clear as to why (adding the link would be also nice).

thanks

mzglinski:
Is there any chance for this to be merged in the near future?

PR #906, opened 2020-04-01
Infer return type from reflection if no phpdoc given

mfn:
Up until now, to have the `get…Attribute` return a useful type information in the class phpdoc, it was **necessary** to specify the `@return` annotation; otherwise the result was always `mixed`.

This now adds the capability to derive the type from the reflection itself.

- Example:
  ```php
  function getFooAttribute(): string
  {
  }
  ```
- before ` * property-read mixed $foo`
- after ` * property-read string $foo`

Note that the docblock is still the *authoritative* source _if_ provided. The reason is that phpdoc expressiveness / supporting tooling still exceeds PHPs type capabilities.

But this also means if the docblock is wrong, so will the type be and this PR won't fix that.

OTOH it removes the need for a docblock for the simple cases (e.g. no iterables of a type).

Tests with lots-o-variations included.

This is in fact already implemented for the `elseif (!method_exists('Illuminate\Database\Eloquent\Model', $method)` case, but this part in fact _prefers_ the native type over the docblock; not sure why and didn't want to change that for now.

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -10,19 +10,19 @@
 
 namespace Barryvdh\LaravelIdeHelper\Console;
 
+use Barryvdh\Reflection\DocBlock;
+use Barryvdh\Reflection\DocBlock\Context;
+use Barryvdh\Reflection\DocBlock\Serializer as DocBlockSerializer;
+use Barryvdh\Reflection\DocBlock\Tag;
 use Composer\Autoload\ClassMapGenerator;
 use Illuminate\Console\Command;
 use Illuminate\Database\Eloquent\Relations\Relation;
-use Illuminate\Support\Str;
 use Illuminate\Filesystem\Filesystem;
+use Illuminate\Support\Str;
 use ReflectionClass;
-use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
-use Barryvdh\Reflection\DocBlock;
-use Barryvdh\Reflection\DocBlock\Context;
-use Barryvdh\Reflection\DocBlock\Tag;
-use Barryvdh\Reflection\DocBlock\Serializer as DocBlockSerializer;
 
 /**
  * A command to generate autocomplete information for your IDE
@@ -438,7 +438,7 @@ class ModelsCommand extends Command
                     $name = Str::snake(substr($method, 3, -9));
                     if (!empty($name)) {
                         $reflection = new \ReflectionMethod($model, $method);
-                        $type = $this->getReturnTypeFromDocBlock($reflection);
+                        $type = $this->getReturnType($reflection);
                         $this->setProperty($name, $type, true, null);
                     }
                 } elseif (Str::startsWith($method, 'set') && Str::endsWith(
@@ -814,6 +814,16 @@ class ModelsCommand extends Command
         return $this->laravel['config']->get('ide-helper.model_camel_case_properties', false);
     }
 
+    protected function getReturnType(\ReflectionMethod $reflection): ?string
+    {
+        $type = $this->getReturnTypeFromDocBlock($reflection);
+        if ($type) {
+            return $type;
+        }
+
+        return $this->getReturnTypeFromReflection($reflection);
+    }
+
     /**
      * Get method return type based on it DocBlock comment
      *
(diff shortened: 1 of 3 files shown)


src/Console/ModelsCommand.php
```diff
@@ -814,6 +814,16 @@ protected function hasCamelCaseModelProperties()
         return $this->laravel['config']->get('ide-helper.model_camel_case_properties', false);
     }
 
+    protected function getReturnType(\ReflectionMethod $reflection): ?string
+    {
+        $type = $this->getReturnTypeFromDocBlock($reflection);
+        if ($type) {
+            return $type;
+        }
```
mfn, line 822: As mentioned, same behaviour as before: prefer docblock

tests/Console/ModelsCommand/Getter/Models/Simple.php
```diff
@@ -47,11 +47,58 @@ public function getAttributeReturnTypeIntOrNullAttribute(): ?int
     {
     }
 
-    public function getAttributeReturnsImportedClass(): DateTime
+    public function getAttributeReturnsImportedClassAttribute(): DateTime
     {
     }
 
-    public function getAttributeReturnsFqnClass(): \Illuminate\Support\Facades\Date
+    public function getAttributeReturnsFqnClassAttribute(): \Illuminate\Support\Facades\Date
```
mfn, line 54: Bugfix of my own tests 😅

tests/Console/ModelsCommand/Getter/Test.php
```diff
@@ -62,12 +62,25 @@ public function test(): void
  * Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\Getter\Models\Simple
  *
  * @property integer $id
- * @property-read mixed $attribute_return_type_int_or_null
+ * @property-read int|null $attribute_return_type_int_or_null
```
mfn, line 65: This shows the improvement…

tests/Console/ModelsCommand/Getter/Test.php
```diff
@@ -62,12 +62,25 @@ public function test(): void
...
+ * @property-read bool|null $attribute_returns_nullable_float
+ * @property-read \stdClass|null $attribute_returns_nullable_std_class
+ * @property-read \stdClass $attribute_returns_std_class
+ * @property-read void $attribute_returns_void
  * @property-read \what|\ever|\we-write/here $attribute_takes_phpdoc_literal
  * @property-read int $attribute_with_int_return_phpdoc
  * @property-read string $attribute_with_int_return_type_and_but_phpdoc_string
  * @property-read int $attribute_with_int_return_type_and_phpdoc
- * @property-read mixed $attribute_with_int_return_type
+ * @property-read int $attribute_with_int_return_type
```
mfn, line 83: …this one as well

mfn:
Not sure why travis doesn't show up integrated here anymore, PR tests are executed => https://travis-ci.org/github/barryvdh/laravel-ide-helper/builds/669882120

mfn:
ping @barryvdh thanks for considering 🙇

belamov:
I don't think that documentation should be the authoritative source of truth and we should always prioritize code itself and fallback to documentation if no return type provided

mfn:
> I don't think that documentation should be the authoritative source of truth and we should always prioritize code itself and fallback to documentation if no return type provided

Unfortunately this will not work in practice because PHPs type system is inferior in terms of what types and combinations you can express.

Thereof it's my opinion the only sensible thing to do it to make the phpdoc authoritative!

tests/Console/ModelsCommand/Getter/Models/Simple.php
belamov: will fix

belamov:
i guess your talking about something like this
```php
/**
 * @return \App\Models\Post[]
 */
public function allPosts(): array
{
    // ...
}
```

didn't think of this, it makes sense in this example

mfn:
Yes, exactly!

splinter89:
Wow, thanks @mfn, this is exactly what I was looking for.
@barryvdh Please check this out :pray:
These "mixed" are killing me, and if I manually fix them, I can't use "--reset" option anymore.

mfn:
>  These "mixed" are killing me, and if I manually fix them, I can't use "--reset" option anymore.

You can add an explicit function phpdoc `@return`, then it works also with reset (with that I mean you don't / shouldn't edit it directly in the class phpdoc)

But yeah, having this PR merged would be great 😄

mfn:
Friendly ping @barryvdh 🤗

mr-feek:
thanks @mfn , this will be nice!

mfn:
It is! Already deleted lots of now-redundant phpdocs in my projects

PR #910, opened 2020-04-08
Refactor checking a lumen version

pkboom:
Makes checking a lumen version compact.


mfn:
Personal opinion: don't see a need to change perfectly fine working code for the sake of change 👎 

Also, this code part has not tests, so …

PR #916, opened 2020-04-15
Add support for facades which resolve to classes that are sometimes not bound in the container

stancl:
Context: https://github.com/stancl/tenancy/issues/353#issuecomment-612906249

Didn't run any tests so hopefully this doesn't break anything.

### DIFF
--- a/src/Alias.php
+++ b/src/Alias.php
@@ -271,7 +271,11 @@ class Alias
         try {
             //If possible, get the facade root
             if (method_exists($facade, 'getFacadeRoot')) {
-                $root = get_class($facade::getFacadeRoot());
+                if (is_object($facade::getFacadeRoot())) {
+                    $root = get_class($facade::getFacadeRoot());
+                } else {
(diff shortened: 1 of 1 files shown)


stancl:
@mfn 

The test would be that I didn't break anything & this now works:

```php
class Foo {}

class FooFacade extends Facade
{
    public function getFacadeAccessor()
    {
        return Foo::class;
    }
}
```

An alias should be generated even if no specific instance of `Foo` is currently bound in the container.

mfn:
Can you add this as a test? So no one else breaks "your" change :)

stancl:
Seems like my solution doesn't fix the bug: https://github.com/stancl/tenancy/issues/353#issuecomment-614434837

Will look deeper into it once I get the chance.

barryvdh:
So is this required or not?

stancl:
This will require some more work probably. Will open a new PR if I get to it.

Issue #917, opened 2020-04-17
forceDelete() and restore() should not be static


Krunch:
When using the `use SoftDeletes;` trait `ide-helper:model` will add these 2 `@method` in the phpdoc:

```
 * @method static bool|null forceDelete()
 * @method static bool|null restore()
```

Those methods should not be static as they are not defined as static inside the trait.

mfn:
Hmm.

I think they should be simply removed, specifically these two lines https://github.com/barryvdh/laravel-ide-helper/blob/18588eacf3028c636ef2079b0c868b7a03c0683d/src/Console/ModelsCommand.php#L844-L845

In my opinion any decent editor/IDE/static code analyzer can infer methods from traits automatically anyway, no need to duplicate this (not so with the other because they're magic mixing in the phpdoc and are fine).

mfn:
Made https://github.com/barryvdh/laravel-ide-helper/pull/918

Krunch:
You are right, I am using phpstorm and those methods are auto-completed fine without those 2 lines. Thank you.

PR #918, opened 2020-04-17
Remove unnecessary and wrong definition of SoftDelete methods

mfn:
These two methods are already regular methods on the trait, there's no
need to add them and additionally they're not static, so their
definition was changed for the worse.

The other methods are magic so it's fine to keep them.

Fixes https://github.com/barryvdh/laravel-ide-helper/issues/917


barryvdh:
Thanks!

PR #920, opened 2020-04-20
if a model implements interfaces, include them in the stub

mr-feek:
We've been running into this over at https://github.com/psalm/psalm-plugin-laravel/issues/21, and I think this might be appropriate to include here?

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -660,6 +660,10 @@ class ModelsCommand extends Command
         $classname = $reflection->getShortName();
         $originalDoc = $reflection->getDocComment();
         $keyword = $this->getClassKeyword($reflection);
+        $interfaceNames = array_diff_key(
+            $reflection->getInterfaceNames(),
+            $reflection->getParentClass()->getInterfaceNames()
+        );
 
         if ($this->reset) {
             $phpdoc = new DocBlock('', new Context($namespace));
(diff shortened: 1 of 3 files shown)


mr-feek:
Sure thing RE a test, I should be able to get to it in a few days.
-----
That is also an excellent question. I'm relatively new to psalm, so I don't know the answer to that. I do believe phpstan only reads the docblocks, and ignores any other definitions (ie interfaces) in the stub file. 

From my initial investigation, it seemed to me that missing these interfaces was causing psalm to "forget" the interfaces from the model definitions.  I'd love to play around  by pulling this commit into my local project to test if it fixes it, I'll do that later this week and let you know what  i find :)

mfn:
> Psalm has the capability of reading phpstorm

Ah!

phpstan and insofer larastan don't do that, only the sources are directly are analyzed.

However: larastan (phpstan?) will take advantage of the phpdocs generated into the model class files.

mr-feek:
So this did help out psalm, I got rid of several errors locally. I'll go ahead and get a test written, as there is two bugs I found :)

mr-feek:
Let me know if there's anything else I can do

src/Console/ModelsCommand.php
```diff
@@ -660,6 +660,10 @@ protected function createPhpDocs($class)
         $classname = $reflection->getShortName();
         $originalDoc = $reflection->getDocComment();
         $keyword = $this->getClassKeyword($reflection);
+        $interfaceNames = array_diff_key(
+            $reflection->getInterfaceNames(),
+            $reflection->getParentClass()->getInterfaceNames()
+        );
```
mfn, line 666: What does this dark magic thing do? 🧙‍♂️ :)

mr-feek: So `ReflectionClass::getInterfaceNames` get's the names of every interface that the underlying class satisfies. Since we only _need_ to  document this models interfaces (the parent interfaces automatically are implemented via inheritance), I find the diff to know what interfaces this model explicitly implements. We could go ahead and list out all of them, but it isn't necessary.  `getInterfaceNames` always returns an array, so this should be safe.

mfn: Purrfect, thank you!

barryvdh:
Thanks!

PR #924, opened 2020-04-22
Add static return type to builder methods

### LINKED ISSUE #923: Custom Collections do not work with static query buillder methods
PHP: 7.4.4
Package: 2.7.0

**Testing Successful**
Just upgraded and checked https://github.com/barryvdh/laravel-ide-helper/pull/903 in a few applications and we get custom collection code completion on

```php
Foo::get()
Foo::all()
```

Also if you build a query using `query` then you do get code completion:

```php
Foo::query()->where('bar', 'baz')->get() // completes to custom collection
```

**Problem**
But if I build a query directly from the model there is no completion even though it does return the custom collection instance 😿.

```php
Foo::where('bar', 'baz')->get() // completes to Builder[]|Collection 
```

**Cause**
The ide-helper.php file generates the static where functions etc but is misssing `static` on some of the return types. 

_ide-helper.php_
```php
/**
* Add a basic where clause to the query.
*
* @param \Closure|string|array $column
* @param mixed $operator
* @param mixed $value
* @param string $boolean
* @return \Illuminate\Database\Eloquent\Builder 
* @static 
*/ 
public static function where($column, $operator = null, $value = null, $boolean = 'and')
{
    return $instance->where($column, $operator, $value, $boolean);
}
```

**Solution**
Add the `static` return type to all the relevant builder functions `with`, `where*` etc. 

_ide-helper.php_
```php
@return \Illuminate\Database\Eloquent\Builder|static
```

dmason30:
Fixes #923 

So this is an attempt to add `static` to the return type on the builder methods generated by `ide-helper:generate`. I am not 100% sure if this is the right location for this but it wasn't immediately obvious where else it could go. 

I have tested this on a number of my projects and it works as expected.

_ide-helper.php_
```php
* @return \Illuminate\Database\Eloquent\Builder|static
...
public static function where($column, $operator = null, $value = null, $boolean = 'and')
```

Unfortunately the Generator command doesn't have any tests - which is understandable due to its complexity. But I have added a test for the `Method` class I have changed

### DIFF
--- a/tests/MethodTest.php
+++ b/tests/MethodTest.php
@@ -3,6 +3,7 @@
 namespace Barryvdh\LaravelIdeHelper\Tests;
 
 use Barryvdh\LaravelIdeHelper\Method;
+use Illuminate\Database\Eloquent\Builder;
 use PHPUnit\Framework\TestCase;
 
 class ExampleTest extends TestCase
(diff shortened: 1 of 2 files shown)


tests/MethodTest.php
```diff
@@ -48,6 +49,33 @@ public function testOutput()
...
+ * @static 
+ */';
+        $this->assertEquals($output, $method->getDocComment(''));
+        $this->assertEquals('with', $method->getName());
+        $this->assertEquals('\\'.Builder::class, $method->getDeclaringClass());
+        $this->assertEquals('$relations', $method->getParams(true));
+        $this->assertEquals(['$relations'], $method->getParams(false));
+        $this->assertEquals('$relations', $method->getParamsWithDefault(true));
+        $this->assertEquals(['$relations'], $method->getParamsWithDefault(false));
+        $this->assertEquals(true, $method->shouldReturn());
```
mfn, line 76: I would suggest to use `assertSame` everywhere, don't see a reason for not using the strictes possible check here

dmason30: @mfn Yes I use `assertSame` on all my projects this was just a copy and paste of the other output test. I have changed the whole test to now use `assertSame` for consistency.

dmason30:
@mfn This not only fixes custom collections but will also gives autocompletion custom model scopes/methods after using a static builder method e.g:

```php
Foo::where('bar', 'baz')->customTypeScopeMethod($hintedArgs)->with(...) // Etc
```

tests/MethodTest.php
mfn: I know what you mean 😄 

I always try to be bit of boy scout here and trying to leave the place in a better shape I found it… can't help this OCD 🤷‍♀️

barryvdh:
Thanks!

PR #921, opened 2020-04-21
Ability to use patterns for model_locations

4n70w4:
Example: app/Services/*/Models

PR #925, opened 2020-04-22
Document that model_locations/dir now supports glob

mfn:
Feature was added with https://github.com/barryvdh/laravel-ide-helper/pull/921

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -89,6 +89,9 @@ return array(
     | Define in which directories the ide-helper:models command should look
     | for models.
     |
+    | glob patterns are supported to easier reach models in sub-directories,
+    | e.g. `app/Services/* /Models` (without the space)
+    |
     */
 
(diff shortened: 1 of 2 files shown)


config/ide-helper.php
```diff
@@ -89,6 +89,9 @@
     | Define in which directories the ide-helper:models command should look
     | for models.
     |
+    | glob patterns are supported to easier reach models in sub-directories,
+    | e.g. `app/Services/* /Models` (without the space)
```
barryvdh, line 93: So why the space?

mfn: Yeah well :) because otherwise it would end the phpdoc 💥

mfn: And I tried (maybe too hard) to give a useful example

barryvdh: hehe good point.

PR #934, opened 2020-05-15
Fixed error reading dynamic properties

andrey-helldar:
Now, when generating, the `get_class_methods` method is used, which does not return `protected` methods from the model, so it is not possible to get the correct list of model properties:

Using `ReflectionClass` instead, we can get all the methods available in the model and generate the correct doc-block:

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -443,10 +443,7 @@ class ModelsCommand extends Command
         $methods = $reflection->getMethods();
         if ($methods) {
             usort($methods, function ($a, $b) {
-                if ($a->name == $b->name) {
-                    return 0;
-                }
-                return ($a->name < $b->name) ? -1 : 1;
+                return $a->name <=> $b->name;
             });
             foreach ($methods as $reflection_method) {
                 $method = $reflection_method->name;
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -439,10 +439,17 @@ class ModelsCommand extends Command
      */
     protected function getPropertiesFromMethods($model)
     {
-        $methods = get_class_methods($model);
+        $reflection = new ReflectionClass($model);
+        $methods = $reflection->getMethods();
         if ($methods) {
-            sort($methods);
-            foreach ($methods as $method) {
+            usort($methods, function ($a, $b) {
+                if ($a->name == $b->name) {
+                    return 0;
+                }
+                return ($a->name < $b->name) ? -1 : 1;
+            });
+            foreach ($methods as $reflection_method) {
+                $method = $reflection_method->name;
                 if (Str::startsWith($method, 'get') && Str::endsWith(
                     $method,
                     'Attribute'
@@ -458,7 +465,7 @@ class ModelsCommand extends Command
                 } elseif (Str::startsWith($method, 'set') && Str::endsWith(
                     $method,
                     'Attribute'
-                ) && $method !== 'setAttribute'
+                ) && $method !== 'setAttribute' && $method !== 'setClassCastableAttribute'
                 ) {
                     //Magic set<name>Attribute
                     $name = Str::snake(substr($method, 3, -9));
--- a/tests/Console/ModelsCommand/Getter/Models/Simple.php
+++ b/tests/Console/ModelsCommand/Getter/Models/Simple.php
@@ -101,4 +101,99 @@ class Simple extends Model
     public function getAttributeReturnsVoidAttribute(): void
     {
     }
+
+    /**
+     * @return int
+     */
+    protected function getProtectedAttributeWithIntReturnPhpdocAttribute()
+    {
+    }
+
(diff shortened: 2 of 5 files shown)


src/Console/ModelsCommand.php
```diff
@@ -439,10 +439,17 @@ protected function getPropertiesFromTable($model)
...
+        $reflection = new ReflectionClass($model);
+        $methods = $reflection->getMethods();
         if ($methods) {
-            sort($methods);
-            foreach ($methods as $method) {
+            usort($methods, function ($a, $b) {
+                if ($a->name == $b->name) {
+                    return 0;
+                }
+                return ($a->name < $b->name) ? -1 : 1;
```
mfn, line 449: I belief this can be simplified withL
```suggestion
                return $a->name <=> $b->name;
```

andrey-helldar:
> Strange use-case 😄

Another use case: there are `first_name` and `last_name` columns in the database and in the code:
```
protected function getFullNameAttribute()
{
  return $this->first_name . ' ' . $this->last_name;
}

// $item->full_name
```

andrey-helldar:
@mfn, Laravel [suggests](https://laravel.com/docs/7.x/eloquent-mutators#defining-an-accessor) creating public methods for mutators. But I think this is bad because during development the IDE will display both methods and properties, which will increase both development time and will show unnecessary methods. Therefore, it is better to write them `protected` or `private`.

Tests added 🙂

andrey-helldar:
> My assumption for internal methods would be to just avoid the magic but hey.
It sounds interesting, but there are some nuances. For example, when transferring the model somewhere in the form of an array, you need to save the result of such methods.

For example, in resources it will be prettier to write like this:
```
return [
  'foo' => $this->foo,
  'bar' => $this->bar,
  'baz' => $this->baz,
];
```

than so:
```
return [
  'foo' => $this->foo,
  'bar' => $this->getBarAttribute(),
  'baz' => $this->baz,
];
```

src/Console/ModelsCommand.php
```diff
@@ -458,7 +462,7 @@ protected function getPropertiesFromMethods($model)
                 } elseif (Str::startsWith($method, 'set') && Str::endsWith(
                     $method,
                     'Attribute'
-                ) && $method !== 'setAttribute'
+                ) && $method !== 'setAttribute' && $method !== 'setClassCastableAttribute'
```
mfn, line 465: I just realized why this exclusion of this method is in :-/

andrey-helldar: This is the [standard Laravel](https://github.com/laravel/framework/blob/7.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L725-L751) method in `protected` visibility required for custom attribute handling.

I doubt that it needs to be added to the helper and even in the form of `class_castable`.

andrey-helldar: If necessary, I can remove the exception of this method from the code.
But this is the standard [Laravel](https://github.com/laravel/framework/blob/7.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L725-L751) method in protected visibility required for custom attribute handling.

mfn: I think it's fine and makes sense.

Only other way to "fix" this would be to change the name in Laravel.

andrey-helldar: I doubt that Taylor will agree to this)

mfn: Me too 😄

mfn:
I'm not sure about this PR anymore. This is quite an interesting angle, but as I said: I'm not sure.

Indeed, making those getters/setter `protected` in fact works. It never crossed my mind, so props for this 👍 

Important technical bit: your change iterates of all methods, not just public ones. It's possible to write a `private function getFooAttribute` and it would make a `@property mixed $foo` but **private** does **not** work (I just tested it). Maybe this should be taking into consideration?

I've never seen this in the wild though. And the official docs don't mention this (at least now on the page you also linked to; as you said they give only example with `public`). What if they make some awesome internal change and accidentally break this? Unlikely, but…

---

I might have more technical feedback on the PR and there would be some changes necessary due to recent refactoring, but I think it's best if @barryvdh also gives feedback whether he wants to accept this (IMHO) esoteric approach… (maybe this whole thing is too clever for me and eludes my practical point of view 😄 )

andrey-helldar:
I apologize for the long reply. Today I can check and make adjustments.

mfn:
I'm not convinced that endorsing `protected` attribute getter would be smart for this library.

Never seen this before, to me this sounds like creative solution to a problem outside of this library and I would vote rather not including this.

@barryvdh what do you think?

barryvdh:
I think it's a bit of an edge-case. Not really sure what problems could arise (eg. with existing protected methods)

andrey-helldar:
Magic Laravel allows you to get significance despite the level of the protection of the method. At the same time, to make a method public, in my opinion, too much, as the IDE will suggest them when choosing, which will be superfluous due to the number of information.
Of course, if I understood you correctly.

PR #936, opened 2020-05-16
#935 Using 'self' instead of FQCN on Models

### LINKED ISSUE #935: Using 'self' instead of FQCN in models helper generator's @method definitions
Why we don't use just `self` instead of `\App\User\SocialNetwork` (or another model namespace) ?

We can do this because, model definition is related to current class.

So,

`@method static \Illuminate\Database\Eloquent\Builder|\App\User\SocialNetwork whereCreatedAt( $value )`

must replaced with

`@method static \Illuminate\Database\Eloquent\Builder|self whereCreatedAt( $value )`

I will try to create PR when I will available

samirmh-dev:
Related issues: closes #935

If we use `-W` option it will write docs to the classes directly. At this time we need to replace the `Fully Qualified Class Name` with self. 
If we don't use `-W` option, it means, we using helper file instead of writing Models itself. At this time we don't need to replace FQCN with `self`


mfn:
> At this time we need to replace the Fully Qualified Class Name with self.

…

> At this time we don't need to replace FQCN with `self`

…

IMHO we should not do that. What was used to be a single clear code path not becomes entangles in behaviour, depending on whether it's written within the class or the external one.

The general code already has so many cases, I'm on 👎 changing this at the moment.

samirmh-dev:
> IMHO we should not do that. What was used to be a single clear code path not becomes entangles in behaviour, depending on whether it's written within the class or the external one.
> 
> The general code already has so many cases, I'm on 👎 changing this at the moment.

thanks for feedback. You are right on most points @mfn. I just wanted to share it with community :) 

> > At this time we need to replace the Fully Qualified Class Name with self.
> 
> …
> 
> > At this time we don't need to replace FQCN with `self`
> 
> …
> 

I mean when using `-W` parameter we need to replace FQCN with `self`. And if we don't used `-W` parameter replacing is not necessarily 🤔 sorry for fast typo combined with bad english 😂

mfn:
> I mean when using `-W` parameter we need to replace FQCN with `self`. And if we don't used `-W` parameter replacing is not necessarily 🤔 sorry for fast typo combined with bad english 😂

It's fine, I understood it 👍 I just didn't agree with it 🤷‍♀️

mr-feek:
I don't think there is much of a benefit to this, other than conciseness! I also agree that reducing code complexity should probably be the priority here, therefore I am also 👎 If the codebase were a bit less spaghetti, this would probably be welcome 😅

mfn:
@samirmh-dev is you use-case maybe solved with https://github.com/barryvdh/laravel-ide-helper/pull/954 which was merged recently?

samirmh-dev:
@mfn hey 👋 yeah it looks like already done. We can close issue I created 👍

mfn:
@barryvdh PR can be dropped :}

Issue #938, opened 2020-05-18
assertRegExp() will be removed on PHPUnit 10


samirmhsnv:
> assertRegExp() is deprecated and will be removed in PHPUnit 10

We need to replace it with **assertMatchesRegularExpression()** .

PR #939, opened 2020-05-18
#938 Replaced assertRegExp() with assertMatchesRegularExpression

samhsnv:
Closes: #938

### DIFF
--- a/tests/Console/EloquentCommandTest.php
+++ b/tests/Console/EloquentCommandTest.php
@@ -41,7 +41,7 @@ class EloquentCommandTest extends TestCase
         $tester = $this->runCommand($command);
 
         $expectedContent = '/**
- * 
+ *
  *
  * @mixin \Eloquent
  * @mixin \Illuminate\Database\Eloquent\Builder
@@ -51,8 +51,8 @@ abstract class Model implements';
         $this->assertSame($expectedContent, $actualContent);
 
         $display = $tester->getDisplay();
-        $this->assertRegExp(';Unexpected no document on Illuminate\\\Database\\\Eloquent\\\Model;', $display);
-        $this->assertRegExp(';Wrote expected docblock to .*/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php;', $display);
+        $this->assertMatchesRegularExpression(';Unexpected no document on Illuminate\\\Database\\\Eloquent\\\Model;', $display);
+        $this->assertMatchesRegularExpression(';Wrote expected docblock to .*/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php;', $display);
     }
 
     private function getVendorModelFilename(): string
--- a/tests/Console/EloquentCommandTest.php
+++ b/tests/Console/EloquentCommandTest.php
@@ -51,8 +51,14 @@ abstract class Model implements';
         $this->assertSame($expectedContent, $actualContent);
 
         $display = $tester->getDisplay();
-        $this->assertMatchesRegularExpression(';Unexpected no document on Illuminate\\\Database\\\Eloquent\\\Model;', $display);
-        $this->assertMatchesRegularExpression(';Wrote expected docblock to .*/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php;', $display);
+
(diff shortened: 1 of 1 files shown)


tests/Console/EloquentCommandTest.php
```diff
@@ -41,7 +41,7 @@ public function testCommand()
         $tester = $this->runCommand($command);
 
         $expectedContent = '/**
- * 
+ *
```
mfn, line 44: It seems the build broke due to this => https://travis-ci.org/github/barryvdh/laravel-ide-helper/builds/688241498

Please revert the change in this line

lupinitylabs: I had a similar problem when adding a test case to the MethodTest.php. PHPStorm removes trailing whitespaces automatically, and I haven't found a way to make it stop this. However, I think it would be a better idea to explicitly put those whitespaces instead of hoping that code formatting will leave them in place. Thus, maybe better do

```
$expectedContent = "/**\n";
$expectedContent .= " * \n";
...
```

instead?

samirmh-dev: > I had a similar problem when adding a test case to the MethodTest.php. PHPStorm removes trailing whitespaces automatically, and I haven't found a way to make it stop this. However, I think it would be a better idea to explicitly put those whitespaces instead of hoping that code formatting will leave them in place. Thus, maybe better do
> 
> ```
> $expectedContent = "/**\n";
> $expectedContent .= " * \n";
> ...
> ```
> 
> instead?
> 
> (I must admit, however, that I find it weird to test for whitespaces to trail some lines and not others in the first place...)

Great idea!

mfn: Hmm. So, hello there, I came up with the tests in the first place, blame's on me 😄 

I was thinking about this for a second, but came to the following conclusion: if your editor is automagically trimming EOL spaces _inside_ string, you might want to reconsider this.

EOL whitespaces outside strings is insignificant, whereas inside, it is significant.

Of course you can fix the _generator_ here to not add the EOL whitespace which may end up in the actual code, but this should be a separate change.

lupinitylabs: Hey! No blame :-)

You are absolutely right, it is odd that trailing spaces within multi line string literals are subject to modification in any way. However, a variety of IDEs seem to do this and my issue with the multi line string literal are not necessarily the whitespaces, but their existence in the first place. The phpdoc start (/**) does not have any trailing whitespaces, the next line (short description) does, the next line (long description) does not. Then all the parameters don't have trailing whitespaces, the return type does have a trailing whitespace, and so does the @static. The closing line then does not have a trailing whitespace again.

What I am trying to say is that it just seems a little inconsistent to me, and I think, just as you mentioned, the best way might be to get rid of all trailing whitespaces in phpdoc generation, as they don't really serve a useful purpose and are certainly not important enough to test for their existence...

Just a thought.

samirmh-dev:
@mfn I fixed that whitespace problem. Currently build failing for PhpUnit version

mfn:
Yeah, so, as long as this package doesn't drop older Laravel versions we also have to support older PHPUnit versions not knowing about the methods ;-)

Either you come up with a fix for that or we need a decision in general here first: until which version shall we support L5?

https://laravel.com/docs/master/releases

Basically it's L5.5 until August this year, everything else below L6 has not even security fixes so I personally also wouldn't bother supporting them.

[I don't know what happens to phpunit; years ago upgrades were smooth sailing but in "recent times" every release has so many breaking changes it's no fun anynore]

tests/Console/EloquentCommandTest.php
mfn: FTR, I believe this now moot as https://github.com/barryvdh/laravel-ide-helper/pull/958 was merged basically all tests were switched to snapshots; I think this problem is now solved anyway

barryvdh:
It's more about PHP versions, not Laravel versions.

tests/Console/EloquentCommandTest.php
lupinitylabs: > FTR, I believe this now moot as #958 was merged basically all tests were switched to snapshots; I think this problem is now solved anyway

Looking good :+1:

Issue #942, opened 2020-05-22
Option for generating `Collection|ElementType[]` types as `Collection<int, ElementType>` instead


bmewburn:
Many static analysis tools now support generic notation for `Iterator` or `ArrayAccess` implementations. In this situation `Collection|ElementType[]` is not useful as it does not convey the real type and makes workarounds necessary within static analysis tools to correct for this. Is there an option to generate the stubs in generic notation? Or would it be possible to add an option?

mr-feek:
I’d be willing to help implement this if a PR would be welcome. I’ve been reverse engineering a lot of this logic for psalm

mfn:
We need to be careful because whilst this improves the static analysis tools, this will reduce usefulness of the typehints for IDEs. (*)

Remember: the primary goal is for IDEs to help here :-)

It's only recently that tools like https://github.com/nunomaduro/larastan embrace this information too and in fact expand upon it.

(*) PhpStorm being the most popular one, I know that this syntax `Collection<int, ElementType>` _currently_ will not yield PhpStorm understanding the this is the same as `ElementType[]`

mr-feek:
@bmewburn I'm interested in what issue you are  running into in which you need this option first-party in laravel-ide-helper.

In larastan, I wrote a parser that treats `Collection|Model[]` as `Collection<Model>` here: https://github.com/nunomaduro/larastan/pull/479

And psalm can infer these types by setting`allowPhpstormGenerics='true'` in your config  https://psalm.dev/docs/running_psalm/configuration/#allowphpstormgenerics


------

@mfn I totally understand your point about this being primarily an **ide-helper**. I just wanted to point  out that psalm has a LSP implemented, which means that (with the LSP plugin) phpstorm direcly benefits from the improvements we are making in static analysis tooling!

mfn:
> phpstorm direcly benefits from the improvements we are making in static analysis tooling

I don't follow this as PhpStorm doesn't support this proposed syntax change? Or am I missing something? 🤔

mr-feek:
@mfn check this out: https://psalm.dev/docs/running_psalm/language_server/ 

It's pretty neat. 

----

Don't mean to derail this thread. I think ide-helper should continue to write the phpstorm syntax, as devs shouldn't be forced to use the likes of psalm.

bmewburn:
[...]

Issue #756, opened 2019-01-23
Using ide-helper outside the framework with just illuminate/database


mfn:
I'm creating this issue because maybe it helps others which are interested in a similar solution and maybe there's interest in making the ide-helper code slightly more agnostic.

My scenario:
- I'm not using the Laravel "framework" directly
- I only use https://github.com/illuminate/database (and https://github.com/illuminate/container ) for integrating Eloquent in a legacy code base
  If you never heard about this, you wouldn't know this really works and you _can use_ Eloquent outside the framework.

But working with models without ide-helper is very annoying (also for static analysis like phpstan) so I was seeking a solution to make it possible to use ide-helper for two things:
- models phpdoc generation
- metadata generation for PhpStorm

I did not evaluate the generation of Facade data as I'm not using them.

### To make the `models` command work
- use of `base_path()` global helper from the framework
  In `\Barryvdh\LaravelIdeHelper\Console\ModelsCommand::loadModels` it uses the global helper `base_path()` which isn't available outside of the framework (it's defined in [vendor/laravel/framework/src/Illuminate/Foundation/helpers.php](https://github.com/laravel/framework/blob/5f2bf799a50bd6fbd8561c125db8d8b41a5af6fd/src/Illuminate/Foundation/helpers.php#L191))
  - My solution was to write my own global `base_path()` which I only load in the context of generating the models which creates me the correct path
- Use of the `Config` service
  Multiple times the `$this->laravel['config']` service is access to load the configuration. However I'm not using `illuminate/config` either. I've created a one-shot anonymous class with my code which emulates the most basic methods to make this work. I used below code and and bound it to my container via `$laravelContainer->instance('config', …)`; in below code `Arr::get` can be used because of the dependencies of other packages to `illuminate/support`:
  ```php
          return new class($config) implements ArrayAccess {
              /** @var array */
              private $config;
  
              public function __construct(array $config)
              {
                  $this->config = $config;
              }
  
              public function get($key, $defaultVaue = null)
              {
                  return Arr::get($this->config, $key, $defaultVaue);
              }
  
              public function offsetExists($offset)
              {
                  throw new RuntimeException('Not implemented because we didn\'t need it yet');
              }
  
              public function offsetGet($offset)
              {
                  return $this->get($offset);
              }
  
              public function offsetSet($offset, $value)
              {
                  throw new RuntimeException('Not implemented because we didn\'t need it yet');
              }
  
              public function offsetUnset($offset)
              {
[...]

PR #945, opened 2020-05-28
Model hooks

wimski:
In the last couple of Laravel projects I did, I needed extra IDE help for models (translated properties, attachments, etc.). There currently is no way to generate custom information so I wrote a model hook system. Everyone can then easily extend the generation of model information. In order to make it  actually useful, I changed some methods in the models command from protected to public.

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -146,6 +146,21 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Models hooks
+    |--------------------------------------------------------------------------
+    |
+    | Define which hook classes you want to run for models to add custom information
+    |
+    | Hooks should implement Barryvdh\LaravelIdeHelper\Contracts\ModelHookInterface.
+    |
+    */
+
+    'model_hooks' => array(
+        // App\Support\IdeHelper\MyModelHook::class
+    ),
+
+    /*
+    |--------------------------------------------------------------------------
     | Extra classes
     |--------------------------------------------------------------------------
     |
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -155,9 +155,9 @@ return [
     |
     */
 
-    'model_hooks' => array(
+    'model_hooks' => [
         // App\Support\IdeHelper\MyModelHook::class
-    ),
+    ],
 
     /*
     |--------------------------------------------------------------------------
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -1363,12 +1363,12 @@ class ModelsCommand extends Command
      */
     protected function runModelHooks($model): void
     {
-        $hooks = $this->laravel['config']->get('ide-helper.model_hooks', array());
+        $hooks = $this->laravel['config']->get('ide-helper.model_hooks', []);
 
         foreach ($hooks as $hook) {
             $hookInstance = $this->laravel->make($hook);
 
-            if (! $hookInstance instanceof ModelHookInterface) {
+            if (!$hookInstance instanceof ModelHookInterface) {
                 throw new \RuntimeException(
                     'Your IDE helper model hook must implement Barryvdh\LaravelIdeHelper\Contracts\ModelHookInterface'
                 );
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -11,6 +11,7 @@
 
 namespace Barryvdh\LaravelIdeHelper\Console;
 
+use Barryvdh\LaravelIdeHelper\Contracts\ModelHookInterface;
 use Barryvdh\Reflection\DocBlock;
 use Barryvdh\Reflection\DocBlock\Context;
 use Barryvdh\Reflection\DocBlock\Serializer as DocBlockSerializer;
@@ -279,6 +280,9 @@ class ModelsCommand extends Command
                     $this->getSoftDeleteMethods($model);
                     $this->getCollectionMethods($model);
                     $this->getFactoryMethods($model);
+
+                    $this->runModelHooks($model);
+
                     $output                .= $this->createPhpDocs($name);
                     $ignore[]              = $name;
                     $this->nullableColumns = [];
@@ -336,7 +340,7 @@ class ModelsCommand extends Command
      *
      * @param \Illuminate\Database\Eloquent\Model $model
      */
-    protected function castPropertiesType($model)
+    public function castPropertiesType($model)
     {
         $casts = $model->getCasts();
         foreach ($casts as $name => $type) {
@@ -412,7 +416,7 @@ class ModelsCommand extends Command
      *
      * @param \Illuminate\Database\Eloquent\Model $model
      */
-    protected function getPropertiesFromTable($model)
+    public function getPropertiesFromTable($model)
     {
         $table = $model->getConnection()->getTablePrefix() . $model->getTable();
         $schema = $model->getConnection()->getDoctrineSchemaManager();
@@ -509,7 +513,7 @@ class ModelsCommand extends Command
     /**
      * @param \Illuminate\Database\Eloquent\Model $model
      */
-    protected function getPropertiesFromMethods($model)
+    public function getPropertiesFromMethods($model)
     {
         $methods = get_class_methods($model);
         if ($methods) {
@@ -721,7 +725,7 @@ class ModelsCommand extends Command
      * @param string|null $comment
      * @param bool        $nullable
      */
-    protected function setProperty($name, $type = null, $read = null, $write = null, $comment = '', $nullable = false)
+    public function setProperty($name, $type = null, $read = null, $write = null, $comment = '', $nullable = false)
     {
         if (!isset($this->properties[$name])) {
             $this->properties[$name] = [];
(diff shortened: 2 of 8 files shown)


src/Console/ModelsCommand.php
```diff
@@ -976,4 +980,18 @@ protected function checkForCustomLaravelCasts(string $type): ?string
...
+    /**
+     * @param \Illuminate\Database\Eloquent\Model $model
+     */
+    protected function runModelHooks($model): void
+    {
+        $hooks = $this->laravel['config']->get('ide-helper.model_hooks', array());
+
+        foreach ($hooks as $hook) {
+            /** @var ModelHookInterface $hookInstance */
+            $hookInstance = new $hook();
```
mfn, line 993: IMHO this should resolve the class from the container so DI can be properly used.

config/ide-helper.php
```diff
@@ -111,6 +111,21 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | Define which hook classes you want to run for models to add custom information
+    |
+    | Hooks should implement Contracts\ModelHookInterface.
+    |
+    */
+
+    'model_hooks' => array(
+
```
mfn, line 126: Maybe an commented pseudo-example to be clear this expects something like `// App\Services\IdeHelper\MyHook::class` or so?

src/Console/ModelsCommand.php
Wimski: Great point. I'll definitely add that. I was also thinking of checking if the created instance is an instance of the `ModelHookInterface` to guard against missing method.

config/ide-helper.php
Wimski: I should probably also update the readme to include a bit about using model hooks.

mfn: Now that you mention it, good idea 😄

mr-feek:
**Context:**

While working on https://github.com/psalm/psalm-plugin-laravel, I've desired the ability to "change the template that gets rendered" (akin to https://github.com/barryvdh/laravel-ide-helper/issues/942). I understand that would require a lot of refactoring, so I haven't gotten around to do it yet.

------

**Feedback:**

The verbiage `run` seems to me it may be a bit confusing, as the hook can only **add** information to be written, right? Shall we consider making this a bit more clear on the interface level? Thoughts?

mfn:
@Wimski could this PR solve https://github.com/barryvdh/laravel-ide-helper/pull/930 ?

Wimski:
> @Wimski could this PR solve #930 ?

@mfn this PR doesn't add functionality to the IDE helper. It only opens up some of the existing functionality so you can add information from sources the IDE helper doesn't know about on its own. So this could solve #930 if there is already functionality in the IDE helper to write/define custom tags (which I didn't look into). If not, then this PR won't help with that.

mfn:
I thought about this again…

https://github.com/barryvdh/laravel-ide-helper/pull/945#issuecomment-637607758
> I personally feel this class could be reworked a lot, and therefore would be hesitant to exposing it publicly.

It matches my sentiment from https://github.com/barryvdh/laravel-ide-helper/pull/945#pullrequestreview-421267672
> I feel angsty opening up ModelsCommand suddenly being so public because it's so much spaghetti in there

But currently it's a catch-22.

No one came forward with a clear plan to clean this up and I can imagine such a refactoring could be quite intrusive.

For example what I described in https://github.com/barryvdh/laravel-ide-helper/issues/756 is kind of critical to my current use of ide-helper (I do use ide-helper within a real Laravel project but also in one were individual components are used) and I had to jump through quite hacky hoops to get it working. Was like 80% duck typing and 20% relying on extending code.

> No one came forward with a clear plan

That's kinda a point also, because I can surely come forward with a nice PR I feel is the "clean code of 2020" but what problem would it solve? Enabling things not possible before should be part of that vision.


----

Although the PR is quite unobtrusive technically, it could open can of worms in the current state hard to close, once the hooks "catch on". I'd err on the side of @mr-feek and question if it's the right to go ahead of this.

----
Apologies for all the meta babbling…

GuidoHendriks:
What's the status on this? Would really appreciate this change.

mfn:
The technical implementation here I think is nice and unobtrusive, except for the fact it opens up everything `public`ly. For a long time I was torn on this one and was leaning towards accepting it.

But the more I think about the implications, the more I'm worried.

I think the high level idea is great (providing a way of extending / intercepting ide-helpers functionality).

But as I already mentioned, there are so many smaller issues with the code base and then guaranteeing them all as part of the model hooks to work that way what previously was rather "internal details" worries me the most.

Alone the fact we've passing a full Symfony _Command_ into the hook and thus making it the primary point of contract IMHO feels absolutely wrong:
```php
interface ModelHookInterface
{
    public function run(ModelsCommand $command, Model $model): void;
}
```

----
I'm aware not every project on earth must fulfill the "clean architecture 2020 guidelines", let's be realistic. But open things up that way right now can cause a lot of issues for users of hooks when we want to improve / refactor the code base. That's why accepting the whole `ModelsCommand` is so dangerous.

I think more water should flow down the river first before agreeing on an implementation for such a topic.

----
@barryvdh how do you feel about this?

mfn:
@barryvdh there was a similar PR opened recently which achieves the same as this one -> https://github.com/barryvdh/laravel-ide-helper/pull/1157

I gave technical arguments in the other PR why I prefer this would but seems we lost track of it, my last comment went unanswered :)

My TL;DR: If you are fine with opening the ide-helper for this, I'm fine too (and ppl really want to have this feature…)

wimski:
@mfn I've rebased and fixed my branch and added an entry in the change log, so it's ready to go if you are 😉

mfn (approved):
Let's do it, 👍 from my side

Rolling over the ⚽ to @barryvdh 😏

daniel-de-wit:
Would love to use this 👍

mfn:
@barryvdh would you be fine with this hook?

Miljoen:
This MR looks great to me :)

wimski:
@barryvdh 👋

caugner:
FYI @barryvdh This was a breaking change for [`psalm/plugin-laravel`](https://github.com/psalm/psalm-plugin-laravel), which has a [`FakeModelsCommand` that overwrites this method](https://github.com/psalm/psalm-plugin-laravel/blob/eed0e6f3b90a49f763191be20659622aab23bf34/src/FakeModelsCommand.php#L50), see: https://github.com/psalm/psalm-plugin-laravel/issues/137

wimski:
@barryvdh I guess it should've been released under `10.0.0`. I also didn't realize this would've been a breaking change, which it obviously is by changing the visibility of methods 😢 

Maybe remove the following from the `CHANGELOG.md`:

```markdown
[Next release](https://github.com/barryvdh/laravel-ide-helper/compare/vX.X.X...master)
```

You won't know the version number until you're ready to do a release, because it depends on the changes since the last release.

mfn:
@caugner thanks for bringing this to attention!

> Maybe remove the following from the CHANGELOG.md:

This line is fine IMHO, it always shows the diff from the last release to current master (it was ~broken~ wrong recently due to a mistake 🤷‍♀️

Breaking changes: unfortunate, I guess. No one thought about that. I mean, hardly anyone would assume the use  ide-helper as an actual library though it's of course a valid use case. In this case it's slightly different. Or actually not.

We probably shouldn't roll back now, likely not worth it? And going forward, being more concise about this. Personally I don't have a problem pumping out more major versions in case we see a break somewhere, I'm not attached to version numbers or how they look.

mfn:
Next breaking change incoming, see https://github.com/barryvdh/laravel-ide-helper/pull/1198/files#r606697817

barryvdh:
I think it's a bit of an edge case and indeed not the intended/expected usage. But we could agree to bump minor versions for this, so plugins can specify minor versions.

caugner:
I'm not sure how other comparable projects handle this, but one could argue that `protected` methods on non-`final` classes are part of the interface, even though it certainly is an edge case.

Since specifying minor versions is rather unusual (afaik, as only major versions are expected to contain breaking changes) 
it would probably be best to bump the major version. `psalm/plugin-laravel` currently targets `^2.8.0`, so a minor version bump (although semantically more correct; it's a feature, not a bugfix) would not make any difference for that case.

What would be reasons *against* a major version bump?

barryvdh:
That all end-users need to upgrade their dependencies.

mr-feek:
Making a public class method visibility change in a minor version will still have the potential to leave lots of users in a broken state. We released versions of `psalm-plugin-laravel` saying that we work with `"barryvdh/laravel-ide-helper": "^2.8.0"`. So if anyone updates `ide-helper` without updating `psalm-plugin-laravel`, `psalm-plugin-laravel` will be broken.

The only _real, backwards compatible solution_  here is to make a hotfix release reverting the changes, and then re-releasing the breaking changes in a major version.

IMO "model hooks" brings a ton of new functionality and increases the public interface of this package which in itself could warrant a major version upgrade.

-------

That being said, I understand 100% that it was not _expected_ for this class to be used outside the codebase. But we did because it brought great functionality to statically analyzing laravel projects and it wasn't marked as final 😄 

My push is for a major version change, but I understand the rationale of being a bit reluctant to do so.

PR #947, opened 2020-06-03
Added the support for _int4 (integer[]) type to Postgresql

joaovargasantunes:
"Exception: Unknown database type _int4 requested, Doctrine\DBAL\Platforms\PostgreSQL92Platform may not support it."

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -361,6 +361,7 @@ class ModelsCommand extends Command
         $schema = $model->getConnection()->getDoctrineSchemaManager($table);
         $databasePlatform = $schema->getDatabasePlatform();
         $databasePlatform->registerDoctrineTypeMapping('enum', 'string');
+        $databasePlatform->registerDoctrineTypeMapping('_int4', 'string');
 
         $platformName = $databasePlatform->getName();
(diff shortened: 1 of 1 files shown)


src/Console/ModelsCommand.php
```diff
@@ -361,6 +361,7 @@ protected function getPropertiesFromTable($model)
         $schema = $model->getConnection()->getDoctrineSchemaManager($table);
         $databasePlatform = $schema->getDatabasePlatform();
         $databasePlatform->registerDoctrineTypeMapping('enum', 'string');
+        $databasePlatform->registerDoctrineTypeMapping('_int4', 'string');
```
mfn, line 364: But isn't `_int4` rather an "array of int4", thus of type `array`?

Please try the following first if it fixes your problem: in your ide-helper config, add this:
```php
    'custom_db_types' => [
        'postgresql' => [
            '_int4' => 'array', //Map postgres "array of ints" type to array
        ],
    ],

```

mfn:
@barryvdh I vote for dropping the PR; not feedback from author and I'm using in fact `_int4` and can vouch that `string` is not correct, but `array` is.

src/Console/ModelsCommand.php
victordcp: I had this same issue, and you solution worked perfectly.
Thanks a lot.

PR #954, opened 2020-06-16
[Model] Simplify full namespaces for already included resources

### LINKED ISSUE #565: Simplify full namespaces for already included resources
Hello,

by using phpstorm, there is an option `Simplify FQN`. Is it possible, that `ide-helper:models` command would do that?

I'm not talking about whole file modification, that automatically write `use Carbon\Carbon;` and then in php docblock would be only ` * @property Carbon|null $created_at`,
*but at least if class already has such includes.*

For example:
```
namespace Module\Sample;

/**
 * Class SampleClass
 *
 * @package Module\Sample
 * @property int $id
 * @property \Carbon\Carbon|null $created_at
 * @property \Carbon\Carbon|null $updated_at
 * @method static \Illuminate\Database\Eloquent\Builder|\Module\Sample\SampleClass whereCreatedAt($value)
 * @method static \Illuminate\Database\Eloquent\Builder|\Module\Sample\SampleClass whereDeletedAt($value)
 * @method static \Illuminate\Database\Eloquent\Builder|\Module\Sample\SampleClass whereId($value)
 * @mixin \Eloquent
 */
class SampleClass extends Model
```
Should be (if Carbon included and SampleClass itself, but Builder not)

```
namespace Module\Sample;

/**
 * Class SampleClass
 *
 * @package Module\Sample
 * @property int $id
 * @property Carbon|null $created_at
 * @property Carbon|null $updated_at
 * @method static \Illuminate\Database\Eloquent\Builder|SampleClass whereCreatedAt($value)
 * @method static \Illuminate\Database\Eloquent\Builder|SampleClass whereDeletedAt($value)
 * @method static \Illuminate\Database\Eloquent\Builder|SampleClass whereId($value)
 * @mixin \Eloquent
 */
class SampleClass extends Model
```

LANGERGabriel:
Fixes #565 .
Also fixes a bug where cast type was not properly added when the return type was found via phpdoc.


mfn:
> Simplify full namespaces for already included resources

I didn't realize this until I tried it out: every FQCN which at some point is manually imported, will then not be written as FQCN.

Say if you import `\Illuminate\Database\Eloquent\Builder` => it will then be written as just `Builder`. Aliases seem to be also correctly supported.

MAJOR thing, thank you 🙏

mfn:
I noticed it doesn't cover the case where models are relations, they are still kept in the FQN.
```php
namespace App\Models;

class ModelOne extends …
```
```php
namespace App\Models;

/**
 *
 * @property-read \App\ModelOne|null $ticket
 */
class ModelTwo extends …
{
    public function modelOne(): HasOne
    {
        return $this->hasOne(ModelOne::class);
    }
}
```
When I import `ModelOne`, ide-helper still writes it back as `\App\ModelOne`.

PR #959, opened 2020-06-22
composer normalize

Gummibeer:
This PR normalizes the `composer.json` by running https://github.com/ergebnis/composer-normalize and adds a GitHub action that keeps the `composer.json` normalized.

### DIFF
--- a/.github/workflows/composer-normalize.yml
+++ b/.github/workflows/composer-normalize.yml
@@ -1,0 +1,25 @@
+name: normalize composer.json
+
+on:
+  push:
+    paths:
+      - composer.json
+
+jobs:
+  normalize:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v2
+
+      - name: Validate Composer configuration
+        run: composer validate --strict
+
+      - name: Normalize composer.json
+        run: |
+          composer global require ergebnis/composer-normalize
+          composer normalize
+
+      - uses: stefanzweifel/git-auto-commit-action@v4.0.0
+        with:
+          commit_message: normalize composer.json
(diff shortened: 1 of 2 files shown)


.github/workflows/composer-normalize.yml
```diff
@@ -0,0 +1,28 @@
...
+
+      - name: Validate Composer configuration
+        run: composer validate --strict
+
+      - name: Normalize composer.json
+        run: |
+          composer global require ergebnis/composer-normalize
+          composer normalize
+
+      - uses: stefanzweifel/git-auto-commit-action@v4.0.0
```
mfn, line 26: Will this work on PRs made from forks?

I read https://github.com/stefanzweifel/git-auto-commit-action/issues/25 but am unsure what the outcome here would be.

Gummibeer: You are right, it will depend on the permission settings in the fork.
I will remove the `pull_request` trigger and run it only on master branch on push!?

Gummibeer: https://github.com/barryvdh/laravel-ide-helper/pull/959/commits/0af2ec3af3131a234b071885da1eb6e982ed630e

barryvdh:
Doesn't this seem like a bit of overkill just to format the composer.json? I mean, how often does this change?

Gummibeer:
@barryvdh you are right that it doesn't happen this often.
But my experience is that exactly these uncommon changes end in a not normalized `composer.json`. The GH Action is also free it takes `~30sec`. So it does not block an action container for long, it prevents maintainers from caring about the structure/sort of `composer.json` and it also validates the `composer.json`.

mfn:
Again also here (as in: in this repo, have seen others) we've a problem with the Travis integration not properly reporting back although all tests are green:
image

@barryvdh off-topic drive-by question: are you open to switch remove travis in favor of github actions?

barryvdh:
Yeah I converted a few already. Let's check that with https://github.com/barryvdh/laravel-ide-helper/pull/960

PR #960, opened 2020-06-22
Use Github Actions

barryvdh:


.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,72 @@
...
+    - name: Setup PHP
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: ${{ matrix.php }}
+        coverage: none
+
+    - name: Install dependencies
+      run: |
+        composer remove phpro/grumphp --no-interaction --no-update --dev
+        composer require "laravel/framework:${{ matrix.laravel }}" "orchestra/testbench:${{ matrix.testbench }}" "orchestra/testbench-dusk:${{ matrix.testbench }}" --no-interaction --no-update
```
mr-feek, line 65: fwiw, I don't think `orchestra/testbench-dusk` is necessary?

barryvdh: ah yes I copied it from laravel-cors which does use it. Thanks

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,72 @@
...
+        composer update --${{ matrix.dependency-version }} --prefer-dist --no-interaction --no-suggest
+
+    - name: Execute Unit Tests
+      run: vendor/bin/phpunit
+
+    - name: Check PSR-12 Codestyle
+      run: vendor/bin/phpcs --standard=psr2 src/
```
mfn, line 72: It feels wrong to me to run this for every job in the matrix => I vote for making this a separate non-matrix job

mfn: Also, this should re-use the composer scripts, i.e. simply `composer check-style`

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,72 @@
...
+        coverage: none
+
+    - name: Install dependencies
+      run: |
+        composer remove phpro/grumphp --no-interaction --no-update --dev
+        composer require "laravel/framework:${{ matrix.laravel }}" "orchestra/testbench:${{ matrix.testbench }}" --no-interaction --no-update
+        composer update --${{ matrix.dependency-version }} --prefer-dist --no-interaction --no-suggest
+
+    - name: Execute Unit Tests
+      run: vendor/bin/phpunit
```
mfn, line 69: Should also use composer scripts, i.e. `composer test`

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,72 @@
...
+          - laravel: 5.6.*
+            php: 7.4
+          - laravel: 5.5.*
+            php: 7.4
+
+    name: P${{ matrix.php }} - L${{ matrix.laravel }} - ${{ matrix.dependency-version }} - ${{ matrix.os }}
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v1
```
mfn, line 54:
```suggestion
      uses: actions/checkout@v2
```

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,72 @@
...
+            dependency-version: prefer-stable
+        exclude:
+          - laravel: 5.7.*
+            php: 7.4
+          - laravel: 5.6.*
+            php: 7.4
+          - laravel: 5.5.*
+            php: 7.4
+
+    name: P${{ matrix.php }} - L${{ matrix.laravel }} - ${{ matrix.dependency-version }} - ${{ matrix.os }}
```
mfn, line 50: There's no variation in the matrix.os?
```suggestion
    name: P${{ matrix.php }} - L${{ matrix.laravel }} - ${{ matrix.dependency-version }}
```

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,72 @@
...
+  - cron: '0 0 * * *'
+
+jobs:
+  php-tests:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        php: [7.4, 7.3, 7.2]
+        laravel: [7.*]
```
mr-feek, line 14: the composer.json file claims support for `^5.5 || ^6 || ^7"`, do we want to expand this matrix here too? It looks like it might be an oversight here that we are missing them, as they are in the `includes` section below 😄

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,72 @@
+on:
+  push:
+  pull_request:
+  schedule:
+  - cron: '0 0 * * *'
+
+jobs:
+  php-tests:
+    runs-on: ubuntu-latest
```
mfn, line 9: Suggest to add this somewhere:
```yaml
    env:
      COMPOSER_NO_INTERACTION: 1
```
and remove all the `--no-interaction`

.github/workflows/run-tests.yml
mfn: Maybe it's smarter to actually define all things in the matrix and rather use exclude, would that work?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,72 @@
...
+
+    - name: Setup PHP
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: ${{ matrix.php }}
+        coverage: none
+
+    - name: Install dependencies
+      run: |
+        composer remove phpro/grumphp --no-interaction --no-update --dev
```
mfn, line 64: As per `env:` suggestion aove, `--no-interaction` can be removed

.github/workflows/run-tests.yml
barryvdh: I think it's a bit overkill to test all those combinations, that's why I left them out the matrix, but included those 2 in the includes.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,72 @@
...
+        composer require "laravel/framework:${{ matrix.laravel }}" "orchestra/testbench:${{ matrix.testbench }}" --no-interaction --no-update
+        composer update --${{ matrix.dependency-version }} --prefer-dist --no-interaction --no-suggest
```
mfn, line 66: Would also add `--no-progress`

.github/workflows/run-tests.yml
mr-feek: Figured that might be the case, need to draw a line somewhere 👍

barryvdh:
Not really sure if I can just run pull_request OR push, not both. But at least it works.

.github/workflows/run-tests.yml
barryvdh: What would be the benefit?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,72 @@
...
+
+    strategy:
+      matrix:
+        php: [7.4, 7.3, 7.2]
+        laravel: [7.*]
+        testbench: [5.*]
+        dependency-version: [prefer-lowest, prefer-stable]
+        include:
+          - laravel: 7.*
+            testbench: 5.*
```
mfn, line 19: I'm wonder why this is necessary?

Got them working quite nicely without having to specify this , see https://github.com/mfn/laravel-ide-helper/actions/runs/143792113 ( I was already working on PR but now stopped in favor of yours, joining forces 💪 )

.github/workflows/run-tests.yml
barryvdh: Done

.github/workflows/run-tests.yml
barryvdh: Does that matter?

.github/workflows/run-tests.yml
barryvdh: Yeah we could do that, this is more in line what Travis did and it's probably faster (because it re-uses the composer stuff), but we could change that I guess.

.github/workflows/run-tests.yml
mfn: No need to repeat `--no-interaction` everywhere, "shorted" custom commands and none of the composer invocations want interaction so the env var applies to all. It was added to composer for the CI use-case

.github/workflows/run-tests.yml
mfn: Makes sure the composer scripts are not just looking nice but also knows "to work", i.e. a single place specifies how things are supposed to be executed.

So IMHO, yup, there's sense in having this.

barryvdh:
@mfn Merging this now, can you pull request your suggestions?

.github/workflows/run-tests.yml
barryvdh: Hmm I also though it was ugly, but remembered having problems with it before. But PR away :)

barryvdh:
Probably also a good time to switch to PSR-12 instead of PSR-2.

.github/workflows/run-tests.yml
barryvdh: ok

mfn:
`\o/`

PR #961, opened 2020-06-22
Small improvements for github actions

mfn:
I.e. some of the not-yet-implemented feedback from https://github.com/barryvdh/laravel-ide-helper/pull/960

- clean up mention of travis
- prefer composer scripts for final execution
- use env var for no-interactive
- adjust .gitignore exports
- separate non-matrix job for phpcs

Oh, and gave the workflow file a name so it appears more nicely in e.g. https://github.com/barryvdh/laravel-ide-helper/actions

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,6 @@
 
 [![Latest Version on Packagist][ico-version]][link-packagist]
 [![Software License][ico-license]](LICENSE.md)
-[![Build Status][ico-travis]][link-travis]
 [![Total Downloads][ico-downloads]][link-downloads]
 
 **Complete PHPDocs, directly from the source**
@@ -299,8 +298,6 @@ The Laravel IDE Helper Generator is open-sourced software licensed under the [MI
 
 [ico-version]: https://img.shields.io/packagist/v/barryvdh/laravel-ide-helper.svg?style=flat-square
 [ico-license]: https://img.shields.io/badge/license-MIT-brightgreen.svg?style=flat-square
-[ico-travis]: https://img.shields.io/travis/barryvdh/laravel-ide-helper/master.svg?style=flat-square
 [ico-downloads]: https://img.shields.io/packagist/dt/barryvdh/laravel-ide-helper.svg?style=flat-square
 [link-packagist]: https://packagist.org/packages/barryvdh/laravel-ide-helper
-[link-travis]: https://travis-ci.org/barryvdh/laravel-ide-helper
 [link-downloads]: https://packagist.org/packages/barryvdh/laravel-ide-helper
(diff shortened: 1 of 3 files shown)


mfn:
Unrelated to this PR, but some of the installations take awfully / unnecessary long (i.e. up to 1m something)

Two solutions:
- use composer dependency caching => https://github.com/shivammathur/setup-php#cache-composer-dependencies
- switch to composer v2 (setup-php supports this via `tools: composer:v2`)
  However this currently conflicts with grumphp which is declared a composer plugin but only supports v1
  [I don't like grumphp 😄 ](https://github.com/barryvdh/laravel-ide-helper/pull/866)

v2 would be preferred but…

mfn:
@barryvdh all green √

barryvdh:
We don't really need grumphp for the tests, so we could remove them before installing the rest (composer remove ... --no-update )
But I do like Composer v1 for now, because that's still the stable.

barryvdh:
Would it make sense to have 2 actions instead? Or what is the difference?

README.md
```diff
@@ -2,7 +2,6 @@
 
 [![Latest Version on Packagist][ico-version]][link-packagist]
 [![Software License][ico-license]](LICENSE.md)
-[![Build Status][ico-travis]][link-travis]
```
barryvdh, line 5: Can we add the Actions badges?

mfn:
> We don't really need grumphp for the tests, so we could remove them before installing the rest (composer remove ... --no-update )

Done!

mfn:
> Would it make sense to have 2 actions instead? Or what is the difference?

It's a matter / taste of organizing things.

One example, for the badge I added via https://github.com/barryvdh/laravel-ide-helper/pull/966 it's now "easy" to have one badge which covers in fact phpunit and phpcs. Otherwise with two workflows, we "would" need two badges if it would be relevant.

mfn:
>  But I do like Composer v1 for now, because that's still the stable.

Good call, because we currently trigger a bug there and can't use it anyway => reported via https://github.com/composer/composer/issues/8903#issuecomment-647930096

barryvdh:
Or sorry missed this. I reported the same bug :P

mfn:
What I wrote in https://github.com/barryvdh/laravel-ide-helper/pull/961#issuecomment-647617608
> use composer dependency caching

I tried this, actually it doesn't _really_ improve things. It seems the composer v1 _resolver_ is so slow and downloading things is that least of concerns here 💥 
- uncached: https://github.com/mfn/laravel-ide-helper/runs/799782806?check_suite_focus=true
- cached: https://github.com/mfn/laravel-ide-helper/runs/799799103?check_suite_focus=true

So currently nothing actionable until composer v2 will be usable for us here.

barryvdh:
Yeah I tried it as well, bit strange.

PR #962, opened 2020-06-22
[Chore] Introduce psalm

mr-feek:
👋 Any interest in introducing a static analysis tool like psalm? I currently set it to the least strict setting, but can make more prs tightening it up if that is of any interest

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<files psalm-version="3.12.0@d46283075d76ed244f7825b378eeb1cee246af73">
+  <file src="src/Generator.php">
+    <UndefinedClass occurrences="8">
+      <code>'\Laravel\Lumen\Application'</code>
+      <code>\Auth</code>
+      <code>\DB</code>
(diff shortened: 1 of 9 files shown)


psalm-baseline.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<files psalm-version="3.12.0@d46283075d76ed244f7825b378eeb1cee246af73">
+  <file src="src/Generator.php">
+    <UndefinedClass occurrences="8">
```
mr-feek, line 4: These are a bit annoying, I think psalm itself can work around these though. I've filed an issue: https://github.com/vimeo/psalm/issues/3629

src/Generator.php
```diff
@@ -283,36 +283,6 @@ function ($alias) {
         );
     }
 
-    /**
-     * Get the driver/connection/store from the managers
-     *
-     * @param $alias
-     * @return array|bool|string
-     */
-    public function getDriver($alias)
```
mr-feek, line 292: this seems to be dead code

mfn:
Very cool!

PR #963, opened 2020-06-23
[Chore] Composer 2 support

mr-feek:
👋 I think this should allow usage with composer 2. There isn't an official tag yet, so `^2.0@dev` should suffice to allow dev stability versions of the release. Naturally once the official composer 2 release is made, we should update the dependency to be `"composer/composer": "^1.6 || ^2.0",`

------

I have verified that our (ide-helper's) usage of composer is limited to `ClassMapGenerator::createMap`, whose signature doesn't seem to have changed in composer 2.

------

`grumphp` added support for composer 2 in [`0.18.1`](https://github.com/phpro/grumphp/pull/768), hence why I have updated it in this PR.

Related: https://github.com/psalm/psalm-plugin-laravel/issues/89

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "illuminate/view": "^5.5 || ^6 || ^7",
         "mockery/mockery": "^1.3",
         "orchestra/testbench": "^3 || ^4 || ^5",
-        "phpro/grumphp": "^0.17.1",
+        "phpro/grumphp": "^0.19.0",
         "squizlabs/php_codesniffer": "^3.5",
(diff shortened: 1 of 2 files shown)


mr-feek:
This doesn't seem like it can be fully tested because of the `--prefer-stable` flags in the github workflow build matrix. 

When I do a composer update without the ``--prefer-stable`` flag, `composer/semver (3.0.0)` is installed.
With the prefer stable flag, `composer/semver (1.5.1)` is installed

PR #965, opened 2020-06-23
chore: use composer scripts in grumphp

mr-feek:
This reduces the # of places we need to change configurations for various dev tools. Now grumphp will just execute the composer script listed under the `script` key.

I updated grumphp because this later version has support for running various different composer scripts


barryvdh:
This conflicts with your other PR

mfn:
I've had again problems with grumphp github hooks. I wish the installation of a package would not automagically mess with my local hooks :/

(problem was: installed packages, then tested something and removed vendor/, moved on, came back, made a non-code change, couldn't even commit, rage quit :rage1: 😄 )

PR #968, opened 2020-06-23
GHA Composer v2

barryvdh:


barryvdh:
Doesn't seem Composer 2 is stable enough yet. Logged an issue here: https://github.com/composer/composer/issues/8941#issuecomment-647970871

.github/workflows/run-tests.yml
```diff
@@ -59,12 +59,14 @@ jobs:
       with:
         php-version: ${{ matrix.php }}
         coverage: none
+        tools: composer:v2
         
     - name: Install dependencies
       run: |
+        composer self-update --2
```
mfn, line 66: This should be redundant as setup-php does this already.

barryvdh: Yeah I added it because of the weird error I got, wanted to make sure it was the very latest version.

barryvdh:
It works 🎉

mfn:
No it doesn't :)

They didn't reopen the bug yet, but see https://github.com/composer/composer/issues/8903#issuecomment-651225567 and further comments.

It also requires a specific PHP runtime version 🙀 (7.2) and `composer.json` combination to crash

mfn (approved):
@barryvdh I took the liberty and rebased it (and also removed the `self-update` part, not necessary), and now it works => good to merge!

(I guess because we got rid of 5.5 )

PR #969, opened 2020-06-23
Make writing relation count properties optional

AegirLeet:
Accidentally fucked up https://github.com/barryvdh/laravel-ide-helper/pull/837 by deleting the branch or force-pushing (can't reopen), so here's a new PR.

I've rebased on top of current master, added a test for `write_model_relation_count_properties` and also added another test for `write_model_magic_where` while I was at it.

CC @mfn

### DIFF
--- a/tests/Console/ModelsCommand/MagicWhere/Test.php
+++ b/tests/Console/ModelsCommand/MagicWhere/Test.php
@@ -1,0 +1,150 @@
+<?php declare(strict_types=1);
+
+namespace Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\MagicWhere;
+
+use Barryvdh\LaravelIdeHelper\Console\ModelsCommand;
+use Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\AbstractModelsCommand;
+use Illuminate\Filesystem\Filesystem;
+use Mockery;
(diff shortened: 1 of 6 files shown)


tests/Console/ModelsCommand/MagicWhere/Test.php
```diff
@@ -0,0 +1,150 @@
...
+ * @property \Illuminate\Support\Carbon|null $created_at
+ * @property \Illuminate\Support\Carbon|null $updated_at
+ * @method static \Illuminate\Database\Eloquent\Builder|Post newModelQuery()
+ * @method static \Illuminate\Database\Eloquent\Builder|Post newQuery()
+ * @method static \Illuminate\Database\Eloquent\Builder|Post query()
+ * @mixin \Eloquent
+ */
+class Post extends Model
+{
+}
```
mfn, line 144: Shouldn't this include a relation so we see the `_count` property?

AegirLeet:
The *ModelsCommand/MagicWhere* part isn't related to _count properties, I just noticed it was missing a test and decided to add it as well.

_count being added (default behavior) is already tested in some other tests, like in *ModelsCommand/Relations*. So *ModelsCommand/RelationCountProperties* only tests that they disappear when the new option is set to `true`.

PR #971, opened 2020-06-24
Drop support for Laravel 5.5

mfn:
For consideration.

See https://github.com/barryvdh/laravel-ide-helper/pull/958#issuecomment-648567330


README.md
```diff
@@ -31,7 +31,7 @@ Require this package with composer using the following command:
 composer require --dev barryvdh/laravel-ide-helper
 ```
 
-If you are using Laravel versions older than 5.5, add the service provider to the `providers` array in `config/app.php`:
+If you want to manually load the service provider, add the following class to the `providers` array in `config/app.php`:
```
mfn, line 34: Decide to keep this, can be useful for those not wanting to use the auto discover feature

mr-feek:
I for one would be saddened by this change, but I am very biased as I have many enterprise apps on 5.5 for various reasons. 

Could we consider dropping the composer.json version constraints from this diff as the package does currently work fine with 5.5? It seems to me like this is a workaround for another composer issue. At the end of the day, I do understand that there needs to be a line drawn for supported versions. 

I understand I’m most likely in the minority. I wonder if we can find some sort of stats around what versions of Laravel are currently using this package to help inform.

barryvdh:
I'm also not a big fan of dropping support just because tests don't run. But 5.5 is security releases only, and that ends in 2 months. So I would assume most LTS users will be upgrading to 6.0 soon.

Keeping 5.5 installable and not test it, could be an option, but we can't guarantee it works then.

barryvdh:
I'm gonna hold off on this for now, as the tests are fixed.

mfn:
@barryvdh we might have already dropped L5.5 support in certain cases 😬

In https://github.com/barryvdh/laravel-ide-helper/pull/868 the minimum PHP version was bumped to 7.2, but Laravel 5.5 is `>= 7.0` => https://github.com/laravel/framework/blob/a36b3a0206f772eda47dea47b6683ecb2ce7db00/composer.json#L18

So if PHP 7.0/7.1 are used with L5.5, ide-helper might not be able to be installed (or work: as we don't run CI against older PHP versions).

Nothing immediately actionable, just wanted to point it out 🤷‍♀️

barryvdh:
Yeah but PHP 7.1 is also not supported. 7.2 + L5.5 is.

PR #980, opened 2020-06-27
Improve readme in certain areas

mfn:
Whilst working through hundreds of issues I made note of some recurring patterns (like [custom_db_types](https://github.com/barryvdh/laravel-ide-helper/issues?q=is%3Aissue+custom_db_types+)) to follow-up to improve the README where some of the available configuration options are not mentioned.

Overview of the changes:
- remove mention of sublime command (which itself has been removed in https://github.com/barryvdh/laravel-ide-helper/pull/976 )
- update phpstorm link to regading `.phpstorm.meta.php` (old link only redirects, use the destination link directly)
- mention that the default _ide_helper.php name can be changed
- document `meta_filename` and hint it can be written into a special subdirectory
- remove hint about \Eloquent or `@mixin`, as the latter is automatically added nowadays
- update how the current phpdoc looks like
- document `write_model_magic_where`
- document `write_model_relation_count_properties`
- document `custom_db_types`
- clarify installation regarding auto-discovery vs. manual
- mention to avoid caching the config in dev

I figured a PR for each change would be too much but still kept focus single commit changes here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,8 @@ If for some reason you want manually control this:
   }
   ```
 
+> Note: Avoid caching the configuration in your development environment, it may cause issues after installing this package; respectively clear the cache beforehand via `php artisan cache:clear` if you encounter problems when running the commands
+
 ## Usage
 
 _Check out [this Laracasts video](https://laracasts.com/series/how-to-be-awesome-in-phpstorm/episodes/15) for a quick introduction/explanation!_
--- a/README.md
+++ b/README.md
@@ -31,26 +31,33 @@ Require this package with composer using the following command:
 composer require --dev barryvdh/laravel-ide-helper
 ```
 
-If you are using Laravel versions older than 5.5, add the service provider to the `providers` array in `config/app.php`:
+This package makes use of [Laravels package auto-discovery mechanism](https://medium.com/@taylorotwell/package-auto-discovery-in-laravel-5-5-ea9e3ab20518), which means if you don't install dev dependencies in production, it also won't be loaded.
 
-```php
-Barryvdh\LaravelIdeHelper\IdeHelperServiceProvider::class,
-```
-
-In Laravel, instead of adding the service provider in the `config/app.php` file,
-you can add the following code to your `app/Providers/AppServiceProvider.php` file, within the `register()` method:
-
-```php
-public function register()
-{
-    if ($this->app->environment() !== 'production') {
-        $this->app->register(\Barryvdh\LaravelIdeHelper\IdeHelperServiceProvider::class);
+If for some reason you want manually control this:
+- add the package to the `extra.laravel.dont-discover` key in `composer.json`, e.g.
+  ```json
+  "extra": {
+    "laravel": {
+      "dont-discover": [
+        "barryvdh/laravel-ide-helper",
+      ]
     }
-    // ...
-}
-```
-
-This will allow your application to load the Laravel IDE Helper on non-production environments.
+  }
+  ```
+- Add the following class to the `providers` array in `config/app.php`:
+  ```php
+  Barryvdh\LaravelIdeHelper\IdeHelperServiceProvider::class,
+  ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -164,14 +182,38 @@ Or can be ignored by setting the `ignored_models` config
 
 > Note: With namespaces, wrap your model name in double-quotes (`"`): `php artisan ide-helper:models "API\User"`, or escape the slashes (`Api\\User`).
 
-For properly recognition of `Model` methods (i.e. `paginate`, `findOrFail`) you should extend `\Eloquent` or add
```
mfn, line 167: This was deliberately removed:
- are we _really_ suggesting people should extend their models from `\Eloquent`? To the best of my knowledge, this is not necessary at all 🤔
- same with add `/** @mixin \Eloquent */`, this is nowadays always added?
  (note: the diff is hard to read, there two lines removed, the second one is much further down)

README.md
```diff
@@ -219,12 +261,17 @@ $app->make('events')->fire();
 
 // When the key is not found, it uses the argument as class name
 app('App\SomeClass');
+// Also works with
+app(App\SomeClass::class);
 ```
 
 > Note: You might need to restart PhpStorm and make sure `.phpstorm.meta.php` is indexed.
+>
```
mfn, line 269: Improves readability, otherwise it's rendered as as single sentence

README.md
```diff
@@ -201,7 +243,7 @@ For this to work, you must also publish the PhpStorm Meta file (see below).
 
 ## PhpStorm Meta for Container instances
 
-It's possible to generate a PhpStorm meta file to [add support for factory design pattern](https://confluence.jetbrains.com/display/PhpStorm/PhpStorm+Advanced+Metadata).
+It's possible to generate a PhpStorm meta file to [add support for factory design pattern](https://www.jetbrains.com/help/phpstorm/ide-advanced-metadata.html).
```
mfn, line 246: Updated because the former always redirects

README.md
```diff
@@ -104,6 +115,8 @@ The `Illuminate/Support/helpers.php` is already set up, but you can add/remove y
 If you don't want to write your properties yourself, you can use the command `php artisan ide-helper:models` to generate
 PHPDocs, based on table columns, relations and getters/setters.
 
+> Note: this command requires a working database connection to introspect the table of each model
```
mfn, line 118: There were some issues of people not realizing this is a requirement for the `:models` command

README.md
```diff
@@ -85,7 +96,7 @@ You can configure your `composer.json` to do this each time you update your depe
 },
 ```
 
-You can also publish the config file to change implementations (ie. interface to specific class) or set defaults for `--helpers` or `--sublime`.
+You can also publish the config file to change implementations (ie. interface to specific class) or set defaults for `--helpers`.
```
mfn, line 99: Was removed 2 days ago because it was deprecated / a no-op anyway

README.md
```diff
@@ -31,26 +31,35 @@ Require this package with composer using the following command:
...
+  public function register()
+  {
+      if ($this->app->environment() !== 'production') {
+          $this->app->register(\Barryvdh\LaravelIdeHelper\IdeHelperServiceProvider::class);
+      }
+      // ...
+  }
+  ```
+
+> Note: Avoid caching the configuration in your development environment, it may cause issues after installing this package; respectively clear the cache beforehand via `php artisan cache:clear` if you encounter problems when running the commands
```
mfn, line 62: This also sometimes tripped people

README.md
```diff
@@ -31,26 +31,35 @@ Require this package with composer using the following command:
 composer require --dev barryvdh/laravel-ide-helper
 ```
 
-If you are using Laravel versions older than 5.5, add the service provider to the `providers` array in `config/app.php`:
```
mfn, line 34: Since we only support 5.5+, it's not necessary to refer to older versions.

As auto-discovery thus is the standard, I rephrased the installation section based on this and added more details how to manually do this in case it's desired

Issue #666, opened 2018-05-24
Bug: `ide-helper:models --write` breaks @SuppressWarnings(PHPMD.*) notation


pablorsk:
When `php artisan ide-helper:models --write` is runned, notations like `@SuppressWarnings(PHPMD.ExcessiveClassComplexity)` was changed to `@SuppressWarnings (PHPMD.ExcessiveClassComplexity)` (check space before `(`).

### Original model
```php
/**
 * @SuppressWarnings(PHPMD.ExcessiveClassComplexity)
 */
class SomeModel extends Model
{
    public function childrens()
    {
        return $this->hasMany(Document::class);
    }
}
```
### After ide-helper:models
```php
/**
 * App\Models\SomeModel
 *
 * @SuppressWarnings (PHPMD.ExcessiveClassComplexity)
 * @property-read \App\Collection|\App\Models\Document[] $childrens
 * @mixin \Eloquent
 */
class SomeModel extends Model
{
    public function childrens()
    {
        return $this->hasMany(Document::class);
    }
}
```

### Diff
```diff
- * @SuppressWarnings(PHPMD.ExcessiveClassComplexity)
+ * @SuppressWarnings (PHPMD.ExcessiveClassComplexity)
```

mfn:
I can still reproduce this :-/ I _suspect_ the problem is within https://github.com/barryvdh/ReflectionDocBlock though I could be wrong

barryvdh:
I think we might nee a custom tag for that.

EspadaV8:
Is there a workaround for this while a fix is implemented?

pablorsk:
We have some problems like this, not just PHPMD. Then you can add and customize a script on your `composer.json`:

```
   "scripts": {
        "ide-helper-models":[
            "php artisan migrate",
            "php artisan ide-helper:models --write",
            "find app/* -type f -print0 | xargs -0 sed -i 's/@SuppressWarnings (/@SuppressWarnings(/g'",
            "find app/* -type f -print0 | xargs -0 sed -i 's/@property \\\\Illuminate\\\\Support\\\\Carbon/@property \\\\Carbon\\\\Carbon/g'",
            "find app/* -type f -print0 | xargs -0 sed -i '/* @property int.null $.*_count/d'",
            "find app/* -type f -print0 | xargs -0 sed -i '/* @property-read int.null $.*_count/d'",
[...]

PR #981, opened 2020-06-28
[TEST] Show that custom phpdoc tags without space afterwards are unexpectedly changed

mfn:
## Summary

This adds a test for the (current bug) behaviour of https://github.com/barryvdh/laravel-ide-helper/issues/666

It shows that
- `@SuppressWarnings(PHPMD.ExcessiveClassComplexity)`
  is changed to
- `@SuppressWarnings (PHPMD.ExcessiveClassComplexity)`

If this ever gets fixed, this fails and needs to be adapted,
ensuring it won't break again.


tests/Console/ModelsCommand/CustomPhpdocTags/__snapshots__/Test__testNoSpaceAfterCustomPhpdocTag__1.php
```diff
@@ -0,0 +1,22 @@
...
+namespace Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\CustomPhpdocTags\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
+/**
+ * The `@SuppressWarnings` tag below contains no space before the `(` but when
+ * the class phpdoc is written back, one is inserted.
+ *
+ * @link https://github.com/barryvdh/laravel-ide-helper/issues/666
+ * @SuppressWarnings (PHPMD.ExcessiveClassComplexity)
```
mfn, line 12: This exposes the buggy behaviour, there should be no space between `SuppressWarnings` and `(`

barryvdh:
Shouldn't we add a breaking test and perhaps skip it for now? Seems a bit weird to add a wrong test.

mfn:
I understand but the problem is, you've to "remember" (manually / in your head) that once this bug is fixed to remove the skip.

My take is: I don't think too much, when this get fixed this test basically says "hello, I already exist, I just need to be adjusted".

Also means that we _already_ have the test before the implementation (but it makes sense because we acknowledged it's a bug we would like to see fixed) and the test just needs to change the line some comments.

It's kind of a forward-preparation.

barryvdh:
Yeah but now it looks like custom behavior?

mfn:
Sorry, parse error :) Can you elaborate?

barryvdh:
Like expected behavior*
I'll think about this, how we can just fix it. (Perhaps dirty to just remove the space with a str contains check?

mfn:
> (Perhaps dirty to just remove the space with a str contains check?)

🙀

mfn:
Btw, the idea for this PR was simply that I don't have time to work on a fix right now but I figured we want a test _anyway_ to ensure it works so already adding now one once something finds time saves already having to create a test the next time :}

But if you take look at it now, that would be great: still you would want to use this test I assume, since it's already "there".

mfn (commented):
@barryvdh to elaborate again: this is a pattern to add a test for known issue so, once it is fixed, we adjust the test and already have a test/guarantee it won't break (back) again due to some other bug.

Aka write test for a known broken things which is simply adjusted, once fixed. And voila, that feature already has a (good) test now. IMHO a pattern which sometimes makes sense.

mfn:
Ops, unintentionally closed

mfn:
@barryvdh I'm still in favour of merging this test alerting us of the change, as explained before

barryvdh:
Hmm seems to fail

barryvdh (approved):

PR #986, opened 2020-07-01
Fix phpdoc generate for custom cast with parameter

### LINKED ISSUE #985: wrong type for model cast with parameters
`protected $casts = [ 'sell_price' => Money::class];` generate in docblock `/** @property \App\Support\Money $sell_price */`

but if add param to cast (`protected $casts = [ 'sell_price' => Money::class . ':sell_price'];`), ide-helper generate wrong (mixed) type  `/** @property mixed $sell_price */`

artelkr:
Closes #985


src/Console/ModelsCommand.php
```diff
@@ -332,7 +332,8 @@ protected function castPropertiesType($model)
                     $realType = '\Illuminate\Support\Collection';
                     break;
                 default:
-                    $realType = class_exists($type) ? ('\\' . $type) : 'mixed';
+                    $explodedType = head(explode(':', $type));
+                    $realType = class_exists($explodedType) ? ('\\' . $explodedType) : 'mixed';
```
mfn, line 336: Thanks, I suggest to simplify this; no need for explode and a helper function, [strtok](https://www.php.net/strtok) handles this exactly the way we need; this also means we can keep the original `$realType = ` line:
```suggestion
                    // In case of an optional custom cast parameter , only evaluate
                    // the `$type` until the `:`
                    $type = strtok($type, ':');
                    $realType = class_exists($type) ? ('\\' . $type) : 'mixed';
```

artelkr: thanks for that suggestion

mfn (approved):
@artelkr I took the liberty and adapted with master and added a changelog entry

@barryvdh from my side this is good to merge!

artelkr:
@mfn oh thanks, i completely forgot about it

mfn:
@artelkr it happens 😬 We also weren't super fast in merging, so there's that! Thanks for the PR!

PR #987, opened 2020-07-08
Created a positiblity to add custom relation type

efinder2:
In a project of mine I use the [eloquent-has-many-deep](https://github.com/staudenmeir/eloquent-has-many-deep) library. To enable the ide-helper for it, I wrote a possibility to add a config option to add custom relation types.

I know the code isn't tested yet. I'll add it as soon as the idea of the pull request has been discussed.


src/Console/ModelsCommand.php
```diff
@@ -36,6 +36,20 @@
...
+        'hasManyThrough' => '\Illuminate\Database\Eloquent\Relations\HasManyThrough',
+        'hasOneThrough' => '\Illuminate\Database\Eloquent\Relations\HasOneThrough',
+        'belongsToMany' => '\Illuminate\Database\Eloquent\Relations\BelongsToMany',
+        'hasOne' => '\Illuminate\Database\Eloquent\Relations\HasOne',
+        'belongsTo' => '\Illuminate\Database\Eloquent\Relations\BelongsTo',
+        'morphOne' => '\Illuminate\Database\Eloquent\Relations\MorphOne',
+        'morphTo' => '\Illuminate\Database\Eloquent\Relations\MorphTo',
+        'morphMany' => '\Illuminate\Database\Eloquent\Relations\MorphMany',
+        'morphToMany' => '\Illuminate\Database\Eloquent\Relations\MorphToMany',
+        'morphedByMany' => '\Illuminate\Database\Eloquent\Relations\MorphToMany'
```
mfn, line 50: I suggest to start using the `:class` syntax here then, now that the code is already touched

src/Console/ModelsCommand.php
```diff
@@ -886,6 +888,15 @@ protected function getCollectionClass($className)
         return '\\' . get_class($model->newCollection());
     }
 
+    /**
+     * Returns the available relation types
+     */
+    public function getRelationTypes(): array
```
mfn, line 894: I don't think it needs to be `public` since the basic feature is already configurable?
```suggestion
    protected function getRelationTypes(): array
```

src/Console/ModelsCommand.php
efinder2: Has been resolved in https://github.com/barryvdh/laravel-ide-helper/pull/987/commits/392de13a19d6b4d87e9d0f8cf08d51f5a59ff128

mr-feek:
Is this something we can/should solve via reflection?

efinder2:
It might not be the perfect solution at the moment. Not all code uses a strict return type. The linked library for example hasn't implemented them yet, so reflection might be a great addition to get rid of the mapping for types methods. As fallback I suggest to keep the config option

mfn:
Personally I would reduce the time invested, the config approach seems a pragmatic approach. I bet it doesn't come up often. First time I hear about this requirement in years 😄

efinder2:
I've looked into the source from Laravel. The [relations](https://github.com/laravel/framework/blob/7cf519aa0529031a15f1cd7707b85d134317cf0a/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php#L348) don't have a return type. So the reflection would need Parsing of the return type of the doc block. This isn't ideal.

efinder2:
Only Problem I see in the config approach, may be two libraries modifying the config. For future extension it might be helpful to create a HelperClass for adding class mappings to the config property.
I've also added a test for the change.

config/ide-helper.php
```diff
@@ -242,4 +242,16 @@
...
     'include_class_docblocks' => false,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Additional relation types
+    |--------------------------------------------------------------------------
+    |
+    | Sometimes it's needed to create custom relation types. The key of the array
+    | is the Relationship Method name. The value of the array is the canonical class
+    | name of the Relationship.
```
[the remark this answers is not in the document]
mfn: I think sometimes samples convey more than words but also can't replace em' :)

mfn (approved):
@efinder2 I took the liberty and updated the PR with master, applied my own feedback and added a changelog entry

@barryvdh IMHO this is good to merge!

barryvdh:
Hmm it has conflicts again

barryvdh (approved):

mfn:
@barryvdh fixed!

We'll likely see this more often now with the changelog, adding entries at the top I fear…

PR #990, opened 2020-07-14
Add the option to hide the generated datetime from the generated files

tvbeek:
On some projects I want to be sure that we didn't miss anything in the generated files. That is why I prefer to run the generate commands in the pipeline.

Because of the timestamps the files are always changed, with this PR it is possible to hide the datetime from the generated files so it is possible to see automated if there is any change found and a new file need to be committed.

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -242,4 +242,8 @@ return array(
     */
     'include_class_docblocks' => false,
 
+    /**
+     * Hide the datetime from the generated files
+     */
+    'hide_datetime' => false,
 );
--- a/resources/views/helper.php
+++ b/resources/views/helper.php
@@ -12,7 +12,7 @@
 
 /**
  * A helper file for Laravel, to provide autocomplete information to your IDE
- * Generated for Laravel <?= $version ?> on <?= date("Y-m-d H:i:s") ?>.
+ * Generated for Laravel <?= $version ?><?php if(!(isset($hide_datetime) && $hide_datetime)): ?> on <?= date("Y-m-d H:i:s") ?><?php endif;?>.
  *
  * This file should not be included in your code, only analyzed by your IDE!
  *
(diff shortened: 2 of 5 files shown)


config/ide-helper.php
```diff
@@ -242,4 +242,8 @@
     */
     'include_class_docblocks' => false,
 
+    /**
+     * Hide the datetime from the generated files
+     */
```
mfn, line 247: the format of the comment looks nothing like the others in this file

resources/views/helper.php
```diff
@@ -12,7 +12,7 @@
 
 /**
  * A helper file for Laravel, to provide autocomplete information to your IDE
- * Generated for Laravel <?= $version ?> on <?= date("Y-m-d H:i:s") ?>.
+ * Generated for Laravel <?= $version ?><?php if(!(isset($hide_datetime) && $hide_datetime)): ?> on <?= date("Y-m-d H:i:s") ?><?php endif;?>.
```
mfn, line 15: The PR always set the variable for the templates, so it seems to me `isset` isn't necessary?

mfn: Might I suggest a different approach?

In stead of "if not hide: echo $date; endif" how about the inverse / positive check:
- call the var `show_datetime`
- default to `true`
- write a "positive" check

```suggestion
 * Generated for Laravel <?= $version ?><?php if ($show_datetime) ?> on <?= date("Y-m-d H:i:s") ?><?php endif;?>.
```

Seems less cluttered and also more straight to the point.

(applies also for the check in the other file)

tvbeek:
@mfn thanks for the feedback, I will wait on the answer from @barryvdh on your question about removing instead of adding another config. If that is clear I will update this PR.

mfn:
Makes sense!

----
FTR, here are the reasons we've it in the first place:
- Basically since the beginning the date was added without man fanfare
  => https://github.com/barryvdh/laravel-ide-helper/commit/91033293b9fd879d1951838f913f0a6a3fbd852c
- Later the time was added:
  > Show time in generated files / I think this is useful for debugging.
  
  => https://github.com/barryvdh/laravel-ide-helper/pull/574

The "think … is …useful" sounds a bit fishy to me 😄 I think debugging should happen on different levels and favouring CI workflows in 2020 is more appropriate (aka just removing it).

I think this whole thing may also comes from a time when those files were "shared" online via gists for specific Laravel version, a practice I'd say shouldn't be endorsed at all anymore.

mfn:
_Fun_ fact: I already mentioned that the `_ide_helper_models.php` does not include them.

In fact the test suite was recently refactored to use snapshots for the rewritten/generated PHP files and this wouldn't be possible if a date was written in them…

** :+1: ** for just removing them from my side

barryvdh:
I think I added it because in the earliest versions, the actual gist was used more often. That's why the version and time were included. I don't see a benefit of including it now. Might as well remove it altogether.

mfn:
Sent in https://github.com/barryvdh/laravel-ide-helper/pull/1000

tvbeek:
> 
> 
> Sent in #1000

You are fast! 👍

barryvdh:
Thanks

PR #991, opened 2020-07-15
composer.json: bump branch-alias as 2.7 has already been released

mfn:
ping @barryvdh

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.6-dev"
+            "dev-master": "2.7-dev"
         },
         "laravel": {
             "providers": [


mfn:
@mr-feek thanks for the 👍 , actually I haven't this used much before. Am I right it's supposed to the _the next_ version? I.e. current release is `2.7` and so this should be `2.8-dev`. _Or_ is `2.7-dev`? 🤷‍♀️

mr-feek:
It should be the next version. However, it's also probably totally acceptable to just alias to `2.x-dev` in this project. We don't need to update this field every release.

I'm quite to new to branch level aliases also, I can't be certain what the behavior is if dev-master is aliased to a version behind the latest release. I assume that would mean that if someone wanted to use dev-master, they would have to explicitly do so and wouldn't be able to rely on the alias.

https://getcomposer.org/doc/articles/aliases.md#branch-alias

mfn:
> I can't be certain what the behavior is if dev-master is aliased to a version behind the latest release

Well, I in fact tried to install it today using the dev alias but did not get the latest version 🤷‍♀️

I wasn't ware that `2.x-dev` (literally?) also works. Sounds better to not have to maintain this.

Anyone else with experience here? OTOH it would seem it's definitely safe to merge this one either way.

barryvdh:
Can we change it to 2.7 instead of 2.8?

mfn:
@barryvdh done!

mfn:
Can confirm, works; can now install dev-master also with 2.7-dev 🎉

barryvdh:
Yes next version will be 2.7.1 probably, so dev-version is also 2.7. Once we plan to release 2.8 we can bump it again.

PR #1003, opened 2020-07-31
[PHPCS] Extend to check/fix all source code and turn GHA into auto-commit fixer

mfn:
While reviewing some PRs today I noticed the inconsistency in test source code and wondered why.

Until I realized that phpcs is only applied to `src/`.

- enable phpcs for all dirs containing "code":
  - `config/`
  - `resources/`
  - `tests/`
- run the fixes incl. updating snapshots
- turn the GHA from checking into fixing it and pushing it back


.github/workflows/run-tests.yml
```diff
@@ -55,10 +55,20 @@ jobs:
         with:
           php-version: 7.4
           coverage: none
-        
+          tools: composer:v2
```
mfn, line 58: Makes installing dependencies faster, for me down from ~50s to ~30s

mfn, line 58: Makes installing dependencies faster, for me down from ~50s to ~30s

.github/workflows/run-tests.yml
```diff
@@ -55,10 +55,20 @@ jobs:
...
         run: |
           composer remove phpro/grumphp vimeo/psalm --no-update --dev
           composer update --prefer-dist --no-suggest --no-progress
 
-      - run: composer check-style
+      - run: composer fix-style
+        continue-on-error: true
+
+      # Revert modifications so they don't get commited 💥
+      - run: git checkout -- composer.json
```
mfn, line 69: There's `composer remove` step to avoid installing too many unnecessary dependencies, this must be rolled back before a potential commit hapens

mfn:
@barryvdh mildly related: are you open to switch to another CS fixer?

I noticed "unused import statements" in the tests (copypaste victims I guess) but phpcs can't handle it ( https://github.com/squizlabs/PHP_CodeSniffer/issues/801 ) whereas e.g. https://github.com/FriendsOfPHP/PHP-CS-Fixer supports it (`no_unused_imports` rule).

.github/workflows/run-tests.yml
```diff
@@ -55,10 +55,20 @@ jobs:
...
+
+      - uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: composer fix-style
+          commit_author: laravel-ide-helper <laravel-ide-helper@users.noreply.github.com>
```
mfn, line 74: I think we might need a token here so it properly works, see e.g. https://github.com/spatie/data-transfer-object/blob/bf7f41df1765b9779f38bef2dd572a707c7d1179/.github/workflows/style.yml#L28-L29

I didn't move forward yet as I can't manage secrets and want to wait whether this will even be accepted 🙏

barryvdh:
Yeah only thing I'm a bit worried about is the merge conflicts on existing pull requests. But for future ones, this would be better indeed to avoid them.

mfn:
@barryvdh I think the impact is negligible because we're talking about the _least touched_ code parts here. `src/` is already checked and the others rarely change

mfn:
❤️

barryvdh:
@mfn So I also tested this in Debugbar, but it looks like when it encounters an issue, it DOES fix the issue, but the rest of the actions are not executed. See https://github.com/barryvdh/laravel-debugbar/pull/1088 after the first commit. A 2nd commit is added that indeed fixes the CS, but then not tests. And I have to re-trigger it with a new commit.
Screenshot 2020-08-17 at 14 49 32

Any ideas?

barryvdh:
I wonder if it helps to split them in 2 separate actions 🤔

mfn:
> but the rest of the actions are not executed

I think it may be related to my comment https://github.com/barryvdh/laravel-ide-helper/pull/1003#discussion_r465122222
> I think we might need a token here so it properly works

See also https://github.com/stefanzweifel/git-auto-commit-action/#commits-of-this-action-do-not-trigger-new-workflow-runs
> When you use the repository's GITHUB_TOKEN to perform tasks on behalf of the GitHub Actions app, events triggered by the GITHUB_TOKEN will not create a new workflow run.
>
> You can change this by creating a new Personal Access Token (PAT), storing the token as a secret in your repository and then passing the new token to the actions/checkout Action step.
>
> - uses: actions/checkout@v2
>   with:
>     token: ${{ secrets.PAT }}

But I'm not sure this is even necessary!

- the workflow which initially triggers the php-cs-fix commit runs through correctly, i.e. the `fix-style` job always is green so the accumulative result is correct and whether or not php-cs-fix does something will not change this
- pushing the changes doesn't create a new workflow but the result from the previous one, which triggered this one, is unchanged _anyway_

The downside I can imagine is that when you view the PR, you see the last workflow triggered by php-cs-fix which actually didn't run any jobs, meaning you don't see that status (but it's there from the run before).

Did I understand your concern correctly?

barryvdh:
Yeah I don't need to run the workflow again after the cs fix is done, but I do need to have them run in the first place.

mfn:
> but I do need to have them run in the first place

But this shouldn't have been prevented…

Can you link to the specific action/workflow run?

barryvdh:
Yeah see the linked PR and then the first commit, with the red cross.

mfn:
I think I didn't see it for the longest time, you mean that the "fix code style" job on the `pull_request` trigger didn't work:
https://github.com/barryvdh/laravel-debugbar/runs/990971282#step:7:16
```
Run stefanzweifel/git-auto-commit-action@v4
Started: bash /home/runner/work/_actions/stefanzweifel/git-auto-commit-action/v4/entrypoint.sh
INPUT_REPOSITORY value: .
INPUT_BRANCH value: feat-test-inject
error: pathspec 'feat-test-inject' did not match any file(s) known to git
Error: Invalid status code: 1
    at ChildProcess.<anonymous> (/home/runner/work/_actions/stefanzweifel/git-auto-commit-action/v4/index.js:17:19)
    at ChildProcess.emit (events.js:210:5)
    at maybeClose (internal/child_process.js:1021:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:283:5) {
  code: 1
}
Error: Invalid status code: 1
    at ChildProcess.<anonymous> (/home/runner/work/_actions/stefanzweifel/git-auto-commit-action/v4/index.js:17:19)
    at ChildProcess.emit (events.js:210:5)
    at maybeClose (internal/child_process.js:1021:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:283:5)
```
Maybe it makes sense and we can't run this on `pull_request` but only on `push` (where it worked!)

But IMHO it makes sense to not re-trigger the tests _per se_ because the style fix is not intended to change the outcome of this. But OTOH it shouldn't be responsible for failing.

But if you change the `action/checkout` with a token, you can have this behavour.

----
In a nutshell: it seems a separate workflow file only on push, optionally with a PAT for checkout, could work this around!

PR #1006, opened 2020-08-08
Add better support for macros and mixins

domkrm:
I have added macro support for all macroable classes, not only the alias classes.
But this only works if the macros are declared in any loaded service provider and type hinting is used.
Also no description is possible, only parameter and return types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,18 @@ You can use an in-memory SQLite driver by adding the `-M` option.
 You can choose to include helper files. This is not enabled by default, but you can override it with the `--helpers (-H)` option.
 The `Illuminate/Support/helpers.php` is already set up, but you can add/remove your own files in the config file.
 
+### Automatic PHPDoc generation for macros and mixins
+
+This package can generate PHPDocs for macros and mixins which will be added to the `_ide_helper.php` file.
+
+But this only works if you use type hinting when declaring a macro.
+
+```php
+Str::macro('concat', function(string $str1, string $str2) : string {
+    return $str1 . $str2;
+});
+```
+
 ### Automatic PHPDocs for models
 
 If you don't want to write your properties yourself, you can use the command `php artisan ide-helper:models` to generate
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -13,6 +13,9 @@ namespace Barryvdh\LaravelIdeHelper;
 
 use Illuminate\Foundation\AliasLoader;
 use Illuminate\Support\Collection;
+use Illuminate\Support\Str;
+use Illuminate\Support\Traits\Macroable;
+use ReflectionClass;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class Generator
@@ -220,7 +223,11 @@ class Generator
      */
     protected function getAliasesByExtendsNamespace()
     {
-        return $this->getValidAliases()->groupBy(function (Alias $alias) {
+        $aliases = $this->getValidAliases();
+
+        $this->addMacroableClasses($aliases);
+
+        return $aliases->groupBy(function (Alias $alias) {
             return $alias->getExtendsNamespace();
         });
     }
@@ -290,5 +297,56 @@ class Generator
         } else {
             echo $string . "\r\n";
         }
+    }
+
+    /**
+     * Add all macroable classes which are not already loaded as an alias and have defined macros.
+     *
+     * @param Collection $aliases
+     */
+    protected function addMacroableClasses(Collection $aliases)
+    {
+        $macroable = $this->getMacroableClasses($aliases);
+
+        foreach ($macroable as $class) {
+            $reflection = new ReflectionClass($class);
+
+            if (!$reflection->getStaticProperties()['macros']) {
+                continue;
+            }
+
+            $aliases[] = new Alias($this->config, $class, $class, [], $this->interfaces);
+        }
+    }
+
+    /**
+     * Get all loaded macroable classes which are not loaded as an alias.
+     *
+     * @param Collection $aliases
+     * @return Collection
+     */
+    protected function getMacroableClasses(Collection $aliases)
+    {
+        return (new Collection(get_declared_classes()))
+            ->filter(function ($class) {
+                $reflection = new ReflectionClass($class);
+
+                // Filter out internal classes and class aliases
+                return !$reflection->isInternal() && $reflection->getName() === $class;
+            })
+            ->filter(function ($class) {
+                $traits = class_uses($class);
+
+                // Filter only classes with the macroable trait
+                return isset($traits[Macroable::class]);
+            })
+            ->filter(function ($class) use ($aliases) {
+                $class = Str::start($class, '\\');
+
+                // Filter out aliases
+                return !$aliases->first(function (Alias $alias) use ($class) {
+                    return $alias->getExtends() === $class;
+                });
+            });
     }
 }
--- a/src/Macro.php
+++ b/src/Macro.php
@@ -3,6 +3,7 @@
 namespace Barryvdh\LaravelIdeHelper;
 
 use Barryvdh\Reflection\DocBlock;
+use Barryvdh\Reflection\DocBlock\Tag;
 
 class Macro extends Method
 {
@@ -30,7 +31,26 @@ class Macro extends Method
      */
     protected function initPhpDoc($method)
     {
-        $this->phpdoc = new DocBlock($method);
+        $this->phpdoc = new DocBlock('/** */');
+
+        // Add macro parameters
+        foreach ($method->getParameters() as $parameter) {
+            $type = $parameter->hasType() ? $parameter->getType()->getName() : 'mixed';
+            $type .= $parameter->hasType() && $parameter->getType()->allowsNull() ? '|null' : '';
+
+            $name = $parameter->isVariadic() ? '...' : '';
+            $name .= '$' . $parameter->getName();
+
+            $this->phpdoc->appendTag(Tag::createInstance("@param {$type} {$name}"));
+        }
+
+        // Add macro return type
+        if ($method->hasReturnType()) {
+            $type = $method->getReturnType()->getName();
+            $type .= $method->getReturnType()->allowsNull() ? '|null' : '';
+
+            $this->phpdoc->appendTag(Tag::createInstance("@return {$type}"));
+        }
     }
 
     /**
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -110,6 +110,17 @@ You can use an in-memory SQLite driver by adding the `-M` option.
...
 
+### Automatic PHPDoc generation for macros
+
+The generator can detect macros and mixins which are declared in any loaded service provider.
+A basic autocompletion with parameter and return types can be generated if you use type hinting:
+
+```php
+Str::macro('concat', function(string $str1, string $str2) : string {
+    return $str1 . $str2;
+});
```
mfn, line 121: When I try this, it doesn't autocomplete:
image

Because the method isn't declared as `static`, here's the generated stub from `_ide_helper.php`:
```php
    namespace Illuminate\Support {
    
        class Str {
        
            /**
         * No description available.
         *
         * @param string $str1
         * @param string $str2
         * @return string 
         */
            public function concat($str1, $str2) { }
        
        }
    
    }
```
When I manually change it, it works:
image

Without the `static`, you _can_ hover over it and it shows you the non-static callable version.

I guess it's "either" or "or": static or not, because the `Macroable` trait itself simply looks up either way it's called and doesn't really care.

Would it work if instead of generating the stub with PHP code methods, we would generated something like this:
```php
    namespace Illuminate\Support {

        /**
         * @method static string concat(string $str1, $str2)
         * @method string concat(string $str1, $str2)
         */
        class Str {
        }

    }
```
🤔

src/Macro.php
```diff
@@ -24,14 +24,39 @@ public function __construct(
...
      */
     protected function initPhpDoc($method)
     {
-        $this->phpdoc = new DocBlock($method);
+        // Init DocBlock only with a description
+        $this->phpdoc = new DocBlock('/** No description available. */');
+
+        // Add macro parameters
+        foreach ($method->getParameters() as $parameter) {
+            $type = $parameter->hasType() ? $parameter->getType() : 'mixed';
```
mfn, line 44:
```
./artisan ide-helper:generate

   ErrorException

  Function ReflectionType::__toString() is deprecated

  at vendor/barryvdh/laravel-ide-helper/src/Macro.php:45
    41|
    42|         // Add macro parameters
    43|         foreach ($method->getParameters() as $parameter) {
    44|             $type = $parameter->hasType() ? $parameter->getType() : 'mixed';
  > 45|             $type .= $parameter->hasType() && $parameter->getType()->allowsNull() ? '|null' : '';
    46|
    47|             $name = $parameter->isVariadic() ? '...' : '';
    48|             $name .= '$' . $parameter->getName();
    49|

      +21 vendor frames
  22  artisan:37
      Illuminate\Foundation\Console\Kernel::handle()
```

I guess this is dependent on the PHP version? It's "just" a deprecation but people might run with E_ALL and thus trip over this.

This fixes it:
```suggestion
            $type = $parameter->hasType() ? $parameter->getType()->getName() : 'mixed';
```

src/Macro.php
```diff
@@ -24,14 +24,39 @@ public function __construct(
...
+
+            $name = $parameter->isVariadic() ? '...' : '';
+            $name .= '$' . $parameter->getName();
+
+            $this->phpdoc->appendTag(Tag::createInstance("@param {$type} {$name}"));
+        }
+
+        // Add macro return type
+        if ($method->hasReturnType()) {
+            $type = $method->getReturnType();
```
mfn, line 55: Here I have the same issue:
```
$ ./artisan ide-helper:generate

   ErrorException

  Function ReflectionType::__toString() is deprecated

  at vendor/barryvdh/laravel-ide-helper/src/Macro.php:56
    52|
    53|         // Add macro return type
    54|         if ($method->hasReturnType()) {
    55|             $type = $method->getReturnType();
  > 56|             $type .= $method->getReturnType()->allowsNull() ? '|null' : '';
    57|
    58|             $this->phpdoc->appendTag(Tag::createInstance("@return {$type}"));
    59|         }
    60|     }

      +21 vendor frames
  22  artisan:37
      Illuminate\Foundation\Console\Kernel::handle()
```
```suggestion
            $type = $method->getReturnType()->getName();
```

mfn (changes requested):
Hey 👋 very cool approach!

I left some in-line feedback.

----
Also: All / some? of the whitespace changes in `resources/views/helper.php` generate a _vastly_ differently indented `_ide_helper.php` file:
- run master, generate _ide_helper.php , make a copy
- run your PR, generate _ide_helper.php
- diff them
- 💥

People tend to commit this file and those two versions are essentially "uncomparable".

I don't want to say it's the most important thing in the world, but some of those changes just look unnecessary anyway, so you might want to massage this a bit.

Maybe also rebase against master, there were some formatting changes already, creating merge conflicts currently.

----
Seeing that I had some exceptions thrown my way, we definitely should come up with some tests for this.

README.md
domkrm: I think I have solved the problem by just adding the static tag. PhpStorm should now recognize the macros in static and non-static contexts.

src/Macro.php
domkrm: I have changed this to your suggestion.

src/Macro.php
domkrm: I have changed this to your suggestion.

domkrm:
I have adopted all of your suggestions. But I also have changed many things because I noticed that classes could be duplicate in the helper file if they have macros (Facades, Arr, Str, ...).

src/Generator.php
```diff
@@ -291,4 +296,55 @@ protected function error($string)
...
+
+    /**
+     * Get all loaded macroable classes which are not loaded as an alias.
+     *
+     * @param Collection $aliases
+     * @return Collection
+     */
+    protected function getMacroableClasses(Collection $aliases)
+    {
+        return collect(get_declared_classes())
```
mfn, line 328: > `collect`

Please don't use global helpers, it's a bad practice doing this in libraries. None of the helpers from `Illuminate\Support\helpers.php` is currently in use.

Using the collection of course is fine!

src/Generator.php
```diff
@@ -291,4 +296,55 @@ protected function error($string)
...
+                $reflection = new ReflectionClass($class);
+
+                // Filter out internal classes and class aliases
+                return !$reflection->isInternal() && $reflection->getName() == $class;
+            })
+            ->filter(function ($class) {
+                $traits = class_uses($class);
+
+                // Filter only classes with the macroable trait
+                return isset($traits['Illuminate\Support\Traits\Macroable']);
```
mfn, line 339: Please use `Illuminate\Support\Traits\Macroable::class` and also import it, thanks!

src/Macro.php
```diff
@@ -31,7 +31,26 @@ public function __construct(
...
-        $this->phpdoc = new DocBlock($method);
+        $this->phpdoc = new DocBlock('/** */');
+
+        // Add macro parameters
+        foreach ($method->getParameters() as $parameter) {
+            $type = $parameter->hasType() ? $parameter->getType()->getName() : 'mixed';
+            $type .= $parameter->hasType() && $parameter->getType()->allowsNull() ? '|null' : '';
+
+            $name = $parameter->isVariadic() ? '...' : '';
+            $name .= '$' . $parameter->getName();
```
mfn, line 42: Looks very clean but also duplicate code of what happens in `\Barryvdh\LaravelIdeHelper\Method::getParameters`

- the other method seems more feature-complete though I can't say if this is necessary
- but it's written in a way which makes it not easily re-usable AFAICS, i.e. you just can't call it.

Would you be interested comparing the the two approaches and see if yours is missing something? E.g. default values seems to be handled, etc.

----
Unifying this is topic for a different PR IMHO and shouldn't be done here.

src/Generator.php
```diff
@@ -291,4 +296,55 @@ protected function error($string)
...
+     * @return Collection
+     */
+    protected function getMacroableClasses(Collection $aliases)
+    {
+        return collect(get_declared_classes())
+            ->filter(function ($class) {
+                $reflection = new ReflectionClass($class);
+
+                // Filter out internal classes and class aliases
+                return !$reflection->isInternal() && $reflection->getName() == $class;
```
mfn, line 333:
```suggestion
                return !$reflection->isInternal() && $reflection->getName() === $class;
```

src/Generator.php
```diff
@@ -291,4 +296,55 @@ protected function error($string)
...
+
+                // Filter only classes with the macroable trait
+                return isset($traits['Illuminate\Support\Traits\Macroable']);
+            })
+            ->filter(function ($class) use ($aliases) {
+                $class = Str::start($class, '\\');
+
+                // Filter out aliases
+                return !$aliases->first(function (Alias $alias) use ($class) {
+                    return $alias->getExtends() == $class;
```
mfn, line 346:
```suggestion
                    return $alias->getExtends() === $class;
```

mfn (changes requested):
Approach LGTM and I only have some minor feedback; additionally:
- can you please rebase on master and solve the merge conflict?
- re-add the nice README entry you had?
- add a changelog entry?
  Yes, we now have a changelog 😄 

----
Thanks and sorry for the long delay, rest assured unless I get hit by a bus, the next feedback will be quicker 🤞

mfn:
Btw, do we still need the code in `\Barryvdh\LaravelIdeHelper\Alias::detectMethods` 🤔 What's the difference?
```php
            // Check if the class is macroable
            $traits = collect($reflection->getTraitNames());
            if ($traits->contains('Illuminate\Support\Traits\Macroable')) {
                $properties = $reflection->getStaticProperties();
                $macros = isset($properties['macros']) ? $properties['macros'] : [];
                foreach ($macros as $macro_name => $macro_func) {
                    if (!in_array($macro_name, $this->usedMethods)) {
                        // Add macros
                        $this->methods[] = new Macro(
                            $this->getMacroFunction($macro_func),
                            $this->alias,
                            $reflection,
                            $macro_name,
                            $this->interfaces
                        );
                        $this->usedMethods[] = $macro_name;
                    }
                }
            }
```

domkrm:
I have done a rebase and applied all your suggestions.

domkrm:
> Btw, do we still need the code in `\Barryvdh\LaravelIdeHelper\Alias::detectMethods` 🤔 What's the difference?
> 
> ```
>             // Check if the class is macroable
>             $traits = collect($reflection->getTraitNames());
>             if ($traits->contains('Illuminate\Support\Traits\Macroable')) {
>                 $properties = $reflection->getStaticProperties();
>                 $macros = isset($properties['macros']) ? $properties['macros'] : [];
>                 foreach ($macros as $macro_name => $macro_func) {
>                     if (!in_array($macro_name, $this->usedMethods)) {
>                         // Add macros
>                         $this->methods[] = new Macro(
>                             $this->getMacroFunction($macro_func),
>                             $this->alias,
>                             $reflection,
>                             $macro_name,
>                             $this->interfaces
>                         );
>                         $this->usedMethods[] = $macro_name;
>                     }
>                 }
>             }
> ```

Yes, this code is still needed. My old approach had an own function to detect macros but my new approach uses this part of the code.

mfn (approved):
Thanks, IMHO this looks great `\o/`

@barryvdh IMHO we're good to merge here!

PR #1009, opened 2020-08-18
Fix typos; Optimise if-else branches

DannyvdSluijs:
This PR is cleaning low hanging fruits to bring it up to date:
- Correct typos
- Optimise if-else branches to reduce indentation
- Remove Docblocks for functions which contain type definitions.
- Cleanup unused params and namespace imports

Please feel free to comment and suggest other approaches.

### DIFF
--- a/src/Eloquent.php
+++ b/src/Eloquent.php
@@ -85,25 +85,29 @@ class Eloquent
         }
 
         $filename = $reflection->getFileName();
-        if ($filename) {
-            $contents = $files->get($filename);
-            if ($contents) {
-                $count    = 0;
-                $contents = str_replace($originalDoc, $docComment, $contents, $count);
-                if ($count > 0) {
-                    if ($files->put($filename, $contents)) {
-                        $command->info('Wrote expected docblock to ' . $filename);
-                    } else {
-                        $command->error('File write failed to ' . $filename);
-                    }
-                } else {
-                    $command->error('Content did not change ' . $contents);
-                }
-            } else {
-                $command->error('No file contents found ' . $filename);
-            }
-        } else {
+        if (!$filename) {
             $command->error('Filename not found ' . $class);
+            return;
         }
+
+        $contents = $files->get($filename);
+        if (!$contents) {
+            $command->error('No file contents found ' . $filename);
+            return;
+        }
+
+        $count = 0;
+        $contents = str_replace($originalDoc, $docComment, $contents, $count);
+        if ($count <= 0) {
+            $command->error('Content did not change ' . $contents);
+            return;
+        }
+
+        if (!$files->put($filename, $contents)) {
+            $command->error('File write failed to ' . $filename);
+            return;
+        }
+
+        $command->info('Wrote expected docblock to ' . $filename);
     }
 }
--- a/src/Method.php
+++ b/src/Method.php
@@ -251,35 +251,37 @@ class Method
      */
     protected function normalizeReturn(DocBlock $phpdoc)
     {
-        //Get the return type and adjust them for beter autocomplete
+        //Get the return type and adjust them for better autocomplete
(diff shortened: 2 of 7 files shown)


mfn:
I tried similar attempts but start with code style first, outcome was https://github.com/barryvdh/laravel-ide-helper/pull/871#issuecomment-617360055

DannyvdSluijs:
@mfn I get your point but I also think this PR is a bit different. If needed I can reduce the PR to not move to`[]` for array creation. IMHO the other items fixed are actually thing that contribute to the overall value.

Also assuming that reduce the number of open PR's would make the argument void ? I can assist in seeing if we can get the number of open PR's down.

mfn:
The thing is: I think manually fixing some of the things you changed is pointless.
- `array()` ? they should be auto-converted.
- `list()` ? same
- `static function () {` ? same, should be automated
- maybe others, etc.

There's nothing won if the next PR uses a different style and reviewers have to point things out which should be automatically fixable.

Of course some of the other code changes are unrelated to these kind of style changes, I understand.

OTOH, if you look at the test suite, it's kind in its infancy. So many parts you changed are not covered and TBH, reviewing these kind of code changes is hard / takes a lot of time.

Personally I think time should be rather invested in:
- tools automating the mundane tasks which can't break things
- increase test coverage
- so we can finally rock the boat and reduce the spaghetti-ness 🍝 

This is an old / grown  and,  wouldn't say complex but at times unintuitive code-base, so care has to be taken to not break things left and right.

DannyvdSluijs:
Ive reduced the changes and taken out the `array()` swap for `[]`.
PR has been rebased to upstream master.

src/Eloquent.php
```diff
@@ -85,25 +85,29 @@ public static function writeEloquentModelHelper(Command $command, Filesystem $fi
...
+
+        $contents = $files->get($filename);
+        if (!$contents) {
+            $command->error('No file contents found ' . $filename);
+            return;
+        }
+
+        $count = 0;
+        $contents = str_replace($originalDoc, $docComment, $contents, $count);
+        if ($count <= 0) {
```
mfn, line 101: How can `$count` be < 0 when returned from `str_replace`?

src/Method.php
```diff
@@ -251,35 +251,37 @@ protected function normalizeParams(DocBlock $phpdoc)
...
-            $tag->setContent($returnValue . ' ' . $tag->getDescription());
-            $this->return = $returnValue;
 
-            if ($tag->getType() === '$this') {
-                Str::contains($this->root, Builder::class)
-                    ? $tag->setType($this->root . '|static')
-                    : $tag->setType($this->root);
-            }
-        } else {
+        if (count($returnTags) === 0) {
```
mfn, line 257: IMHO this is fine, it's PHP after all and it's not ambiguous in this context:
```suggestion
        if (!$returnTags) {
```

mfn (approved):
Yay, thanks for getting rid of the worst offenders 😄 

#protip : review with https://github.com/barryvdh/laravel-ide-helper/pull/1009/files?w=1 to keep your sanity

Left some smaller notes, I think this is good => leaving it up for @barryvdh

src/Eloquent.php
DannyvdSluijs: I thing your misreading between `$count` and `$contents`. Count is an optional parameter, passed by reference, that if passed will hold the number of matched and replaced needles.

mfn: > Count is an optional parameter, passed by reference, that if passed will hold the number of matched and replaced needles.

Yes, I think I got this right. But it can't be `< 0`, can it? It can be 0 and > 0.

In a sense, you could just write `if (!$count) {` too but I guess `if ($count === 0) {` is the most crystal clear.

PR #1014, opened 2020-08-23
Fix #1004 Auto-generate PHPDoc for Illuminate\Support\Carbon

### LINKED ISSUE #1004: Model when use Carbon, auto generate docblock not pass by phpstan.
this is my User Model diff after run `php artisan ide-helper:models -WR`

image

it will remove `\Carbon\Carbon` in docblock, would lead to result by phpstan: (https://github.com/nunomaduro/larastan)

```
 ------ -----------------------------------------------------------------------
  Line   User.php
 ------ -----------------------------------------------------------------------
  385    Property App\User::$vip_end (Illuminate\Support\Carbon|null) does not
         accept Carbon\Carbon.
  388    Property App\User::$vip_end (Illuminate\Support\Carbon|null) does not
         accept Carbon\Carbon.
  394    Property App\User::$vip_end (Illuminate\Support\Carbon|null) does not
         accept Carbon\Carbon.
 ------ -----------------------------------------------------------------------
```

Is there something I'm doing wrong? Please advise, thank you very much!

kylekatarnls:
Fix #1004 Auto-generate PHPDoc for Illuminate\Support\Carbon from Carbon\Carbon

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -11,10 +11,12 @@
 
 namespace Barryvdh\LaravelIdeHelper;
 
+use Carbon\Carbon;
 use Illuminate\Foundation\Application;
 use Illuminate\Foundation\AliasLoader;
 use Illuminate\Support\Collection;
 use ReflectionClass;
+use ReflectionException;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class Generator
@@ -82,6 +84,7 @@ class Generator
     {
         $app = app();
         return $this->view->make('helper')
+            ->with('carbonPhpDoc', $this->getCarbonPhpDoc())
             ->with('namespaces_by_extends_ns', $this->getAliasesByExtendsNamespace())
             ->with('namespaces_by_alias_ns', $this->getAliasesByAliasNamespace())
             ->with('helpers', $this->helpers)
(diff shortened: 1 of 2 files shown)


kylekatarnls:
This could help PHPStan and IDE to understand that `Illuminate\Support\Carbon` return `Illuminate\Support\Carbon` instances for methods where the `Carbon\Carbon` PHPDoc mention `Carbon\Carbon` return hints in `@method` annotations.

src/Generator.php
```diff
@@ -82,6 +84,7 @@ public function generatePhpHelper()
     {
         $app = app();
         return $this->view->make('helper')
+            ->with('carbonPhpDoc', $this->getCarbonPhpDoc())
```
kylekatarnls, line 87: Should this variable be in snake_case?

kylekatarnls:
I tested it with the following:
```php
/**
 * @property \Illuminate\Support\Carbon|null $last_action_at
 */
class User
{
  public function foo()
  {
    $this->last_action_at = now()->addDays(5);
  }
}
```
And this still throw the PHPStan error because both the `Illuminate\Support\Carbon` and the `Carbon\Carbon` PHPDoc are considered and so the return of addDays() is saw as `Illuminate\Support\Carbon | Carbon\Carbon` (the same happens for the IDE auto-completion).

mfn:
@kylekatarnls I think we can close this one with the outcome of https://github.com/barryvdh/laravel-ide-helper/issues/1004 ?

kylekatarnls:
Yep

PR #1017, opened 2020-08-24
Test dynamic relationships

barryvdh:
Relationships should be detected based on static analysis, so dynamic properties should not be an issue.
(Test currently failing)

### DIFF
--- a/tests/SnapshotPhpDriver.php
+++ b/tests/SnapshotPhpDriver.php
@@ -9,7 +9,7 @@ class SnapshotPhpDriver implements Driver
 {
     public function serialize($data): string
     {
-        return $data;
+        return (string) $data;
     }
 
     public function extension(): string
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -570,7 +570,11 @@ class ModelsCommand extends Command
                             // can cause errors. Since we don't need constraints we can
                             // disable them when we fetch the relation to avoid errors.
                             $relationObj = Relation::noConstraints(function () use ($model, $method) {
-                                return $model->$method();
+                                try {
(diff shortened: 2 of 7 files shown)


barryvdh:
@mfn I don't understand this testcase exactly. Why is the Display output empty? It throws an error on the commandline, but I can't seem to fetch it. Otherwise it should show that 1 model is generated, so not sure why we expect it to be empty.

@stefanbauer this replicates your issue. Now just need to fix it ;)

barryvdh:
So this is not ideal, but not sure how to get the actual relation without executing the method. It could be with static analysis (php-parser?) but not sure.

barryvdh:
So best option: Use static analysis to detect the type without even running it
2nd best option: add the relationship using default Model instead of the actual model (which at least gives some phpdocs)
3rd best: catch the error (this PR)

barryvdh:
Although static analysis might be doable because we already search for the relationship definition in the code now..

mfn (approved):
LGTM, thanks for the test!

----
> Why is the Display output empty? It throws an error on the commandline, but I can't seem to fetch it

Good question 😅 I'll look into it!

tests/SnapshotPhpDriver.php
```diff
@@ -9,7 +9,7 @@ class SnapshotPhpDriver implements Driver
 {
     public function serialize($data): string
     {
-        return $data;
+        return (string) $data;
```
mfn, line 12: Btw, are you sure this is required? I forgot to mention this, but I tested your PR without the casts and it worked.

Or maybe was it due to error-case I guess?

barryvdh: Yeah in that case it was null.

PR #1019, opened 2020-08-25
Simplify test by mocking filesystem always

barryvdh:
This simplifies the individual tests a bit, because it was exactly the same anyways.

 - Allow full path for models (only if base_path() does not exists to keep BC)
 - Use reflection to get the test path from inside the abstract tests
 - Dynamic mocking of return value so we can define it once and remember output
 - Assert the from mock output directly
 - Load package config to test defaults.


src/Console/ModelsCommand.php
```diff
@@ -97,7 +97,7 @@ public function handle()
         $filename = $this->option('filename');
         $this->write = $this->option('write');
         $this->dirs = array_merge(
-            $this->laravel['config']->get('ide-helper.model_locations'),
+            $this->laravel['config']->get('ide-helper.model_locations', []),
```
barryvdh, line 100: Slightly unrelated, but will fail otherwise when the config is not set.

src/Console/ModelsCommand.php
```diff
@@ -272,7 +272,15 @@ protected function loadModels()
     {
         $models = array();
         foreach ($this->dirs as $dir) {
-            $dir = base_path() . '/' . $dir;
+            if (is_dir(base_path($dir))) {
```
barryvdh, line 275: Only convert to base_path() when it exists, otherwise use as-is

tests/Console/ModelsCommand/AbstractModelsCommand.php
```diff
@@ -21,6 +28,22 @@ protected function setUp(): void
...
+    /**
+     * Get package providers.
+     *
+     * @param  \Illuminate\Foundation\Application  $app
+     *
+     * @return array
+     */
+    protected function getPackageProviders($app)
+    {
+        return [IdeHelperServiceProvider::class];
```
barryvdh, line 46: This loads the model + config from like in a normal app.

tests/Console/ModelsCommand/AbstractModelsCommand.php
```diff
@@ -35,5 +58,45 @@ protected function getEnvironmentSetUp($app)
...
             'prefix' => '',
         ]);
+
+        // Load the Models from the Test dir
+        $config->set('ide-helper.model_locations', [
+            dirname((new \ReflectionClass(static::class))->getFileName()) . '/Models',
+        ]);
+
+        // Don't override integer -> int for tests
+        $config->set('ide-helper.type_overrides', []);
```
barryvdh, line 68: Because the defaults convert integer to int, the tests would fail with default config.

tests/Console/ModelsCommand/AbstractModelsCommand.php
```diff
@@ -35,5 +58,45 @@ protected function getEnvironmentSetUp($app)
...
+            });
+
+        $mockFilesystem
+            ->shouldReceive('put')
+            ->with(
+                Mockery::any(),
+                Mockery::any()
+            )
+            ->andReturnUsing(function ($path, $contents) {
+                $this->mockFilesystemOutput .= $contents;
```
barryvdh, line 90: Append the output, would make it easier to tests multiple models I guess?

mfn (approved):
Awesome!

I also was already unhappy with the boilerplate, thanks for beating me to it 😼

Issue #1024, opened 2020-08-30
[L8] Too few arguments to function Illuminate\View\Engines\PhpEngine::__construct()


SlyDeath:
With **Laravel 8** when i run `php artisan ide-helper:generate` or ` php artisan ide-helper:meta` i've got an error:

>  Too few arguments to function Illuminate\View\Engines\PhpEngine::__construct(), 0 passed in /vendor/barryvdh/laravel-ide-helper/src/IdeHelperServiceProvider.php on line 119 and exactly 1 expected  at /vendor/laravel/framework/src/Illuminate/View/Engines/PhpEngine.php:24

Problem is here:
https://github.com/barryvdh/laravel-ide-helper/blob/707ec990e7657d4a75d793895b996b3d7b7543e8/src/IdeHelperServiceProvider.php#L119

Need pass **Filesystem**:
```
return new PhpEngine($this->app['files']);
```
Or it's Laravel issue? Anyway why not add ¯\\_(ツ)_/¯

mfn:
> Or it's Laravel issue? Anyway why not add ¯\_(ツ)_/¯

You're right it's an "upcoming" bug for L8 because they added a constructor. ide-helper needs to be fixed here.

Are you up for an PR + tests which works in L7/L8?

SlyDeath:
I dont know what is "PR + tests", sorry :) 
I tried thus fix in mine the 7+ and 8+ projects. In both cases the commands were executed without errors.

mfn:
I see, no worries. I didn't realize you already made a PR in your repo and already started one.

SlyDeath:
Thanx!

mfn:
> I dont know what is "PR + tests", sorry :)

- PR => "Pull Request
- "tests" => write tests for the code change ;)

SlyDeath:
Where can I read the manual? :)

PR #1026, opened 2020-08-30
Make createLocalViewFactory compatible with Laravel 8

mfn:
## Summary
- Make createLocalViewFactory compatible with Laravel 8
- move `mockFilesystem` up so we can use it in other tests
- add a (really basic) test for `ide-helper:meta`
### Links
- Fixes https://github.com/barryvdh/laravel-ide-helper/issues/1024

### DIFF
--- a/src/IdeHelperServiceProvider.php
+++ b/src/IdeHelperServiceProvider.php
@@ -15,6 +15,7 @@ use Barryvdh\LaravelIdeHelper\Console\EloquentCommand;
 use Barryvdh\LaravelIdeHelper\Console\GeneratorCommand;
 use Barryvdh\LaravelIdeHelper\Console\MetaCommand;
 use Barryvdh\LaravelIdeHelper\Console\ModelsCommand;
+use Illuminate\Foundation\Application;
 use Illuminate\Support\ServiceProvider;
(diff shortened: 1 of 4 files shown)


mfn:
Before moving forward, I'd like to understand why the exiting L8 tests didn't choke on this.

The constructor wasn't added a day ago or so, it already exists when the first L8 tests were made.

mfn:
> Before moving forward, I'd like to understand why the exiting L8 tests didn't choke on this.

Ok, actually an easy one: because there are no tests for this 🤪

Took me longer because I thought there were and I saw the commands being resolved etc. but this was just because all console commands were always resolved, didn't mean they were executed. They weren't, so this code path isn't triggered.

I'll try to find a way to add a cheap tests here…

barryvdh:
Can't really remember why we resolve our own PhpEngine here, but okay. Is this ready for review?

mfn (commented):
> Can't really remember why we resolve our own PhpEngine here, but okay.

https://github.com/barryvdh/laravel-ide-helper/pull/475#issuecomment-290661408

> Is this ready for review?

It's now!

barryvdh (approved):

PR #1030, opened 2020-08-31
[CODESTYLE] Replace phpcs with php-cs-fixer

mfn:
This is a re-attempt of https://github.com/barryvdh/laravel-ide-helper/pull/871 based on the plea made in https://github.com/barryvdh/laravel-ide-helper/pull/871#issuecomment-683559682
> PSR-12 is fine by me and if you think php-cs-fixer does a better job then that's also fine with me :)

### Why replace one tool with another when they do the same job?
In my "IMHO" experience, php-cs-fixer is the friendlier tool for multiple reasons:
- better docs what fixers are available, how they're configured
  AFAIK today here's no such thing with php_codesniffer
- I think the out of the box experience is better, when I activated `@PSR2` I immediately got superfluous empty lines removed
- there are more (community velocity, contributions, defaults not to PEAR but PSR2, …)

### "u crazy that diff??"
About that… 😏 

- First, there's not yet a finished `@PSR12` preset so I used what's present from https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4943
  This incremential change to the source tree can be seen in commit https://github.com/barryvdh/laravel-ide-helper/commit/def43165c087393b694b420d41edadbe32aa31b2 and if you look at the changes, it's _mostly_ about redundant whitespace and other slight differences
- The point of a defined code style is to have a homogeneous code base so that certain patterns are equally handled, which wasn't the case so far.
  **Thus additional "fixers" have been enabled:**
  - short array_syntax (reason for most of the "noise" in the PR…)
  - no unused imports
  - trailing comma in multi-line arrays
  - single quotes
  - … and a few more
- I did create two configs, one for tests and the rest, because tests all work with `declare(strict_types=1);` but the rest currently doesn't => so for now we enforce it only for tests, until the rest catches up
- github actions was only slightly improved and grumphp was adapted
  Otherwise no real change here

🤞

### DIFF
--- a/.github/workflows/fix-code-style.yml
+++ b/.github/workflows/fix-code-style.yml
@@ -23,15 +23,10 @@ jobs:
           tools: composer:v2
 
       - name: Install dependencies
-        run: |
-          composer remove phpro/grumphp vimeo/psalm --no-update --dev
-          composer update --prefer-dist --no-progress
+        run: composer update --prefer-dist --no-progress
 
       - run: composer fix-style
         continue-on-error: true
-
-      # Revert manual modifications so they don't get commited 💥
-      - run: git checkout -- composer.json
 
       - uses: stefanzweifel/git-auto-commit-action@v4
         with:
--- a/composer.json
+++ b/composer.json
@@ -68,8 +68,8 @@
     "prefer-stable": true,
     "scripts": {
         "analyze": "psalm",
-        "check-style": "phpcs -p --standard=PSR12 config/ resources/ src/ tests/ '--ignore=*__snapshots_*'",
-        "fix-style": "phpcbf -p --standard=PSR12 config/ resources/ src/ tests/ '--ignore=*__snapshots_*'",
+        "check-style": "php-cs-fixer fix --diff --diff-format=udiff --dry-run",
+        "fix-style": "php-cs-fixer fix",
         "test": "phpunit",
         "test-regenerate": "phpunit -d --update-snapshots"
     }
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,7 @@
         "phpdocumentor/type-resolver": "^1.1.0"
     },
     "require-dev": {
+        "friendsofphp/php-cs-fixer": "^2",
         "illuminate/config": "^6 || ^7 || ^8",
         "illuminate/view": "^6 || ^7 || ^8",
         "mockery/mockery": "^1.3",
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,6 @@
         "orchestra/testbench": "^4 || ^5 || ^6",
         "phpro/grumphp": "^0.19.0",
         "spatie/phpunit-snapshot-assertions": "^1.4 || ^2.2 || ^3",
-        "squizlabs/php_codesniffer": "^3.5",
         "vimeo/psalm": "^3.12"
     },
     "config": {
--- a/composer.json
+++ b/composer.json
@@ -68,8 +68,14 @@
     "prefer-stable": true,
     "scripts": {
         "analyze": "psalm",
-        "check-style": "php-cs-fixer fix --diff --diff-format=udiff --dry-run",
-        "fix-style": "php-cs-fixer fix",
+        "check-style": [
+            "php-cs-fixer fix --diff --diff-format=udiff --dry-run",
+            "php-cs-fixer fix --diff --diff-format=udiff --dry-run --config=.php_cs.tests.php"
+        ],
+        "fix-style": [
+            "php-cs-fixer fix",
+            "php-cs-fixer fix --config=.php_cs.tests.php"
+        ],
         "test": "phpunit",
         "test-regenerate": "phpunit -d --update-snapshots"
     }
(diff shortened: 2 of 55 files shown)


.github/workflows/fix-code-style.yml
```diff
@@ -23,16 +23,11 @@ jobs:
...
-          composer remove phpro/grumphp vimeo/psalm --no-update --dev
-          composer update --prefer-dist --no-progress
+        run: composer update --prefer-dist --no-progress
 
       - run: composer fix-style
         continue-on-error: true
 
-      # Revert manual modifications so they don't get commited 💥
-      - run: git checkout -- composer.json
-
```
mfn, line 35: Didn't see the reason why we've to manually fiddle with packages which in turn requires to manually exclude `composer.json` from commiting etc. => "simplified" is the word I was looking for

composer.json
```diff
@@ -67,8 +67,14 @@
     "prefer-stable": true,
     "scripts": {
         "analyze": "psalm",
-        "check-style": "phpcs -p --standard=PSR12 config/ resources/ src/ tests/ '--ignore=*__snapshots_*'",
-        "fix-style": "phpcbf -p --standard=PSR12 config/ resources/ src/ tests/ '--ignore=*__snapshots_*'",
+        "check-style": [
+            "php-cs-fixer fix --diff --diff-format=udiff --dry-run",
+            "php-cs-fixer fix --diff --diff-format=udiff --dry-run --config=.php_cs.tests.php"
+        ],
+        "fix-style": [
```
mfn, line 74: Primary composer commands unchanged, so transparently will work for those used to them

src/Console/GeneratorCommand.php
```diff
@@ -59,7 +58,8 @@ class GeneratorCommand extends Command
      * @param \Illuminate\View\Factory $view
      */
     public function __construct(
-        /*ConfigRepository */ $config,
+        /*ConfigRepository */
+        $config,
```
mfn, line 62: Strange formatting, but so is the comment => should be moved to phpdoc anyway

tests/Console/EloquentCommandTest.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
```
mfn, line 3: 💪

tests/Console/ModelsCommand/CustomCollection/Test.php
```diff
@@ -6,9 +6,6 @@
 
 use Barryvdh\LaravelIdeHelper\Console\ModelsCommand;
 use Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\AbstractModelsCommand;
-use Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\CustomCollection\Collections\SimpleCollection;
-use Illuminate\Filesystem\Filesystem;
-use Mockery;
```
mfn, line 11: Lots of them were leftovers from https://github.com/barryvdh/laravel-ide-helper/pull/1019

tests/Console/ModelsCommand/GeneratePhpdocWithFqn/Models/Post.php
```diff
@@ -6,20 +6,30 @@
...
-use Illuminate\Database\Eloquent\{
-    Builder as EloquentBuilder,
-    Collection,
-    SoftDeletes
-};
+use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
+use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\HasMany;
+use Illuminate\Database\Eloquent\SoftDeletes;
```
mfn, line 13: No grouping, let's be consistent with everything else in the codebase

tests/Console/ModelsCommand/GeneratePhpdocWithFqn/Models/Post.php
```diff
@@ -6,20 +6,30 @@
...
 use Illuminate\Support\Carbon;
 
 class Post extends Model
 {
     use SoftDeletes;
 
+    /**
+     * Special hack to avoid code style fixer removing unused imports
+     * which play a role when generating the snapshot
+     */
```
mfn, line 24: Not the proudest thing to to but IMHO acceptable for a single exception here, as the imported (and not-yet-used use statements) are (only) relevant in this test

tests/Console/ModelsCommand/PHPStormNoInspection/Test.php
```diff
@@ -6,10 +6,6 @@
 
 use Barryvdh\LaravelIdeHelper\Console\ModelsCommand;
 use Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\AbstractModelsCommand;
-use Illuminate\Filesystem\Filesystem;
-use Mockery;
-
-use function file_get_contents;
```
mfn, line 12: 💥

mfn (commented):
Checked all changes, LGTM and especially _feels_ better in the name of consistency.

barryvdh (approved):

barryvdh:
LGTM, but already a conflict ;) Can you fix that so we can merge?

GrahamCampbell:
Could just use StyleCI's psr12 preset?

mfn:
StyleCI is great. So is php-cs-fixer. I happen to prefer the latter; thank you!

PR #1032, opened 2020-09-01
Run `composer fix-style` and `vendor/bin/phpunit -d --update-snapshots` to solve failed build

mfn:
This fixes https://github.com/barryvdh/laravel-ide-helper/runs/1055498817 from master


mfn:
The workflow which commits cannot push to master unless the action/checkout uses a PAT.

But unless this becomes a regular problem, no need to change anything there yet IMHO

barryvdh (approved):

PR #1036, opened 2020-09-04
Remove GrumPHP

mfn:
I would like to respectfully propose the removal of GrumPHP because it automatically activates git hooks

I think that in the context of OSS there is no need for this diligence as we've already proper tools in-place shielding us from all these things via CI anyway. Since we recently also added php-cs-fixer auto-commit, we truly cover everything as automatic as possible.

I understand it's sensitive topic, one which I already scratched in the past and the back-then removal was denied. I will absolutely accept this outcome too and would not try further attempts, but I would like to bring forth some arguments:
- CI already covers everything
- I find the automatic installation and activation of this process offensive
  I don't want to use tools which do this without my clear consent
- It "gets in the way" insofar I might decide to locally commit broken stuff to revisit it later, and grumphp prevents this.
  What and how I work locally is my own decision and no tool should interfere with that.

I understand there are multiple ways to still have grumphp but have the hooks not execute:
- use git own `--no-verify` or the IDE equivalent
- globally disable hooks with workarounds like https://stackoverflow.com/a/56436456/47573

But you first have to run into this to understand to opt out, on whatever level (repo, globally) that is, which I consider not a good DX.

Thank you!


barryvdh:
I guess now we fix it automatically this is fine. Thanks

PR #1038, opened 2020-09-06
[TESTS] Add basic test for Generator

mfn:
Gotta start somewhere 🤷‍♀️

I've no idea how useful _that_ test really is. We might remove it if it's breaking too often because it surely depends on Laravel.

I didn't test yet, but I imagine the snapshots will be different for L6/7/8… let's see what the GHA actions say!

### DIFF
--- a/tests/Generator/GeneratorTest.php
+++ b/tests/Generator/GeneratorTest.php
@@ -1,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Barryvdh\LaravelIdeHelper\Tests\Generator;
+
+use Barryvdh\LaravelIdeHelper\Generator;
+use Barryvdh\LaravelIdeHelper\IdeHelperServiceProvider;
+use Barryvdh\LaravelIdeHelper\Tests\TestCase;
+use Illuminate\Contracts\Config\Repository;
+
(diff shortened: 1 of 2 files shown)


tests/Generator/GeneratorTest.php
```diff
@@ -0,0 +1,47 @@
...
+            realpath(__DIR__ . '/../../resources/views'),
+        ]);
+    }
+
+    public function testGenerator(): void
+    {
+        // Feel free to bump this with a major Laravel release!
+        if ((int) Application::VERSION !== '7.27.0') {
+            self::markTestSkipped('This test only works with Laravel 7.27.0, feel free to bump it and update the snapshot!');
+        }
```
mfn, line 36: I first thought this could be a quick-win, but turns out the generated is (obviously) so specific to any given version, it may even be useless that way.

Should we still merge this for now and improve later?

If within a Laravel release cycle the generated is _relatively_ stable, and still could be useful for anyone changing behaviour of the `Generator`

mfn:
Nah, this approach is useless. I just made a accidental bug in the code actually and they were not even run locally anymore.

This is too brittle, need to come up with something more robust (and maybe just less "featureful").

PR #1039, opened 2020-09-06
[TESTS] Running in CI should not create but rather fail on missing snapshots

mfn:
## Summary

Otherwise they get created on the CI infrastructure but that's it, the
build won't fail although it was forgotten to add the snapshot.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -45,4 +45,4 @@ jobs:
         composer update --${{ matrix.dependency-version }} --prefer-dist --no-suggest --no-progress
 
     - name: Execute Unit Tests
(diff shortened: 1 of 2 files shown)


barryvdh:
I made the phpunit version more explicit. Looks good otherwise

barryvdh (approved):

mfn:
> I made the phpunit version more explicit. Looks good otherwise

Btw, I decided not to do that because the phpunit version is effectively dicdated by orchestra/testbench and the only goal was to make sure we don't use known broken version for our tests.

Otherwise once we bump orchestra which supports phpunit 10 we might stay on 9 for no reason but this constraint.

Issue #1042, opened 2020-09-08
Error: Class 'Illuminate\Foundation\Application' not found in file (When using Lumen)


pataar:
### Versions:
- ide-helper Version: 2.8.1
- Lumen Version: 7.2.1 
- PHP Version: 7.2.24

### Description:

It seems that a change broke the integration with Lumen. IlluminateFoundation\Application is not available there.

DannyvdSluijs:
I guess for anyone trying to reproduce this some additional info can be very helpful. Perhaps you have a stack trace or could you share your composer.lock and composer.json?

pataar:
Due to company restrictions I can't share those files fully. But I can provide more information when needed

mfn:
I stacktrace would be helpful but I can imagine a simple test with lumen and installing it will already expose it (don't have the time right now though)

pataar:
1. `composer create-project --prefer-dist laravel/lumen sample && cd sample`
2. `composer require --dev barryvdh/laravel-ide-helper`
3. `composer require league/flysystem #otherwise it won't run`
4.  Follow the Lumen guidelines from `README.md`
5. `php artisan ide-helper:generate`

6:
![image](https://user-images.githubusercontent.com/3403851/92494178-aeca5a80-f1f5-11ea-89be-0a5df90dd86a.png)

mfn:
Thanks!

The Laravel 8 comaptibility PR https://github.com/barryvdh/laravel-ide-helper/pull/1026 broke Lumen 😢

mfn:
@pataar can you give https://github.com/barryvdh/laravel-ide-helper/pull/1043 a try?

Applying the source change locally and see if generate works should be sufficient.

Thanks

pataar:
Yes, I'll let you know tomorrow

pataar:
Thanks in advance for the fix!

pataar:
@mfn I tested #1043 and the issue seems to be resolved after applying the fix.

PR #1050, opened 2020-09-09
Add integration test workflow

pataar:
## Summary
Introducing a new workflow which sets up a basic Laravel / Lumen project and injects the ide-helper. And makes sure that it can do a successful `generate` call.

@mfn I was wondering if you could review this and decide if it's useful enough.

Lumen tests are failing because of #1042 right now.

I'm not sure about the name, and the way of overwriting the source.

## Type of change

### Checklist


mfn:
> decide if it's useful enough

I can already say that I think this is _super useful_ 👍 

Not _necessarily_ for Laravel, because `orchestra/testbench` I think is basically sufficient, but for Lumen **definitely**.

I did something similar in
- https://github.com/rebing/graphql-laravel/blob/19ea3ed120b96c8c9b45398d0fb1ab1514860a4d/tests/integration-laravel.sh
- https://github.com/rebing/graphql-laravel/blob/19ea3ed120b96c8c9b45398d0fb1ab1514860a4d/tests/integration-lumen.sh

I'll give it a spin later!

.github/workflows/run-fresh.yml
```diff
@@ -0,0 +1,80 @@
...
+              run: cd sample && php artisan ide-helper:generate
+    php-laravel-fresh-run:
+        runs-on: ubuntu-latest
+        timeout-minutes: 15
+        env:
+            COMPOSER_NO_INTERACTION: 1
+        strategy:
+            matrix:
+                php: [7.4, 7.3, 7.2]
+                laravel: [8.*, 7.*, 6.*]
```
mfn, line 57: This needs an exclusion because PHP 7.2 and L8 don't work together

.github/workflows/run-fresh.yml
```diff
@@ -0,0 +1,80 @@
...
+                  path: src
+            - name: Setup PHP
+              uses: shivammathur/setup-php@v2
+              with:
+                  php-version: ${{ matrix.php }}
+                  coverage: none
+                  tools: composer:v2
+            - name: Install dependencies
+              run: |
+                  composer create-project --prefer-dist laravel/laravel:${{ matrix.lumen }} --no-progress sample && cd sample
```
mfn, line 72: Typo: it's not installing the version from the matrix
```suggestion
                  composer create-project --prefer-dist laravel/laravel:${{ matrix.laravel }} --no-progress sample && cd sample
```

.github/workflows/run-fresh.yml
```diff
@@ -0,0 +1,80 @@
...
+            - name: Install dependencies
+              run: |
+                  composer create-project --prefer-dist laravel/lumen:${{ matrix.lumen }} --no-progress sample && cd sample
+                  composer require "barryvdh/laravel-ide-helper" --dev --no-update --no-progress
+                  composer require "league/flysystem" --no-update --no-progress
+                  composer update --${{ matrix.dependency-version }} --prefer-dist --no-progress
+            - name: Overwrite source
+              run: mkdir -p sample/vendor/barryvdh/laravel-ide-helper && cp -a src/. sample/vendor/barryvdh/laravel-ide-helper/ && ls sample/vendor/barryvdh/laravel-ide-helper
+            - name: Dump autoloader
+              run: cd sample && composer dump-autoload
```
mfn, line 44: Instead of inventing steps pretending to simulate to work like composer, I suggest to take a look at https://github.com/rebing/graphql-laravel/blob/19ea3ed120b96c8c9b45398d0fb1ab1514860a4d/tests/integration-lumen.sh#L14-L16

This way composer itself is used and no need to copy around files and dump-autoload manually

.github/workflows/run-fresh.yml
```diff
@@ -0,0 +1,80 @@
...
+            - name: Insert service provider
+              run: echo '$app->register(\\Barryvdh\\LaravelIdeHelper\\IdeHelperServiceProvider::class);' > /tmp/sp && sed -i "97i $(</tmp/sp)" ./sample/bootstrap/app.php
```
mfn, line 46: I'm not saying this approach is wrong just because https://github.com/rebing/graphql-laravel/blob/19ea3ed120b96c8c9b45398d0fb1ab1514860a4d/tests/integration-lumen.sh#L21 does it different _but_ using sed like this with `97i` I find it's less intuitive in case if the file `bootstrap/app.php` changed.

With the other approach, replacing an existing matching strings, I think it's easier to understand the intention.

Maybe this can be improved even further with the _replacement_ approach, in that it aborts if the string was not replaced, meaning the file changed substantially and it's easier to figure this out?

Same applies to the Laravel below

.github/workflows/run-fresh.yml
```diff
@@ -0,0 +1,80 @@
...
+            - name: Execute generate run
+              run: cd sample && php artisan ide-helper:generate
```
mfn, line 48: We can come up with some kind of pro-active assertion here?

Like ensuring that a) the generated file exists and b) it contains something we expect to be there?

This could also pave the way to motivate adding other commands and their expectancy.

Applies to laravel below, too

mfn: Also for example checking/ensuring that storage/logs/* is empty

.github/workflows/run-fresh.yml
```diff
@@ -0,0 +1,80 @@
+name: Fresh
```
mfn, line 1: Maybe?
```suggestion
name: Integration Tests
```
And also change the filename to `run-integration-tests.yml`? would fit nicely into the existing ones:
image

.github/workflows/run-fresh.yml
```diff
@@ -0,0 +1,80 @@
...
+              with:
+                  php-version: ${{ matrix.php }}
+                  coverage: none
+                  tools: composer:v2
+            - name: Install dependencies
+              run: |
+                  composer create-project --prefer-dist laravel/lumen:${{ matrix.lumen }} --no-progress sample && cd sample
+                  composer require "barryvdh/laravel-ide-helper" --dev --no-update --no-progress
+                  composer require "league/flysystem" --no-update --no-progress
+                  composer update --${{ matrix.dependency-version }} --prefer-dist --no-progress
```
mfn, line 40: Where is `dependency-version` defined? 🤔

mfn (changes requested):
@pataar as already mentioned, I think this is a great and very smart idea, kudos 🙏 

Apologies for the many comments, but I believe there's value in taking the experience from existing projects applying them too.

This isn't the first project coming up with this, I remember I was inspired how https://github.com/nunomaduro/larastan/blob/32ab0115ae8b0ba185e04e9af503ad874a9f8a2a/tests/laravel-test.sh did it and was able to improve this even further and slightly give back, and I think we can also benefit from this here.

What do you think about the feedback, is this fine for you?

Thank you!

.github/workflows/run-fresh.yml
pataar: I totally agree

.github/workflows/run-fresh.yml
pataar: Whoops, set that to `--prefer-stable` now. Unless you prefer both `--prefer-lowest` and `--prefer-stable`

pataar:
Thanks for the feedback, that was very useful. Fixed all your points!

barryvdh (approved):

mfn (approved):
Very cool, thanks! 👍

pataar:
Thanks! 🙌

PR #1059, opened 2020-09-17
allow model_locations to have glob patterns

### LINKED ISSUE #1058: Config says glob patterns are supported, but they don't seem to be
### Versions:
- ide-helper Version: v2.8.1
- Laravel Version: v8.4.0
- PHP Version: 7.4.9

### Description:

According to the comment blocks in the config file, the model_locations configuration item should support glob patterns.  Right now it doesn't look like it does, even when I uses the given example in the comment.

### Steps To Reproduce:

- put a model into a subdirectory and then try and add it to the model_locations using glob patterns like this:

```
app/Services/*/Models
```
Instead of working, it will simply say "cannot locate directory ... "

If I look at the code, it seems that it is setup this way and I don't see how it could support glob patterns because it checks if it is a valid directory before using glob().  It fails on line 283 of ModelsCommand.. so it doesn't make it to the glob command on 288.

I looked into it and it's a simple fix so I'll try and send a PR.

isaackearl:
This if statement check for a valid directory was happening to early.  It was checking the string that contained the wildcards to see if it is a valid directory and always failing.  Need to check after the foreach starts. fixes #1058 .

## Summary
Glob patterns currently don't work.  The if statement that checks if the path is a directory happens before the glob helper call.

## Type of change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 - Compatibility with Lumen [\#1043 / mfn](https://github.com/barryvdh/laravel-ide-helper/pull/1043)
+- Allow model_locations to have glob patterns [\#1059 / saackearl](https://github.com/barryvdh/laravel-ide-helper/pull/1059)
 
 2020-09-07, 2.8.1
 -----------------
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -305,13 +305,14 @@ class ModelsCommand extends Command
                 $dir = base_path($dir);
             }
 
-            if (!is_dir($dir)) {
-                $this->error("Cannot locate directory '{'$dir}'");
-                continue;
-            }
-
             $dirs = glob($dir, GLOB_ONLYDIR);
             foreach ($dirs as $dir) {
+
+                if (!is_dir($dir)) {
+                    $this->error("Cannot locate directory '{'$dir}'");
+                    continue;
+                }
+
                 if (file_exists($dir)) {
                     $classMap = ClassMapGenerator::createMap($dir);
 
--- a/tests/Console/ModelsCommand/AllowGlobDirectory/Services/Post/Models/Post.php
+++ b/tests/Console/ModelsCommand/AllowGlobDirectory/Services/Post/Models/Post.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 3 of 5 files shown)


mfn (approved):
Thanks, LGTM!

---
Would you be up for a test for this?

isaackearl:
Thanks @mfn .. I think I could be up for a test tomorrow or the next day.  To be honest I've been on a rampage tonight trying to get some things working and it's now 5 AM.  :)

mfn:
> it's now 5 AM. :)

💥 

But yes, I've heard of that. Humans need 💤 and such 😏

mfn:
Github Actions seems to hang, but nothing on https://www.githubstatus.com/

isaackearl:
@mfn is there a way to force the checks to re-run?  Seems strange that they hung.

isaackearl:
@mfn I wrote a test, but may need some feedback about where I put it etc.  I tried to follow the pattern of the other tests.

isaackearl:
@mfn So I'll probably need some help finishing this test.  It works for me locally when I run using various php versions, however when I run it included with all the rest of the tests, it fails asserting that 'Written new phpDocBlock to' is there.  In other words: run the test by itself 👍 ... run the test with all the rest of them 👎 .

mfn:
Sure, I'll take a look over the weekend!

I'll try and see if I can re-start the tests by closing/opening the PR.

mfn:
> . It works for me locally when I run using various php versions, however when I run it included with all the rest of the tests

I don't know how it even worked for you, as the test neither in isolation nor together with the whole suite worked for me, simply because the "current working directory" as not the one expected.

I pushed a fix (and updated the changelog), all green now.

Are you using windows? Locally I use OSX and the GHA is using Linux, and they behaved the same.

isaackearl:
@mfn 
I'm on Mac os x as well.  Now that I'm seeing your changes I have no idea how this is STILL working for me locally when I run the test individually.  I am running my tests through php storm.  
Screen Shot 2020-09-19 at 3 28 26 PM

I'm on php 7.4.9 installed via homebrew.  

Regardless, I appreciate you updating it and making it truly work.

mfn (approved):
ok, great and thanks; also using phpstorm here 😏

Just wanted to make sure there's no bug hidden, but PR LGTM too.

@barryvdh IMHO this is good to merge!

NickSdot:
Tested. Works here as expected. Thanks!

mfn:
@NickSdot thanks for the feedback 👍

PR #1067, opened 2020-09-23
Fix --no-dev installs

HenkPoley:
## Summary

When barryvdh/laravel-ide-helper is a dev dependency, it is not available with `--no-dev` installs ("release builds"). Which will give errors if you add it as `post-update-cmd` or `post-install-cmd` in `composer.json`.

This fixes that by making an intermediate that checks for `APP_ENV="local"` in `.env`.

## Type of change

### Checklist
- [x] Update the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,45 @@ You can use an in-memory SQLite driver by adding the `-M` option.
 You can choose to include helper files. This is not enabled by default, but you can override it with the `--helpers (-H)` option.
 The `Illuminate/Support/helpers.php` is already set up, but you can add/remove your own files in the config file.
 
+#### Cooperation with release builds
+Create a file `app/Console/Commands/IdeHelperDoIt.php`:
+```php
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\Artisan;
+
+class IdeHelperDoIt extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'ide-helper:do-it';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Run barryvdh/laravel-ide-helper';
+
+    public function handle()
+    {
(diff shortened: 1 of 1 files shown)


HenkPoley:
It does not check if the database is available.

Somebody may want to implement something along the lines of:
```php
        try {
            DB::connection()->getPdo();
        } catch (Exception $e) {
            // If we get here we are not connected
        }
```

Our setup connects to multiple databases, and as dev you'd rather see it break than become silent.

HenkPoley:
Potentially ide-helper could drop such a file? I believe there are recommended ways to 'install files' in Laravel, e.g. to install  `config/<..>.php`

README.md
```diff
@@ -110,6 +110,47 @@ You can use an in-memory SQLite driver by adding the `-M` option.
...
+        if ('local' === config('app.env', 'production')) {
+            Artisan::call('ide-helper:generate');
+            Artisan::call('ide-helper:meta');
+            Artisan::call('ide-helper:models', ['--nowrite' => true]);
+        }
+    }
+}
+```
+
+In `composer.json` you can then add `"@php artisan ide-helper:do-it"` that will gracefully work indepent of `--no-dev` composer installs.
```
mfn, line 150: Maybe it should be more explicit _where_ to add this, as they're multiple keys possible in `scripts`, i.e. mention `post-update-cmd` ?

mfn (commented):
No objections from my side; having it doc only seems useful

Let's see what @barryvdh thinks!

HenkPoley:
I see `class_exists()` could also be a nice test.

netpok:
This should check whether ide-helper is installed:
```php
Artisan::command('ide-helper:run', function () {
    if (App::has('command.ide-helper.generate')) {
        Artisan::call('ide-helper:meta', [], $this->output);
        Artisan::call('ide-helper:generate', [], $this->output);
    }
})->describe('Run ide-helper:generate and meta.');
```

This can be inserted to routes/console.php

argonzalez2016 (approved):

PR #1071, opened 2020-09-28
There is no need to generate comments for query, newquery, newmodelqu…

a443939412:
There is no need to generate static method comments for query, newQuery and newModelQuery

## Summary
Model::newQuery() and Model::newModelQuery()  works incorrectly in laravel 6.8.

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -551,14 +551,14 @@ class ModelsCommand extends Command
                         );
                         $this->setMethod($name, $builder . '|' . $modelName, $args);
                     }
-                } elseif (in_array($method, ['query', 'newQuery', 'newModelQuery'])) {
+                }/* elseif (in_array($method, ['query', 'newQuery', 'newModelQuery'])) {
                     $builder = $this->getClassNameInDestinationFile($model, get_class($model->newModelQuery()));
 
                     $this->setMethod(
                         $method,
                         $builder . '|' . $this->getClassNameInDestinationFile($model, get_class($model))
                     );
-                } elseif (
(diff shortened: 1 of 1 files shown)


mfn:
I'm sorry, I don't understand the purpose of this PR.

It's certainly possible to invoke `<YourModel>::…` with any of these methods; they're caught by `\Illuminate\Database\Eloquent\Model::__callStatic` and forwarded.

Can you explain in more detail what it fixes for your? Also just commenting out code and not adapting tests won't be acceptable.

I'm inclined to reject this PR in the current state without more information, I hope you can understand.

a443939412:
Because it works wrong in laravel 6. X
`dd(Spu::newQuery());`
image

The same is true for the newModelQuery method. (laravel v5.5)
`dd(StockLogs::newModelQuery());`
image

rimace:
@mfn 
> It's certainly possible to invoke `<YourModel>::…` with any of these methods; they're caught by `\Illuminate\Database\Eloquent\Model::__callStatic` and forwarded.

I see what you mean, but PHP does not call `__callStatic` if the a method with that name exists.
So No, static call to `newQuery()` or `newModelQuery()` will not get forwarded.
When I try to call \<A Model\>::newQuery() PHP 7.4 and 8.0 throws the error `Non-static method Illuminate\Database\Eloquent\Model::newQuery() cannot be called statically` .

It only works with `query()`, because it is a static method.

a443939412:
Conflicts has been resolved

mfn:
Please add tests covering this, right now I don't even need to trigger the workflow runs because there are none. Thank you!

PR #1076, opened 2020-10-07
composer.json: remove --prefer-lowest

mfn:
## Summary
IMHO it's impractical to expect this package to have to work with the
"first release of Laravel for any major release" even when there are
already sever more releases: no one is/should stick to that version.

See also  https://github.com/barryvdh/laravel-ide-helper/pull/1074#pullrequestreview-504182121

## Type of change

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -21,12 +21,11 @@ jobs:
       matrix:
         php: [7.4, 7.3, 7.2]
         laravel: [8.*, 7.*, 6.*]
-        dependency-version: [prefer-lowest, prefer-stable]
         exclude:
           - laravel: 8.*
             php: 7.2
 
-    name: P${{ matrix.php }} - L${{ matrix.laravel }} - ${{ matrix.dependency-version }}
+    name: P${{ matrix.php }} - L${{ matrix.laravel }}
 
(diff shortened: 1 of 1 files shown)


barryvdh:
My thought was that if it's not compatible, we should bump the minor version until it is. Usually won't be a problem, but of 7.3 is not compatible for example, but 7.4, why not just require that version as minimum? (Eg 6.6|7.4|8.0)

mfn:
@barryvdh That would be possible, but for https://github.com/barryvdh/laravel-ide-helper/pull/1074#pullrequestreview-504182121 I don't even know at which version within Laravel 8 it starts working 🤷‍♀️

I just can't imagine this prefer-lowest being useful. Try open a bug report with Laravel and reporting a version which is not the latest release on any major version => first thing they say it you've to upgrade.

So I guess my question is: what's the real benefit?

(I know the general benefit, I realize enough projects use this approach, but doesn't make it necessary to follow it everywhere)

mfn:
OP from https://github.com/barryvdh/laravel-ide-helper/pull/1074#issuecomment-706474118 found a workaround, so maybe indeed we don't need this (personally I still think prefer lowest doesn't make sense, but it's not urgent) => closing

PR #1087, opened 2020-11-03
Test against php8

sergiy-petrov:


mfn:
@sergiy-petrov tests are failing

sergiy-petrov:
> @sergiy-petrov tests are failing

Yeah because php cs fixer is not compatible with php8 yet :(

mfn:
> Yeah because php cs fixer is not compatible with php8 yet :(

Ah, ok 😅

Well, no problem, I've seen activity on their repo regarding PHP8 so we just wait.

Thank you for clarifying 👍

mfn:
@sergiy-petrov

Since php-cs-fixer seems to take some more time, I moved ahead with the approach of simply removing it from the unit test matrix, as it isn't needed there anyway.

It turned out there was definitely a bit more to it, thus I'm replacing this PR with https://github.com/barryvdh/laravel-ide-helper/pull/1106

Thanks for your contribution!

PS: TBH I took a stab at PHP8 and totally forgot about your PR, only realized it afterwards

PR #1091, opened 2020-11-11
Reorganize names (Cleaner and more logical command names and generated file names)

fico7489:
- Cleaner and more logical command names and generated file names.

- Skip asking overwrite for models

- New commands are :

```
php artisan ide-helper:generate-facades
php artisan ide-helper:generate-models
php artisan ide-helper:generate-phpstorm
```

new files are :

```
.ide_helper/.facades.meta.php
.ide_helper/.models.meta.php
.ide_helper/.phpstorm.meta.php
```

### DIFF
--- a/.github/workflows/run-integration-tests.yml
+++ b/.github/workflows/run-integration-tests.yml
@@ -40,6 +40,7 @@ jobs:
                   cd sample
                   composer require "league/flysystem" --no-update --no-progress
                   composer update  --prefer-stable --prefer-dist --no-progress
+                  mkdir .ide-helper
 
             - name: Add package from source
               run: |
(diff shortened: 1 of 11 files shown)


fico7489:
@mfn what are you think about that?

mfn:
I can relate to the reasoning.

This project has quite a history already and every resource on the internetz you find will, and probably still for some time to come, shows the "old" commands.

As such, these are breaking changes so we would need a new major version release. Ideally a transition period for the duration of a major version which features both and emits warnings with the old commands?

The project currently lacks regular contributors and the existing ones can hardly handle the current cases well, so my feelings regarding this change are "probably not right now" and would rather focus efforts somewhere else.

Ultimate this is a judgement call for @barryvdh

fico7489:
Ok, I will then use my own fork...

mfn:
Thank you for your understanding!

Issue #1079, opened 2020-10-13
Php 8 & doctrine/dbal 3.0


KangAdmin:
Hello,

Thank you for your work !

I know that's a little early, but do you have any plan to support PHP 8 in dev branch ? 
Your package need package doctrine/dbal ~2.3, 
But to work in PHP 8 doctrine/dbal need 3.0.x-dev

`

    - Conclusion: remove doctrine/dbal 3.0.x-dev

    - barryvdh/laravel-ide-helper 2.8.x-dev requires doctrine/dbal ~2.3 -> satisfiable by doctrine/dbal[2.10.x-dev, 2.11.x-dev, 2.3.x-dev, 2.4.x-dev, 2.5.x-dev, 2.6.x-dev, 2.7.x-dev, 2.8.x-dev, 2.9.x-dev].

    - barryvdh/laravel-ide-helper v2.8.0 requires doctrine/dbal ~2.3 -> satisfiable by doctrine/dbal[2.10.x-dev, 2.11.x-dev, 2.3.x-dev, 2.4.x-dev, 2.5.x-dev, 2.6.x-dev, 2.7.x-dev, 2.8.x-dev, 2.9.x-dev].

    - barryvdh/laravel-ide-helper v2.8.1 requires doctrine/dbal ~2.3 -> satisfiable by doctrine/dbal[2.10.x-dev, 2.11.x-dev, 2.3.x-dev, 2.4.x-dev, 2.5.x-dev, 2.6.x-dev, 2.7.x-dev, 2.8.x-dev, 2.9.x-dev].

`

Thank you !
Best regards

mfn:
Absolutely desirable :)

However I don't have a "production ready" working environment anywhere, so can't test anything properly.

elijahchancey:
Now that dbal 3 has been released, it would be great for laravel-ide-helper to support it.

barryvdh:
Opened PR here: https://github.com/barryvdh/laravel-ide-helper/pull/1094
If it passes, I guess it's good to merge.

imami:
Any updates on this issue?

mfn:
dbal 3.0 support clashes with our currently supported range of Laravel version; so the current plan is to make one more (last) 2.8 release and for 2.9, drop Laravel < 8 and enable dbal => https://github.com/barryvdh/laravel-ide-helper/pull/1114

PR #1094, opened 2020-11-17
Allow Doctrine 3

barryvdh:
Tests Doctrine 3 (https://github.com/barryvdh/laravel-ide-helper/issues/1079)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "ext-json": "*",
         "barryvdh/reflection-docblock": "^2.0.6",
         "composer/composer": "^1.6 || ^2.0@dev",
-        "doctrine/dbal": "~2.3",
+        "doctrine/dbal": "^2.3|^3",
         "illuminate/console": "^6 || ^7 || ^8",
         "illuminate/filesystem": "^6 || ^7 || ^8",
         "illuminate/support": "^6 || ^7 || ^8",
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "ext-json": "*",
         "barryvdh/reflection-docblock": "^2.0.6",
         "composer/composer": "^1.6 || ^2.0@dev",
-        "doctrine/dbal": "^2.3|^3",
+        "doctrine/dbal": "^2.3 || ^3",
         "illuminate/console": "^6 || ^7 || ^8",
         "illuminate/filesystem": "^6 || ^7 || ^8",
         "illuminate/support": "^6 || ^7 || ^8",


barryvdh:
Tests say no: 
> Exception: Class 'Doctrine\DBAL\Driver\PDOSqlite\Driver

https://github.com/doctrine/dbal/blob/3.1.x/UPGRADE.md#bc-break-pdo-based-driver-classes-are-moved-under-the-pdo-namespace
Not sure how feasible it is to support both versions. We can't switch to 3.x yet because it's only PHP7.3+

Might also look here: https://github.com/laravel/framework/pull/35236/files
Not sure if we use those classes directly, or rely on Laravel. In that case we can only support Doctrine 3 on the latest version of Laravel (when tagged)

mfn:
>  Not sure how feasible it is to support both versions. We can't switch to 3.x yet because it's only PHP7.3+

Haven't looked in detail, but maybe we can do:
- switch to the new class
- have a `class_alias` in-place in `setUp` method for all the integration tests, with a comment to remove it once dbal 2.x is dropped?

(sorry didn't fully check the other link, on the run currently)

mfn:
Ah I see, Laravel uses `class_exists(Version::class)` to make this decision which one

a-bashtannik:
Doctrine 3.0 is the first version with PHP 8.0 support, how could be Laravel IDE Helper PHP 8.0 compatible if it doesn't support Doctrine 3.0 yet?

mfn:
@barryvdh maybe a stupid question, but how can we even support dbal 2.* _and_ 3.* given that Laravel 6 and 7 don't support it?

barryvdh:
Ideally we support support doctrine 3 on laravel 8, and doctrine 2 on the rest.

barryvdh:
Maybe we should just finish existing PR's and then bump to Laravel 8 for the new version.

mfn:
> Maybe we should just finish existing PR's and then bump to Laravel 8 for the new version.

hah, was thinking about this too 😅

mfn:
Because, IDK where the problem exactly is, but issue we see in CI is that Laravel < 8 only compatible with dbal 2.* gets 3.* and the error is from within Laravel AFAICS.

And since dbal is not a `require` on Laravel (it's `require-dev` afaik), there's no constraint in place.

barryvdh:
Yeah I thought if it was required, Composer would just filter it out. But with just suggest, that's not possible. So let's just make sure this version is stable and then bump it after release. New features are mostly for Laravel 8 anyways.

barryvdh:
2.8 milestone is for PR's targeted for the current release (2.8.2), 2.9 milestone is for Laravel 8 version only.

mfn:
This also means we can drop PHP 7.2 in 2.9; this is actually a "good thing ™" as it's already EOL since a few days 👍

a-bashtannik:
Psalm output for Laravel 8.17 and `barryvdh/laravel-ide-helper: ^2.8.1`  (and PHP 8.0)

```
PHP Fatal error:  Declaration of Illuminate\Database\PDO\Concerns\ConnectsToDatabase::connect(array $params) must be compatible with Doctrine\DBAL\Driver::connect(array $params, $username = null, $password = null, array $driverOptions = []) in /project/vendor/laravel/framework/src/Illuminate/Database/PDO/MySqlDriver.php on line 16
```

The `connect` [method signature](https://github.com/doctrine/dbal/blob/2.12.x/lib/Doctrine/DBAL/Driver.php#L26) on Doctrine 2.* is 
```
public function connect(array $params, $username = null, $password = null, array $driverOptions = []);
```
And on Doctrine 3.* it [changed](https://github.com/doctrine/dbal/blob/3.0.x/src/Driver.php#L26) to 
```
public function connect(array $params);
```
The last update in [illuminate/database package brings changes](https://github.com/illuminate/database/commit/b6d6e8346cdf09200b5a0886684c8bb962922870) that *should* broke backward compatibility for all packages using Doctrine 2.x - they kept `^2.6|^3.0` support - but probably didn't test it on 2.6. 

The `illuminate/database` package [checks for existence](https://github.com/illuminate/database/commit/b6d6e8346cdf09200b5a0886684c8bb962922870#diff-73e390900355a314da4b6c207c413a655f438c3c9a2197ea1d1f6c504ce834ecR91) `Version` class (that exists in Doctrine 2.x and doesn't in 3.x) and (in our case) tries to use Doctrine 3.x `Driver` interface implementation incompatible with 2.x. We remember, the `illuminate/database` [connect](https://github.com/illuminate/database/commit/b6d6e8346cdf09200b5a0886684c8bb962922870#diff-c0842c2b4954e2c4dbd77fa4bc89b74e0e49d6a07e82bfe061545b53be1ca3baR16) method has Doctrine 3.x compatible signature, not Doctrine 2.x. 

Without Doctrine 2.x vendor package installed (`Version` class is missing), everything works fine because `illuminate/database` switches to [built-in PDO Driver classes](https://github.com/illuminate/database/tree/8.x/PDO)  - looks like Doctrine 3.x will never be used under the hood by `illuminate/database` and that's why the error appears only if 3-rd party package like Laravel IDE Helper installs 2.x version of Doctrine DBAL.

mfn:
@a-bashtannik great analysis 🙏 

Do you think there's a way for supporting 2.* + 3.* somehow or do you also think, moving forward is better (drop < Laravel 8 support, support dbal 3.* only)?

We also have to take into considerations the many (joking a bit here 😅) forces contributing to this project, so for me it's a rather "easy way out" to simply move forward.

mfn:
Here's my draft how to approach this https://github.com/barryvdh/laravel-ide-helper/pull/1114

Will finalize once we made the last 2.8 release

Mexidense:
Any news? :heart:

mfn:
This was superseded by https://github.com/barryvdh/laravel-ide-helper/pull/1114 and released as https://github.com/barryvdh/laravel-ide-helper/releases/v2.9.0

PR #1096, opened 2020-11-20
replace switch case with assoc array

alirezavalipour:
## Summary
replace switch case with assoc array

## Type of change
- [ ]  replace switch case  with assoc array, because it's less code lines and it's more understandable


mfn:
👎 on the whole change, I don't think it improves anything and actually it makes the first `default` branch harder to read and to maintain, should the logic need to be improved:
image

The last line looks so forced without much gain.

PR #1103, opened 2020-12-02
Allow casts with a return type of static or this to reference themselves

hailwood:
Right now a cast cannot specify that it returns static or this.
Allowing so would mean a new cast can extend the old one without having to redefine the get method (e.g. an Enum)
Presently if you specify static/this in the return type then those values are literally used which makes a property look like it returns an instance of the model itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file.
 - Compatibility with Lumen [\#1043 / mfn](https://github.com/barryvdh/laravel-ide-helper/pull/1043)
 - Allow model_locations to have glob patterns [\#1059 / saackearl](https://github.com/barryvdh/laravel-ide-helper/pull/1059)
 - Error when generating helper for macroable classes which are not facades and contain a "fake" method [\#1066 / domkrm] (https://github.com/barryvdh/laravel-ide-helper/pull/1066)
+- Casts with a return type of `static` or `$this` now resolve to an instance of the cast [\#1103 / riesjart](https://github.com/barryvdh/laravel-ide-helper/pull/1103)
 
 2020-09-07, 2.8.1
 -----------------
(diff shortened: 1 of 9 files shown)


mfn:
Can you please add a test showing the behavour change? Thank you!

hailwood:
Thanks for suggesting the tests @mfn, it made me realise a bug in my code :)

I've added hopefully thorough tests for all the different cases.
I also fixed an unrelated issue in another test due to a deprecated phpunit method being used.

tests/Console/EloquentCommandTest.php
```diff
@@ -45,14 +45,25 @@ public function testCommand()
...
+        } else {
+            $this->assertRegExp(
+                ';Unexpected no document on Illuminate\\\Database\\\Eloquent\\\Model;',
+                $display
+            );
+            $this->assertRegExp(
+                ';Wrote expected docblock to .*/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php;',
+                $display
+            );
+        }
```
mfn, line 66: Please revert these changes; assertRegExp works fine; the warning may be annoying but IMHO no need for duping the code here.

mfn (changes requested):
PR LGTM!

However, can you please undo the unrelated phpunit changes as outlined?

Thank you!

hailwood:
@mfn All done :)

mfn:
@hailwood great!

sorry forgot, please add a changelog (also next time: please don't delete the PR template, it's there for a reason to be reminded of such things, thank you!)

mfn (approved):
Thank you 🤗 

@barryvdh LGTM for merge!

hailwood:
Thanks @mfn 
I normally keep the PR template, I think grammarly wiped it out sorry!

What would you put in the Changelog?
Thinking under Fixed 

- Casts with a return type of `static` or `$this` correctly resolve to an instance of the cast [\#1103 / riesjart](https://github.com/barryvdh/laravel-ide-helper/pull/1103)

mfn:
@hailwood yeah, sounds good to me!

PR #1106, opened 2020-12-03
[PHP8] Add initial compatibility

mfn:
## Summary
- the codebase seems compatible except the one place where a deprecated `ReflectionParameter->getClass()` was called
  - replaced it with `->getType()`
  - seems it also improve the type declarations which are generated (test `GeneratePhpdocWithFqn`)
- Lumen 6 and 7 don't support PHP8, so excluded from integration tests
- friendsofphp/php-cs-fixer is (not yet) PHP 8 compatible, thus removed it from the unit tests matrix
  Same was already done for vimeo/psalm in the past
- mockery minimum version had to be bumped to 1.3.3 for PHP8 compat
  figured the bump is fine as it's a dev dependency only anyway
- spatie/phpunit-snapshot-assertions required `^4` for PHP8 compat, so it's added
- in one test we checked the literal PHP errors, which changed from PHP < 8 and PHP 8, so added an `if` condition to `DynamicRelations/Test.php`

The biggest issue though is that doctrine/dbal with `prefer-lowest` in the test matrix is NOT compatible with PHP8, it throws some method declaration mismatch exception.
- Our minimum is 2.3 which also supports PHP 7.2
- the minimum for PHP8 compat is dbal 2.12.0, however starting with dbal 2.11.0 only PHP 7.3 is supported, clashing with our minimum supported version of 7.2 (for Laravel/Lumen 6)

=> therefore I decided to not bump doctrine/dbal in this PR and just don't run the unit test matrix on PHP8 with prefer-lowest until we've a better solution.

> The biggest issue though is that doctrine/dbal with prefer-lowest in the test matrix is NOT compatible with PHP8, it throws some method declaration mismatch exception.

I wonder though why this wouldn't be handled by composer dependency resolution correctly, IDK TBH.

### Links
- Fixes https://github.com/barryvdh/laravel-ide-helper/issues/1108

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -424,7 +424,7 @@ class ModelsCommand extends Command
 
         $database = null;
         if (strpos($table, '.')) {
-            list($database, $table) = explode('.', $table);
+            [$database, $table] = explode('.', $table);
         }
 
         $columns = $schema->listTableColumns($table, $database);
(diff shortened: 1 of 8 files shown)


src/Console/ModelsCommand.php
```diff
@@ -424,7 +424,7 @@ protected function getPropertiesFromTable($model)
 
         $database = null;
         if (strpos($table, '.')) {
-            list($database, $table) = explode('.', $table);
+            [$database, $table] = explode('.', $table);
```
mfn, line 427: Unrelated but Phpstorm just insists in changed it 💥

barryvdh (approved):
Nice!

Issue #1112, opened 2020-12-04
ModelsCommand: array_diff(): Expected parameter 1 to be an array, null given


mfn:
### Versions:
- ide-helper Version: dev-master 9859958
- Laravel Version: 8.16.1
- PHP Version: 7.1.12

### Description:
```
Exception: array_diff(): Expected parameter 1 to be an array, null given
Could not analyze class App\Models\SomeModel.

Trace:
#0 [internal function]: Illuminate\Foundation\Bootstrap\HandleExceptions->handleError(2, 'array_diff(): E...', '/vagrant/api.sw...', 1163, Array)
#1 vendor/barryvdh/laravel-ide-helper/src/Console/ModelsCommand.php(1163): array_diff(NULL, Array)
#2 vendor/barryvdh/laravel-ide-helper/src/Console/ModelsCommand.php(565): Barryvdh\LaravelIdeHelper\Console\ModelsCommand->writeModelExternalBuilderMethods('Builder', Object(App\Models\SomeModel))
#3 vendor/barryvdh/laravel-ide-helper/src/Console/ModelsCommand.php(277): Barryvdh\LaravelIdeHelper\Console\ModelsCommand->getPropertiesFromMethods(Object(App\Models\SomeModel))
#4 vendor/barryvdh/laravel-ide-helper/src/Console/ModelsCommand.php(159): Barryvdh\LaravelIdeHelper\Console\ModelsCommand->generateDocs(Array, Array)
#5 vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(36): Barryvdh\LaravelIdeHelper\Console\ModelsCommand->handle()
#6 vendor/laravel/framework/src/Illuminate/Container/Util.php(40): Illuminate\Container\BoundMethod::Illuminate\Container\{closure}()
#7 vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(93): Illuminate\Container\Util::unwrapIfClosure(Object(Closure))
#8 vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(37): Illuminate\Container\BoundMethod::callBoundMethod(Object(App\Services\Vendor\Illuminate\Foundation\Application), Array, Object(Closure))
#9 vendor/laravel/framework/src/Illuminate/Container/Container.php(610): Illuminate\Container\BoundMethod::call(Object(App\Services\Vendor\Illuminate\Foundation\Application), Array, Array, NULL)
[...]

PR #1089, opened 2020-11-07
Add support for dedicated eloquent builders and type/doc parameter support.

KaloyanYosifov:
#887 

## Summary
Added a new feature where we create a helper file with external eloquent builders [Reference](https://timacdonald.me/dedicated-eloquent-model-query-builders/). 

This is very useful as in my big projects I do use this feature and so does some people -- [Issue](https://github.com/barryvdh/laravel-ide-helper/issues/887).

## Type of change

- [ x ] New feature (non-breaking change which adds functionality)
- [ x ] This change requires a documentation update 
- [ x ] Misc. change (internal, infrastructure, maintenance, etc.)

### Checklist
- [ x ] Existing tests have been adapted and/or new tests have been added
- [ x ] Update the README.md
- [ x ] Code style has been fixed via `composer fix-style`
- [ x ] Tested

PR #1113, opened 2020-12-05
FQCN bugfix when writing external eloquent builder methods inside models.

KaloyanYosifov:
## Summary
There is a bug where when we are writing eloquent builder methods directly into the model, we pass only the short name of the builder instead of the class to the "writeModelExternalBuilderMethods". More details here -- https://github.com/barryvdh/laravel-ide-helper/issues/1112

### Links
- Fixes https://github.com/barryvdh/laravel-ide-helper/issues/1112

## Type of change
Updated the "writeModelExternalBuilderMethods" method, so we only receive the model when writing the external builder methods. This way we can get the full class directly in the method instead of accepting it as a parameter which was a bit risky. 

After we've diffed the base builder methods and the new one's, we get the short name or the full class name  of the builder (based on the user's settings) and write it to the doc

### Checklist


src/Console/ModelsCommand.php
```diff
@@ -492,7 +492,7 @@ protected function getPropertiesFromTable($model)
             if ($this->write_model_magic_where) {
                 $this->setMethod(
                     Str::camel('where_' . $name),
-                    $this->getClassNameInDestinationFile($model, \Illuminate\Database\Eloquent\Builder::class)
+                    $this->getClassNameInDestinationFile($model, get_class($model->newModelQuery()))
```
KaloyanYosifov, line 495: @mfn 

I checked that the new change to use the external builder for the `where` magic methods is not a big one. We just need to use the builder class returned from the (newModelQuery).

tests/Console/ModelsCommand/DoesNotGeneratePhpdocWithExternalEloquentBuilder/__snapshots__/Test__test__1.php
```diff
@@ -90,79 +90,79 @@
...
- * @method static \Illuminate\Database\Eloquent\Builder|Post whereUnsignedSmallIntegerNotNullable($value)
- * @method static \Illuminate\Database\Eloquent\Builder|Post whereUnsignedSmallIntegerNullable($value)
- * @method static \Illuminate\Database\Eloquent\Builder|Post whereUnsignedTinyIntegerNotNullable($value)
- * @method static \Illuminate\Database\Eloquent\Builder|Post whereUnsignedTinyIntegerNullable($value)
- * @method static \Illuminate\Database\Eloquent\Builder|Post whereUpdatedAt($value)
- * @method static \Illuminate\Database\Eloquent\Builder|Post whereUuidNotNullable($value)
- * @method static \Illuminate\Database\Eloquent\Builder|Post whereUuidNullable($value)
- * @method static \Illuminate\Database\Eloquent\Builder|Post whereYearNotNullable($value)
- * @method static \Illuminate\Database\Eloquent\Builder|Post whereYearNullable($value)
+ * @method static \Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\DoesNotGeneratePhpdocWithExternalEloquentBuilder\Builders\PostExternalQueryBuilder|Post whereBigIntegerNotNullable($value)
```
KaloyanYosifov, line 93: @mfn I do not know if it is the right thing. But perhaps if they deactivate the external builder option, we should not use the external builder in the where magic methods? Let me know what you think.

mfn:
@KaloyanYosifov completely unrelated, if you remove the spaces around the `x` in `[x]`, it will rendered nicely as checked checkboxes!
image

KaloyanYosifov:
@mfn 

Thanks, was wondering how that was done 😄

mfn:
Step 0 completed successfully: no errors/problems produced when running this in a private project 💪, will review the code now 👓

src/Console/ModelsCommand.php
mfn: Yes this seems appropriate. And as you said, usually I'm also a fan of not mixing fix but sometimes it's ok => 👍 
Thank you

src/Console/ModelsCommand.php
```diff
@@ -1149,26 +1149,31 @@ protected function getUsedClassNames(ReflectionClass $reflection): array
...
 
-    protected function writeModelExternalBuilderMethods(string $builder, Model $model): void
+    protected function writeModelExternalBuilderMethods(Model $model): void
     {
-        if (in_array($builder, ['\Illuminate\Database\Eloquent\Builder', 'EloquentBuilder'])) {
-            return;
-        }
-
-        $newBuilderMethods = get_class_methods($builder);
+        $fullBuilderClass = '\\' . get_class($model->newModelQuery());
```
mfn, line 1154: Ah, I see! 👍

tests/Console/ModelsCommand/DoesNotGeneratePhpdocWithExternalEloquentBuilder/__snapshots__/Test__test__1.php
mfn: @KaloyanYosifov thank for mentioning it, in fact I already wrote this during my review but re-arranged it now for context:

> Is this test here obsolete?
> 
> Because it says (the namespace, I separated the words): "Does Not Generate Phpdoc With External Eloquent Builder"
> 
> But OTOH I'm unsure, it looks all the same to me with the test in `GeneratePhpdocWithExternalEloquentBuilder`, would appreciate any insights.

So I guess, yes, it shouldn't?

mfn (commented):
This is great, I can follow the fix on this one ;)

But as you mentioned, not sure about this one test behaviour now?

Basically this is "approve-able", let's sort out the last thing here then it's good to merge for me.

Thank you for your quick response 👍

tests/Console/ModelsCommand/DoesNotGeneratePhpdocWithExternalEloquentBuilder/__snapshots__/Test__test__1.php
KaloyanYosifov: @mfn Thank you for your time on this review 🙏 . I think we are going to need this test as we have a new option that defines if we should either use the "external eloquent builder methods" feature or not. And since that option "write_model_external_builder_methods" is true by default, we are disabling it in this [Test](https://github.com/barryvdh/laravel-ide-helper/blob/master/tests/Console/ModelsCommand/DoesNotGeneratePhpdocWithExternalEloquentBuilder/Test.php#L16) to check if it is working correctly (not using the external builder).

KaloyanYosifov (commented):
@mfn Will update the code, so it doesn't use the external builder option in the "where" magic methods if the "write_model_external_builder_methods" is turned off. 

Let me know if I am missing something or is it just this little update?

mfn (approved):
@KaloyanYosifov this LGTM, thank you!

Again tested this in a private project, also good there

@barryvdh good to merge from my side: an important follow-up bugfix for https://github.com/barryvdh/laravel-ide-helper/pull/1089 which will close https://github.com/barryvdh/laravel-ide-helper/issues/1112 and I think we're good to roll a new release after that!

mfn:
@barryvdh awesome! Would be great if you could find time to make a new release in the next days, the changelog should "ready to copy paste" for the release itself.

Issue #1084, opened 2020-10-24
Call to a member function getMethods() on null


celorodovalho:
### Versions:
- ide-helper Version: dev-master (2.8.1)
- Laravel Version: 8.9.0
- PHP Version: 7.4.8

### Description:

I recently try to run the command below and I got this error:

```php
$ artisan ide-helper:generate                                             

   Error 

  Call to a member function getMethods() on null

  at vendor/barryvdh/laravel-ide-helper/src/Macro.php:64
     60▕     {
     61▕         $enclosingClass = $this->method->getClosureScopeClass();
     62▕ 
     63▕         /** @var \ReflectionMethod $enclosingMethod */
  ➜  64▕         $enclosingMethod = Collection::make($enclosingClass->getMethods())
     65▕             ->first(function (\ReflectionMethod $method) {
     66▕                 return $method->getStartLine() <= $this->method->getStartLine()
     67▕                     && $method->getEndLine() >= $this->method->getEndLine();
     68▕             });

      +32 vendor frames 
  33  artisan:37
      Illuminate\Foundation\Console\Kernel::handle()

```

### Steps To Reproduce:

- artisan ide-helper:generate

mfn:
Can you debug which method this is about?

Also maybe the code is simply buggy as `getClosureScopeClass` can return `null`, but nevertheless would be good to understand what specifically triggers this.

standaniels:
This seems to be caused by invokable classes. This how to reproduce this issue:

**Versions**:

- laravel/framework 8.18.1
- barryvdh/laravel-ide-helper 2.8.2
- PHP 7.4.10


## Step 1.

Create class `App\Foo`:

```php
<?php

namespace App;

class Foo
{
    public function __invoke($var)
    {
        return $var;
    }
}
```

## Step 2.

Register the macro by passing a new instance of `App\Foo`:

```php
<?php

namespace App\Providers;

use App\Foo;
// ...

class AppServiceProvider extends ServiceProvider
{
    public function register()
    {
        UrlGenerator::macro('foo', new Foo());
    }

    // ...
}
```

## Step 3.

Run `php artisan ide-helper:generate`.

## Result

```
   Error

  Call to a member function getMethods() on null

  at vendor/barryvdh/laravel-ide-helper/src/Macro.php:64
     60▕     {
     61▕         $enclosingClass = $this->method->getClosureScopeClass();
     62▕
     63▕         /** @var \ReflectionMethod $enclosingMethod */
[...]

PR #1123, opened 2020-12-10
bug-fix : Call to a member function getMethods() on null

imami:
## Summary
This pull request fixes the issue #1084 and
prevent the exception when enclosingClass is null.

## Type of change


mfn:
@imami can you also add a test, showing the effect of this change?

barryvdh:
I think this is already fixed with https://github.com/barryvdh/laravel-ide-helper/pull/1239

PR #1138, opened 2021-01-06
changelog: re-order entry after recent merges in master

mfn:
## Summary
As the title says. Just a minor thing to keep the changelog in order so it's correct for the next release and can be copy pasted.


mfn:
The style fix is unrelated, guess there was an update to php-cs-fixer or whatever => it's "good" IMHO.

barryvdh:
Hmm test is failing still

mfn:
> Hmm test is failing still

See https://github.com/barryvdh/laravel-ide-helper/pull/1139 , it's unrelated to changelog/php-cs-fixer but a result of the macro merge recently (probably had an outdated php-cs-fixer I guess

PR #1163, opened 2021-02-27
Auto generate models

netpok:
This pull request allows generating model helpers automatically after migrations executed.

Basically a newer version of #842, started from the current master branch.

Based on the comment by barryvdh the config option is simplified it is named `post_migrate` and it is either false to disable this feature or sets the parameters for the "ide-helper:model" command. It supports both `"--nowrite"` and the `["--nowrite" => true]` format.

The trigger was updated to use the MigrationsEnded event because Laravel will run "incomplete" commands like `migrate:roll` as `migrate:rollback`.

The other helper file generation was dropped in favor of composer post update commands.

### DIFF
--- a/src/IdeHelperServiceProvider.php
+++ b/src/IdeHelperServiceProvider.php
@@ -35,7 +35,7 @@ class IdeHelperServiceProvider extends ServiceProvider implements DeferrableProv
      */
     public function boot()
     {
-        if ($this->app['config']->get('ide-helper.post_migrate', false)) {
+        if ($this->app['config']->get('ide-helper.post_migrate', [])) {
             $this->app['events']->listen(CommandFinished::class, GenerateModelHelper::class);
             $this->app['events']->listen(MigrationsEnded::class, function () {
                 GenerateModelHelper::$shouldRun = true;
--- a/src/IdeHelperServiceProvider.php
+++ b/src/IdeHelperServiceProvider.php
@@ -15,7 +15,10 @@ use Barryvdh\LaravelIdeHelper\Console\EloquentCommand;
 use Barryvdh\LaravelIdeHelper\Console\GeneratorCommand;
 use Barryvdh\LaravelIdeHelper\Console\MetaCommand;
 use Barryvdh\LaravelIdeHelper\Console\ModelsCommand;
+use Barryvdh\LaravelIdeHelper\Listeners\GenerateModelHelper;
+use Illuminate\Console\Events\CommandFinished;
 use Illuminate\Contracts\Support\DeferrableProvider;
+use Illuminate\Database\Events\MigrationsEnded;
 use Illuminate\Foundation\Application;
 use Illuminate\Support\ServiceProvider;
 use Illuminate\View\Engines\EngineResolver;
@@ -32,6 +35,13 @@ class IdeHelperServiceProvider extends ServiceProvider implements DeferrableProv
      */
     public function boot()
     {
+        if ($this->app['config']->get('ide-helper.post_migrate', false)) {
+            $this->app['events']->listen(CommandFinished::class, GenerateModelHelper::class);
+            $this->app['events']->listen(MigrationsEnded::class, function () {
+                GenerateModelHelper::$shouldRun = true;
+            });
+        }
+
         if ($this->app->has('view')) {
             $viewPath = __DIR__ . '/../resources/views';
             $this->loadViewsFrom($viewPath, 'ide-helper');
--- a/src/Listeners/GenerateModelHelper.php
+++ b/src/Listeners/GenerateModelHelper.php
@@ -5,6 +5,7 @@ namespace Barryvdh\LaravelIdeHelper\Listeners;
 use Illuminate\Console\Events\CommandFinished;
 use Illuminate\Contracts\Config\Repository as Config;
 use Illuminate\Contracts\Console\Kernel as Artisan;
+use Illuminate\Support\Arr;
 
 class GenerateModelHelper
 {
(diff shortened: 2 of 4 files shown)


src/Listeners/GenerateModelHelper.php
```diff
@@ -0,0 +1,50 @@
...
+     *
+     * @param  CommandFinished  $event
+     */
+    public function handle(CommandFinished $event)
+    {
+        if (!self::$shouldRun) {
+            return;
+        }
+
+        self::$shouldRun = false;
```
mfn, line 41: Why is this necessary?

mfn (commented):
Generally LGTM, I've a question (inline) though, thanks

src/Listeners/GenerateModelHelper.php
netpok: To access the current output buffer we need to run the command on the `CommandFinished` event, where we can access the commands output.

To achieve this:
- We set the flag on the `MigrationsEnded` event
- If the flag is set we run the command
- But we need to reset the flag before that otherwise it would also trigger on the end of the generate command resulting in an infinite loop.

The best way to avoid this would be single fired events but as far as I know Laravel does not support them.

mfn: Makes sense to me!

Can you please add this as a phpdoc comment to the `$shouldRun` property? I doubt this will be easy to figure out :)

After that I think we're good here.

netpok: I added one

mfn (approved):
LGTM!

netpok:
I came across a project which uses the `spatie/laravel-multitenancy` package, thus running the basic `artisan ide-helper:generate` does not suffice.

Solutions in my mind:
1. Relaxing the constraint by requiring the full command to be entered, e.g: `ide-helper:models -N`, maybe even allowing multiple commands. This solution would mean dropping array parameter compatibility, but I don't see that would be used too much.
2. Allowing to use a custom action, e.g.: allowing to set a callback which defaults to the current implementation `GenerateModelHelper::generateWith(callback);` in AppServiceProvider
3. This is an edge case which the user should handle on it's own, e.g: by overriding the generator listener, which should be possible with the current code by using `$app->register(GenerateModelHelper::class, CustomImplementation::class);`

mfn:
> This solution would mean dropping array parameter compatibility, but I don't see that would be used too much.

At least I can say I agree to that assertion :)

I think I would say, the bottom line for all of this would be improved DX for users of this library. Seems only 1) really fulfills this, once you "have to code" I think it gets worse. Also creates more dependencies on the internal code, which we might rather change one day…

netpok:
I pushed a commit which implements option 1.

src/Listeners/GenerateModelHelper.php
```diff
@@ -45,11 +46,10 @@ public function handle(CommandFinished $event)
...
 
-        $parameters = $this->config->get('ide-helper.post_migrate');
-        $this->artisan->call(
-            is_array($parameters) ? 'ide-helper:models' : 'ide-helper:models ' . $parameters,
-            is_array($parameters) ? $parameters : [],
-            $event->output,
-        );
+        $commands = $this->config->get('ide-helper.post_migrate');
+
+        foreach (Arr::wrap($commands) as $command) {
```
mfn, line 51: The `Arr::wrap` I think will break when using the current `false`, can you check?

`Arr::wrap` has a special case for `null` coercing to an empty array (which would be fine here), but otherwise it wants to process the defaults as `[false]` when I read this correctly.

Maybe it would be just easier to make `post_migrate` always an array, also in the config, with an empty one but one commented example which could user help to quicker start using it:
```php
'post_migrate' => [
    // 'ide-helper:models --nowrite',
],
```

mfn: Further, please also add the default to `$this->config->get()`, not everyone upgrading the library immediately upgrades the config files (though incidentally, in _this_ case the `null` default would just work but it just would seem inconsistent then).

netpok: It should not matter since the [event listener is not registered when falsy values are provided](https://github.com/barryvdh/laravel-ide-helper/pull/1163/files#diff-08724b347a720a68357b78ce96d5c46a5fcb0c90bc3784594620a0af65e2bf45R38)

netpok: Simplifying the config makes sense so I did that. Also the package config is always merged with the user config so there shouldn't be an event where the config entry is missing.

src/IdeHelperServiceProvider.php
```diff
@@ -32,6 +35,13 @@ class IdeHelperServiceProvider extends ServiceProvider implements DeferrableProv
      */
     public function boot()
     {
+        if ($this->app['config']->get('ide-helper.post_migrate', false)) {
```
mfn, line 38: Thanks for the other change and the clarification! I think for consistency, we also should default to `[]` here now:
```suggestion
        if ($this->app['config']->get('ide-helper.post_migrate', [])) {
```

mfn (approved):
One last comment, otherwise LGTM!

src/Listeners/GenerateModelHelper.php
mfn: > Also the package config is always merged with the user config

That's interesting, we definitely had issues in the past due to missing config values and no defaults in the code 🤷‍♀️

src/IdeHelperServiceProvider.php
netpok: I thought leaving this as false since its a condition but ok, changed it.

src/Listeners/GenerateModelHelper.php
netpok: Interesting, this [line should](https://github.com/barryvdh/laravel-ide-helper/blob/master/src/IdeHelperServiceProvider.php#L57) prevent that.

mfn:
@barryvdh IMHO good to merge!

netpok:
@mfn sorry to bother you but could you fix my name in the changelog next time you edit it?

Issue #1166, opened 2021-03-01
Do not generate magic wheres for multi-word camelCase properties


amcsi:
Although Eloquent supports camelCase property names if you set `$snakeAttributes = false;` on your model, it's currently impossible for magic "wheres" to work for properties with more than one word in them, because the magic wheres happen in the Query Builder layer which has no knowledge about Eloquent.

So here is an example of ide-helper docblocks generated:

```php

/**
 * App\Person.
 *
 * @property int $id
 * @property string $name
 * @property \Illuminate\Support\Carbon|null $createdAt
 * @property \Illuminate\Support\Carbon|null $updatedAt
 * @method static \App\Database\Eloquent\Builder|Person whereId($value)
 * @method static \App\Database\Eloquent\Builder|Person whereName($value)
 * @method static \App\Database\Eloquent\Builder|Person whereCreatedAt($value)
 * @method static \App\Database\Eloquent\Builder|Person whereUpdatedAt($value)
 * 
 * @mixin \Eloquent
 */
```

In the above, reading all of `id`, `name`, `createdAt`, `updatedAt` works. But out of the wheres, only `whereId()` and `whereName()` work. `whereCreatedAt` and `whereUpdatedAt` do not.

I accept that it's not possible to get those methods to work in Laravel, however it would like to request to not generate just those multi-word docblock methods for camelCase (e.g. whereCreatedAt, whereUpdatedAt). But it should still generate the docblocks for single name properties (`whereId()` and `whereName()`) as they work fine.

PR #1167, opened 2021-03-01
Do not generate magic wheres for multi-word camelCase properties

amcsi:
## Summary
See #1166

## Type of change

### Checklist
- [ ] Update the README.md


amcsi:
I'm not sure if I should mark this as a bugfix or a breaking change. It's kind of both, but leans more towards the bugfix side.

Camel-case multi-word magic wheres really don't work, so this ought to be a considered a bugfix. People using camel case properties with magic where enabled should experience all their 2+ word `where...()` methods gone, but the single word wheres would be kept.

amcsi:
I think there ought to be a unit test for the logic determining whether a name of a property would be considered two words. I'm just not sure to which PHP class I should push the new utility method so I could cover it with a unit test.

amcsi:
Alright, I've changed the logic to be much simpler and to make use of the same method (`Str::snake()`) that Laravel uses to resolve magic wheres.

amcsi:
@mfn may I please ask for some feedback for this PR?

mfn:
@amcsi may I ask what's your use case for using `'model_camel_case_properties' => true` in the config?

amcsi:
@mfn Sure. It's simply a pre-existing standard at the company I work for. There have been other DB projects (some web projects, some not) already using this standard of camelCase column names, and our CTO asked us to keep column names in Laravel also as camelCase. As such our base model class has `public static $snakeAttributes = false;` set; what I believe is an option added to Laravel to be able to accommodate such standards.
And as such, I set `'model_camel_case_properties' => true` so that the properties like `createdAt` would `@property Carbon $createdAt` rather than `@property Carbon $created_at`, because that's how the properties need to be accessed _(though this could be automatically be detected by checking $snakeAttributes per model rather than having to check the all-affecting 'model_camel_case_properties' ide-helper property, but that's besides the point now)_.

mfn (commented):
> _(though this could be automatically be detected by checking $snakeAttributes per model rather than having to check the all-affecting 'model_camel_case_properties' ide-helper property, but that's besides the point now)_.

Uh, that's actually a nice idea!

Thanks, I got the use-case now!

PR seems safe to me, what do others think?

amcsi:
One thing I admittedly haven't done yet is actually trying out these changes in our company's project to confirm it really works well ok practice. I only relied on the tests in the ide-helper projects.

I will do so tomorrow.

That's one thing you could wait for before merging.

mfn:
Thanks, we'll wait :)

When I find time I do the same, no matter how much tests we have, there are some many factors in play. I usually temporarily just copy over the vendor files directly, it's the fastest 😅

amcsi:
@mfn 

Ok, so I've tested this out a little...

The partial magic wheres generation works...

But I realized something... I was wrong about something. In our project we don't actually use `model_camel_case_properties`. I mistakenly thought we did.
I also realized that it's incorrect to do the partial wheres generation if `model_camel_case_properties` is set, because what that is meant for is if your columns were actually snake_cased in the DB, but you wanted to access the properties camelCased; there's no reason to affect magic wheres, because they already accept camelCased (and resolve to camel_cased in the DB queries).

So what I should actually do is for the model for which magic wheres are being generated, it's the static `$snakeAttributes = false;` that needs to be checked; if that is false, it is _then_ that multi-word property wheres should not be generated.

Do you agree?

mfn:
Do you mean you want to change the behaviour of `model_camel_case_properties=true` and additionally respect Eloquents `$snakeAttributes`?

Or do you mean doing this independent? I think a separate PR would be easier to discuss this further.

amcsi:
@mfn No. I mean to not do anything at all with `model_camel_case_properties`, because it's pointless because of the reasons I wrote.

...and instead of reading the global `model_camel_case_properties` config to determine if multi-word wheres should not be generated, it's only `$snakeAttributes` (and per model) that should be read to determine if multi-word wheres should not be generated.

I would actually need to create a new PR for the new changes, because the `Test__test__1.php` file is for a model that does not have `$snakeAttributes` set to false; I'll need a new test model.

I'm closing this PR for now.

PR #1177, opened 2021-03-15
Update composer scripts to run only in dev mode

netpok:
## Summary
This PR updates the suggested composer integration to run only in dev environment. This should do the same as #1067 without introducing a new command.

Also the PR suggest executing the same commands in the post-install section, which can be helpful if repository is access from multiple locations or have different branches with different composer.json

The command:
`[ $COMPOSER_DEV_MODE -eq 0 ] || $PHP_BINARY artisan ide-helper:generate`: test that composer is not in dev mode or execute ide-helper:generate if it is.

If more readable the following is the same:
`test $COMPOSER_DEV_MODE -eq 0 || $PHP_BINARY artisan ide-helper:generate`

This should work under Linux, Mac and WSL (Windows).

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,14 +84,19 @@ Note: `bootstrap/compiled.php` has to be cleared first, so run `php artisan clea
 
 This will generate the file `_ide_helper.php` which is expected to be additionally parsed by your IDE for autocomplete. You can use the config `filename` to change its name.
 
-You can configure your `composer.json` to do this each time you update your dependencies:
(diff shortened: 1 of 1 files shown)


mfn (commented):
Will this work on windows? 🤔

netpok:
Turns out no, because windows is crap

mfn:
😬

PR #1189, opened 2021-03-18
Fix for when factory class does not exist

acurrieclark:
## Summary
Laravel artisan make:model generates a model which is automatically given the HasFactory trait.

If the Factory class has not also been created, an error is thrown causing model analysis to stop. This pull request checks if an error is thrown when looking for the factory class, and returns if one is found.

Not sure if you require a test for this or not. Let me know and I will look at adding one.

## Type of change
[ x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -1093,7 +1093,11 @@ class ModelsCommand extends Command
         }
 
         $modelName = get_class($model);
-        $factory = get_class($modelName::factory());
+        try {
+            $factory = get_class($modelName::factory());
+        } catch (\Error $error) {
+            return;
+        }
         $factory = '\\' . trim($factory, '\\');
 
         if (!class_exists($factory)) {
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -1095,7 +1095,9 @@ class ModelsCommand extends Command
         $modelName = get_class($model);
         try {
             $factory = get_class($modelName::factory());
-        } catch (\Error $error) {
+        } catch (\Throwable $t) {
+            $this->warn(sprintf('Error resolving factory from model %s: %s', $modelName, $t->getMessage()));
+
             return;
         }
         $factory = '\\' . trim($factory, '\\');


barryvdh:
I think it's good to silently fail, it's not our job to fix/detect problems imho

acurrieclark:
Forgive me, but I don't quite understand.

The check which `laravel-ide-helper` does to see if the helpers should be used is by looking for the inclusion of the `HasFactory` trait. If i run:

```bash
php artisan make:model Book
```

then the model file is created, with the `HasFactory` trait _automatically added_, but no factory is generated. If I run `php artisan ide-helper:models` at that point, no model properties will be generated for the new `Book` model, as it fails at the point which factories are checked for.

It seems that the script shouldn't stop when the factory isn't found, but should continue to generate model properties as normal.

Alternatively, there could be a different check for the use of a factory? Happy to amend the pull request if there is a preferred fix.

The only other remedy I can think of is to either create a Factory class for every model or manually remove the `HasFactory` trait from each model which doesn't have a factory class already.

barryvdh:
Yeah I meant to say I approve this PR

acurrieclark:
Ah, apologies. Thanks very much

src/Console/ModelsCommand.php
```diff
@@ -1093,7 +1093,11 @@ protected function getFactoryMethods($model)
         }
 
         $modelName = get_class($model);
-        $factory = get_class($modelName::factory());
+        try {
+            $factory = get_class($modelName::factory());
+        } catch (\Error $error) {
+            return;
+        }
```
mfn, line 1100: I would suggest to not silently do this, but at least emit a warning: we do the same in other places, e.g. https://github.com/barryvdh/laravel-ide-helper/blob/c5c9b2b9517696e1d17ce64790be42b8bb76a4d2/src/Console/ModelsCommand.php#L623

(achtung: did not test the code…)
```suggestion
        } catch (\Throwable $t) {
            $this->warn(sprintf('Error resolving factory from model of %s: %s', $modelName, $, $t->getMessage()));

            return;
        }
```

mfn (changes requested):
Basically I approve the approach but silently doing it isn't that good and we've prior art to output to the console.

- please check out my suggestion
- please add a changelog
- can you try adding a test?

Thank you @acurrieclark

ahmed-aliraqi:
Sorry, I did not notice that the issue was solved here but i think we don't need to call `factory()` merthod because it create a new instance regardless of whether the factory is exist or not.
anyway the default namespace of the factories is `\Database\Factories` and if anyone need to customize it will use `newFactory()` method so I created a new PR. to resolve this. #1196

mfn:
Thank you for the PR, though after considering we went with https://github.com/barryvdh/laravel-ide-helper/pull/1196 !

acurrieclark:
No problem. Apologies for the delay in adding tests etc. The other solution seems to work nicely.

Thanks for the feedback.

PR #1193, opened 2021-03-22
Skip post_migrate when running tests

netpok:
## Summary
This PR skips post migrate hooks when application is running unit tests. This should not cause any problems but it's unnecessary to generate the helpers hundred times. Also test should not have a sideeffect.

## Type of change

### Checklist

### DIFF
--- a/src/IdeHelperServiceProvider.php
+++ b/src/IdeHelperServiceProvider.php
@@ -19,7 +19,6 @@ use Barryvdh\LaravelIdeHelper\Listeners\GenerateModelHelper;
 use Illuminate\Console\Events\CommandFinished;
 use Illuminate\Contracts\Support\DeferrableProvider;
 use Illuminate\Database\Events\MigrationsEnded;
-use Illuminate\Foundation\Application;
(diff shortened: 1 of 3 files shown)


src/IdeHelperServiceProvider.php
```diff
@@ -35,7 +34,7 @@ class IdeHelperServiceProvider extends ServiceProvider implements DeferrableProv
      */
     public function boot()
     {
-        if ($this->app['config']->get('ide-helper.post_migrate', [])) {
+        if (!$this->app->runningUnitTests() && $this->app['config']->get('ide-helper.post_migrate', [])) {
```
mfn, line 37: Will this work with Lumen? 😬

netpok: It should: https://github.com/laravel/lumen-framework/blob/8.x/src/Application.php#L896

mfn: Thank you, LGTM!

Could you please resolve the merge conflicts? Don't see any blockerS!

mfn (approved):
Thank you!

PR #1205, opened 2021-04-03
Make all tests pass cross platform

wimski:
## Type of change

Some tests would not pass on all platforms, because they were depended on platform specific directory separators and line endings. That has been fixed in this PR.

### Checklist

### DIFF
--- a/tests/Console/EloquentCommandTest.php
+++ b/tests/Console/EloquentCommandTest.php
@@ -49,8 +49,11 @@ class EloquentCommandTest extends TestCase
             ';Unexpected no document on Illuminate\\\Database\\\Eloquent\\\Model;',
             $display
         );
+        $modelClassFilePath = preg_quote(
+            str_replace('/', DIRECTORY_SEPARATOR, '/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php')
+        );
         $this->assertRegExp(
-            ';Wrote expected docblock to .*/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php;',
+            ';Wrote expected docblock to .*' . $modelClassFilePath . ';',
             $display
         );
     }
--- a/tests/Console/ModelsCommand/DynamicRelations/Test.php
+++ b/tests/Console/ModelsCommand/DynamicRelations/Test.php
@@ -18,22 +18,24 @@ class Test extends AbstractModelsCommand
         ]);
 
         if (PHP_VERSION_ID >= 80000) {
-            $errors = <<<TXT
-Error resolving relation model of Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\DynamicRelations\Models\Dynamic:dynamicBelongsTo() : Attempt to read property "created_at" on null
-Error resolving relation model of Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\DynamicRelations\Models\Dynamic:dynamicHasMany() : Attempt to read property "created_at" on null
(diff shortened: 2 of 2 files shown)


mfn (changes requested):
I _guess_ you're talking about windows?

We should cover this platforms in CI, otherwise they will silently break again.

wimski:
> I _guess_ you're talking about windows?

Yes 😄 

> We should cover this platforms in CI, otherwise they will silently break again.

Is this something I can do (in a travis file)?

mfn:
We're using github actions, see https://github.com/barryvdh/laravel-ide-helper/blob/master/.github/workflows/run-tests.yml

I know it [supports windows](https://github.com/actions/virtual-environments#available-environments) but I never used it.

.github/workflows/run-tests.yml
```diff
@@ -48,3 +48,43 @@ jobs:
...
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          coverage: none
+          extensions: pdo_sqlite, fileinfo
```
mfn, line 80: What is `fileinfo` needed for 🤔 ?

wimski:
@mfn

I've added a job to run the tests on Windows. It only runs on PHP 8 with prefer-stable, because we only care that the test suite will pass, not all specific variations.

.github/workflows/run-tests.yml
wimski: Otherwise the packages will not install.

wimski: Otherwise the packages will not install.

.github/workflows/run-tests.yml
```diff
@@ -48,3 +48,43 @@ jobs:
 
     - name: Execute Unit Tests
       run: composer test-ci
+
+  php-windows-tests:
```
mfn, line 52: Well, _that_ is really nice!

Unless I miss something, it's essentially the same as `php-tests:` except:
- `runs-on` (duh!)
- `Set git to use LF` 😬
- `fileinfo` PHP extension


I would also agree we don't need to run it on all variations, so we could just add this via an [`include:`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-additional-values-into-combinations) and have a condition for the `Set git to use LF` (we can just always add `fileinfo`, don't see much trouble with it).

This reduces issue when we adapt the flow and having to think it's duplicated for the most parts.

What do you think?

wimski: How would that work exactly? I'm new to these actions so I just did some copy/paste until it worked.

mfn: Here's an example on https://github.community/t/create-matrix-with-multiple-os-and-env-for-each-one/16895/2 if you want to try it.

Though in this case `windows-2019` would **not** be in the initial matrix, _only_ in the (one) `include:` directive.

Regarding the condition: it's possible to use (limited) `if` condition on steps, see:
- https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif
- random blog with samples 😄 -> https://dev.to/tomassirio/if-statements-on-github-actions-545d

wimski: I've merged the jobs. They seem to pass, but I don't see the different OS settings reflected (in the name for example) 🤔

wimski: I think I've got it all working the way we want.

mfn (approved):
Nice, thank you!

wimski:
@barryvdh 👋

barryvdh:
Thanks!

PR #1212, opened 2021-04-13
Use platformName to determine db type

stockalexander:
Use platformName instead of default connection name to determine used dbms. Previous solution isn't working, when custom names are used.

This should fix the problem mentioned in https://github.com/barryvdh/laravel-ide-helper/issues/480 for custom connection names, too. Example config/database.php section:

    'default' => env('DB_CONNECTION', 'main_db'),
    'connections' => [
        'main_db' => [
            'driver' => 'mysql',
            'host' => env('DB_HOST_MAIN', '127.0.0.1'),
            'port' => env('DB_PORT_MAIN', '3306'),
            'database' => env('DB_DATABASE_MAIN', ''),
...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file.
 [Next release](https://github.com/barryvdh/laravel-ide-helper/compare/v2.10.0...master)
 --------------
 
+### Fixed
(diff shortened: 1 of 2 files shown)


mfn (approved):
Perfectly makes sense to me!

Pure luck it worked in most cases. I also checked that the platform names returned from DBAL match the existing `case` statements and added a changelog entry.

Thank you!

PR #1214, opened 2021-04-22
Use lower case column name on generated model properties.

yajra:
## Summary

This will fix the generated model properties when using oracle. 

Fixes: https://github.com/yajra/laravel-oci8/issues/641
Related PR: https://github.com/yajra/laravel-oci8/pull/648

Before PR:

```php
namespace App\Models{
/**
 * App\Models\Permission
 *
 * @property int $ID
 * @property string $NAME
```

After PR:

```php
namespace App\Models{
/**
 * App\Models\Permission
 *
 * @property int $id
 * @property string $name
```

## Type of change

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -440,8 +440,7 @@ class ModelsCommand extends Command
             return;
         }
 
-        foreach ($columns as $column) {
-            $name = $column->getName();
+        foreach ($columns as $name => $column) {
             if (in_array($name, $model->getDates())) {
                 $type = $this->dateClass;
             } else {
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -440,7 +440,8 @@ class ModelsCommand extends Command
             return;
         }
 
-        foreach ($columns as $name => $column) {
+        foreach ($columns as $column) {
+            $name = strtolower($column->getName());
             if (in_array($name, $model->getDates())) {
                 $type = $this->dateClass;
             } else {


src/Console/ModelsCommand.php
```diff
@@ -440,8 +440,7 @@ public function getPropertiesFromTable($model)
             return;
         }
 
-        foreach ($columns as $column) {
-            $name = $column->getName();
```
mfn, line 444: AFAIK this is necessary because it takes care of fully qualifying the column names in case this is necessary.

We don't cover this in the tests and I also don't have a setup makes use of this, so I can't really verify.

But looking at the implementation I see on my system from `getName()` I doubt we can just do this change?

`\Doctrine\DBAL\Schema\AbstractAsset`:
```php
    /**
     * Returns the name of this schema asset.
     *
     * @return string
     */
    public function getName()
    {
        if ($this->_namespace !== null) {
            return $this->_namespace . '.' . $this->_name;
        }

        return $this->_name;
    }
```

I also read through the references issues/PRs but not sure how the fix relates to this, sorry 🤷‍♀️

yajra: > AFAIK this is necessary because it takes care of fully qualifying the column names in case this is necessary.

Thanks, how about forcing to lower case the name? The PR just aims to use lower case property so that it can be properly used with Oracle.

```php
$name = strtolower($column->getName());;
```

> I also read through the references issues/PRs but not sure how the fix relates to this, sorry

Yes, will not fix the issue but more of a compatibility with the package. The PR I drafted on OCI8 will fix it.

mfn: `getName()` can return a quoted string (to support mixtures of upper and lower case for e.g. Postgres) and _then_ lower casing this will break.

I.e. AFAIK it can return `"Some_column"`, the quotes are here to explicitly support the mix of cases, and this would turn `"some_column"` and break.

PR #1198, opened 2021-04-01
Allowing Methods to be set or unset in ModelHooks

jenga201:
## Summary
Expanding PR #945 to include support in ModelHooks to unset and set Methods.
This can be useful for defining class inheritance properly for the IDE.

## Type of change
- [x] New feature (non-breaking change which adds functionality)

### Checklist
- [x] Existing tests have been adapted and/or new tests have been added
- [ ] Add a CHANGELOG.md entry
- [x] Update the README.md
- [x] Code style has been fixed via `composer fix-style`

PR #1215, opened 2021-04-25
Add support to have lower case properites via a config variable.

jwhulette:
## Summary
Allow column properties to be transformed to lower case via a configuration variable.

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -151,9 +151,7 @@ return [
     |
     */
 
-    'ignored_models' => [
-
-    ],
+    'ignored_models' => [],
 
     /*
     |--------------------------------------------------------------------------
@@ -196,9 +194,7 @@ return [
     |
     */
 
-    'interfaces' => [
-
-    ],
+    'interfaces' => [],
 
     /*
     |--------------------------------------------------------------------------
@@ -226,9 +222,28 @@ return [
     |  ),
     |
     */
-    'custom_db_types' => [
+    'custom_db_types' => [],
 
-    ],
+    /*
+     |--------------------------------------------------------------------------
+     | Set the column name to lower case
+     |--------------------------------------------------------------------------
+     | Some databases, like Oracle return the column names in upper case
+     |
+     | For example, normally you would see this:
+     |
+     |  * @property \Illuminate\Support\Carbon $CREATED_AT
+     |  * @property \Illuminate\Support\Carbon $UPDATED_AT
+     |
+     | With this enabled, the properties will be this:
+     |
+     |  * @property \Illuminate\Support\Carbon $created_at
+     |  * @property \Illuminate\Support\Carbon $updated_at
+     |
+     | Note, it is currently an all-or-nothing option.
+     |
+     */
+    'model_lower_case_properties' => true,
 
     /*
      |--------------------------------------------------------------------------
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -151,7 +151,9 @@ return [
     |
     */
 
-    'ignored_models' => [],
+    'ignored_models' => [
+
+    ],
 
     /*
     |--------------------------------------------------------------------------
@@ -194,7 +196,9 @@ return [
     |
     */
 
-    'interfaces' => [],
+    'interfaces' => [
+
+    ],
 
     /*
     |--------------------------------------------------------------------------
@@ -222,7 +226,9 @@ return [
     |  ),
     |
     */
-    'custom_db_types' => [],
+    'custom_db_types' => [
+
+    ],
 
     /*
      |--------------------------------------------------------------------------
@@ -230,7 +236,7 @@ return [
      |--------------------------------------------------------------------------
      | Some databases, like Oracle return the column names in upper case
      |
-     | For example, normally you would see this:
+     | For example, with Oracle normally you would see this:
      |
      |  * @property \Illuminate\Support\Carbon $CREATED_AT
      |  * @property \Illuminate\Support\Carbon $UPDATED_AT
@@ -243,7 +249,7 @@ return [
      | Note, it is currently an all-or-nothing option.
      |
      */
-    'model_lower_case_properties' => true,
+    'model_lower_case_properties' => false,
 
     /*
      |--------------------------------------------------------------------------
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -11,38 +11,40 @@
 
 namespace Barryvdh\LaravelIdeHelper\Console;
 
-use Throwable;
-use ReflectionType;
-use ReflectionClass;
-use ReflectionObject;
-use ReflectionNamedType;
-use Illuminate\Support\Str;
-use Illuminate\Console\Command;
-use Barryvdh\Reflection\DocBlock;
-use Barryvdh\Reflection\DocBlock\Tag;
-use Illuminate\Filesystem\Filesystem;
-use Illuminate\Database\Eloquent\Model;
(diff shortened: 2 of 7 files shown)


mfn:
- Please re-submit without any of the style changes
- have you checked if it's possible using the new hooks? (https://github.com/barryvdh/laravel-ide-helper/pull/945 and https://github.com/barryvdh/laravel-ide-helper/pull/1198 )

jwhulette:
I have reverted the style changes.

I have looked into the hooks but I cannot see a way to change the case of the column name, as the hooks seem to be for adding additional information.

mfn:
> Allow column properties to be transformed to lower case via a configuration variable.

What I don't understand so far: is this a technical or cosmetic change?

ide helper just provides convenient IDE assistance, what is the different on `$model->COLUMN` was `$model->column` in terms of functionality with oracle? Are both cases working equally?

jwhulette:
It's a bit of both.

When using the https://github.com/yajra/laravel-oci8 library, it returns the properties lower case by default. 

Although it can be overridden, I prefer the lower case properties. 

So, $model->COLUMN works if I manually set the properties to upper case, and $model->column works when using the default.

The issue is for methods. If using the default of uppercase, calling wherePAGENUMBER($value) results in the error: ORA-00904: "P_A_G_E_N_U_M_B_E_R": invalid identifier

Where as if the property is lower case, as wherePageNumber($value), it works properly. 

Please, let me know if you have any further questions.

mfn:
So this config change only makes sense together with https://github.com/yajra/laravel-oci8 ?

Just so I get this correct

Using regular oci8:
- `$model->COLUMN` works
- `$model->column` does not
- `$model->whereCOLUMN` does not
- `$model->whereFoo` ??

Using laravel-oci8:
- `$model->COLUMN` does not work
- `$model->column` works
- `$model->whereCOLUMN` does not (basically: this can never work due to the transformation as you pointed out `C_O_…`
- `$model->whereFoo` works

Unrelated to this, I don't think this should be enabled by default. And since you mentioned Oracle explicitly in the config documentation, I think it's fair to mention laravel-oci8 , as it seems to be required anyway for a sane integration.

Is this correct so far?

jwhulette:
> 
> 
> So this config change only makes sense together with https://github.com/yajra/laravel-oci8 ?
I believe it makes sense for any project that uses Oracle. As Oracle defaults to returning names as upper case.
> 
> Just so I get this correct
> 
> Using regular oci8:
- Since laravel-oci8 uses the OCI8 under the hood I would assume the same results, but I have not tested regular OCI8 by itself.
> 
>     * `$model->COLUMN` works
> 
>     * `$model->column` does not
> 
>     * `$model->whereCOLUMN` does not
> 
>     * `$model->whereFoo` ??
> 
> 
> Using laravel-oci8:
> 
>     * `$model->COLUMN` does not work
 - This does work if I change the default case to upper, as the package defaults to converting them to lower
>     * `$model->column` works
- Yes works
>     * `$model->whereCOLUMN` does not (basically: this can never work due to the transformation as you pointed out `C_O_…`
- No, does not work
>     * `$model->whereFoo` works
- Yes works
> 
> Unrelated to this, I don't think this should be enabled by default. And since you mentioned Oracle explicitly in the config documentation, I think it's fair to mention laravel-oci8 , as it seems to be required anyway for a sane integration.

Yes it should be defaulted to false, I have fixed the config file.
> 
> Is this correct so far?

config/ide-helper.php
```diff
@@ -219,6 +219,27 @@
 
     ],
 
+    /*
+     |--------------------------------------------------------------------------
+     | Set the column name to lower case
+     |--------------------------------------------------------------------------
+     | Some databases, like Oracle return the column names in upper case
+     |
+     | For example, normally you would see this:
```
mfn, line 228:
```suggestion
     | For example, normally with Oracle you would see this:
```

src/Console/ModelsCommand.php
```diff
@@ -442,6 +442,11 @@ public function getPropertiesFromTable($model)
 
         foreach ($columns as $column) {
             $name = $column->getName();
+
+            if ($this->hasLowerCaseProperties()) {
```
mfn, line 446: Something got b0rked with your last change, the method `hasLowerCaseProperties` isn't present anymore

tests/Console/ModelsCommand/GenerateBasicPhpdocLower/Test.php
```diff
@@ -0,0 +1,31 @@
...
+class Test extends AbstractModelsCommand
+{
+    protected function getEnvironmentSetUp($app)
+    {
+        parent::getEnvironmentSetUp($app);
+
+        $app['config']->set('ide-helper.model_lower_case_properties', true);
+    }
+
+    public function test(): void
```
mfn, line 19: This test doesn't really perform the test you intended it to do, because none of the migrations provide upper case column definitions.

mfn (changes requested):
I went once more through the PR and from my side, I'm ok with the general goal.

Looked now a bit closer and left some inline feedback, i.e. in the current version of the PR you accidentally removed `hasLowerCaseProperties()` so it doesn't really work.

Some other feedback: I appreciate you fixing the code style back, but it really doesn't make sense to keep your first commit like that. I ask you kindly if you can in the end just rebase and clean this up with a single proper commit; please also update against latest master in the meantime.

Lastly, please also add a changelog entry and see if you find a good spot in the README to mention the new config.

Thank you!

tests/Console/ModelsCommand/GenerateBasicPhpdocLower/Test.php
jwhulette: Added proper tests by creating column names that are uppercase

src/Console/ModelsCommand.php
jwhulette: Re-added the missing function

config/ide-helper.php
jwhulette: Updated doc block with your suggestion.

jwhulette:
Rebased and updated Changelog. I didn't find a good spot in the readme.

src/Console/ModelsCommand.php
```diff
@@ -17,8 +17,6 @@
 use Barryvdh\Reflection\DocBlock\Serializer as DocBlockSerializer;
 use Barryvdh\Reflection\DocBlock\Tag;
 use Composer\Autoload\ClassMapGenerator;
-use Doctrine\DBAL\Exception as DBALException;
-use Doctrine\DBAL\Types\Type;
```
mfn, line 21: These lines should not be removed 🤔

mfn (changes requested):

src/Console/ModelsCommand.php
jwhulette: Sorry about that. I put the missing lines back.

src/Console/ModelsCommand.php
```diff
@@ -17,9 +17,9 @@
 use Barryvdh\Reflection\DocBlock\Serializer as DocBlockSerializer;
 use Barryvdh\Reflection\DocBlock\Tag;
 use Composer\Autoload\ClassMapGenerator;
+use Illuminate\Console\Command;
 use Doctrine\DBAL\Exception as DBALException;
 use Doctrine\DBAL\Types\Type;
-use Illuminate\Console\Command;
```
mfn, line 22: :sadpanda:

but fix-style will correct this, so nothing to do

mfn (approved):
LGTM!

Ping @barryvdh , from my side good to merge

PR #1224, opened 2021-05-21
add config option to remove the overwrite models file prompt

powellblyth:
## Summary
This adds a config option called `` that removes the `Do you want to overwrite the existing model files?` prompt and always answers "yes" when running the command.  
This is backwards compatible, as omitting the config or setting it false leaves exactly as it was

## Type of change

- [X] This change requires a documentation update

### Checklist
- [X] Add a CHANGELOG.md entry
- [X] Update the README.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 All notable changes to this project will be documented in this file.
 
 [Next release](https://github.com/barryvdh/laravel-ide-helper/compare/v2.10.0...master)
+### Added
+- New configuration option `always_overwrite_model_files` to default to writing the PHP doc bloc comment (unless overridden with a switch)
+  Note: If set to false, or omitted, the previous BC functionality occurs
 --------------
 
 2021-04-09, 2.10.0
--- a/README.md
+++ b/README.md
@@ -140,6 +140,13 @@ Alternatively using the `--write-mixin (-M)` option will only add a mixin tag to
 writing the rest in (`_ide_helper_models.php`).
 The class name will be different from the model, avoiding the IDE duplicate annoyance.
 
+Alternately, you can set the `always_overwrite_model_files` flag to true, which is identical to using `--write` above
+
+```php
+'always_overwrite_model_files' => true,
+```
+Setting this to `false` has no effect
+
 > Please make sure to back up your models, before writing the info.
 
 Writing to the models should keep the existing comments and only append new properties/methods.
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -74,6 +74,16 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Always Overwrite Model Files
+    |--------------------------------------------------------------------------
+    |
+    | Set to true to force 'yes' as an answer to 'Do you want to overwrite the existing model files?'
+    |
+    */
+    'always_overwrite_model_files' => false,
+
+    /*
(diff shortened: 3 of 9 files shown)


mfn:
Can't the same achieved with passing `--write` or what is the difference?

powellblyth:
> Can't the same achieved with passing `--write` or what is the difference?

Yes it can be achieved that way, but you have to remember the command line switch (or teach your team). Consistent behaviour can be achieved through config.

I wrote this for laziness, but imagine an agency where several different projects had several different standards for where this meta should go, this would help ensure that updating the model was the preferred path.

mfn:
TBH instead of adding "yet another config" option for such a opinionated thing I would suggest using a script definition to run to achieve this, there are so many easy options (composer.json, Makefile, phing).

Personally I also can't remember them either neither does any in my team maybe even know the exact nature of the commands and args. We just have exactly this and no issue with; it's phing in my case:
```xml
  <target name="ide-helper" description="Generate _ide_helper.php">
    <!-- ensure cache is empty before running code -->
    <exec command="${artisan} clear-compiled" passthru="true"/>
    <exec command="${artisan} ide-helper:generate" passthru="true"/>
    <exec command="${artisan} ide-helper:models --write --reset" passthru="true"/>
    <exec command="${artisan} ide-helper:meta" passthru="true"/>
    <!-- ensure we did't leave a cache file from running the ide-helper -->
    <exec command="${artisan} clear-compiled" passthru="true"/>
  </target>
```

But that's just my 2c

powellblyth:
> TBH instead of adding "yet another config" option for such a opinionated thing I would suggest using a script definition to run to achieve this, there are so many easy options (composer.json, Makefile, phing).
> 
> Personally I also can't remember them and probably neither does any in my team maybe even know the exact nature of the commands and args. We just have exactly this and no issue with; it's phing in my case:
> 
> ```
>   <target name="ide-helper" description="Generate _ide_helper.php">
>     
>     <exec command="${artisan} clear-compiled" passthru="true"/>
>     <exec command="${artisan} ide-helper:generate" passthru="true"/>
>     <exec command="${artisan} ide-helper:models --write --reset" passthru="true"/>
>     <exec command="${artisan} ide-helper:meta" passthru="true"/>
>     
>     <exec command="${artisan} clear-compiled" passthru="true"/>
>   </target>
> ```
> 
> But that's just my 2c

a good 2c, or in my case about £0.015

For future historians, I've added
```json
    "ide-helper": "./vendor/bin/sail artisan ide-helper:models --write"
```
to my composer-json file and can run this using `composer ide-helper`

You are right it's very opinionated, but I suppose that's why I put it into config so as not to force my opinion.
It does solve my use case. I'll leave the PR here a little longer in case someone needs it, unless someone tells me it definitely won't be merged.

Thanks for the tip, I had not looked at composer scripts before.

Messhias (approved):

Messhias (approved):
I tested locally with provided example and worked without issues, but I guess could have more examples. 

But I don't see the reason also to have a lot tests since this one passed through without breaking the current package (y).

mfn:
I'm 👎 adding a separate config for which is already a switch available which can be added to any kind of scripting tool to share in a team.

Messhias:
> Yeah bit conflicted. I only ever use the --write flag, but I feel it's kinda breaking BC, that's why I added the prompt.

So it's better to we update the PR or exclude it?

PR #1231, opened 2021-07-13
Consider foreign key constraints with isRelationNullable

atushi-koga:
## Summary
When we run ide-helper:models on a class that defines a method that uses belongsTo, if the foreign key used for the relation is not null, a doc comment will be generated with the property returned as the relation class.

```php
<?php

/**
 * @property-read App\User $user  <- returned type is relation class
 * @mixin \Eloquent
 */
class Phone extends Model
{
    public function user()
    {
        return $this->belongsTo('App\User');
    }
}
```

However, there are times when we want to get the relation class **without using the foreign key constraint**, so we check for the existence of the relation class.
Therefore, static analysis with the current doc comments will fail, with the null check always resulting in false.

```php
$phone = Phone::firstOrFail();
if (is_null($phone->user)) {  // static analysis says 「It's dead code because it's always false」
    throw new RuntimeException('User Not Found');
}
$user = $phone->user
```

This problem can be solved by checking for the presence of foreign key constraints.
Is it possible to import this change?

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -100,6 +100,10 @@ class ModelsCommand extends Command
      * @var bool[string]
      */
     protected $nullableColumns = [];
+    /**
+     * @var string[]
+     */
+    protected $foreignKeyConstraintsColumns = [];
 
     /**
      * During initialization we use Laravels Date Facade to
@@ -415,6 +419,7 @@ class ModelsCommand extends Command
      * Load the properties from the database table.
      *
      * @param \Illuminate\Database\Eloquent\Model $model
+     * @throws \Doctrine\DBAL\Exception
      */
     public function getPropertiesFromTable($model)
     {
@@ -440,6 +445,7 @@ class ModelsCommand extends Command
             return;
         }
 
+        $this->setForeignKeys($schema, $table);
         foreach ($columns as $column) {
             $name = $column->getName();
             if (in_array($name, $model->getDates())) {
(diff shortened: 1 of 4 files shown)


Messhias (approved):

barryvdh:
So this would make $user nullable? Can you add a testcase for this?

barryvdh (changes requested):
Please add tests to validate this behavior.

Messhias:
> Please add tests to validate this behavior.

Hi @barryvdh, I'll try to give a test case to validate this behavior, I don't know if I can do it today, but I'll try at least on weekend to do that. 

I didn't open the PR but I guess I can help handle this.

Thank you,

atushi-koga:
@Messhias 
Oh, I'm glad for your help as I'm still new to this.

Messhias:
> @Messhias
> Oh, I'm glad for your help as I'm still new to this.

You wanna try by yourself to do the tests? If you are not able to do it I can do it, but probably it'll be only on weekends if I arrange some time to do it.

atushi-koga:
@Messhias 
Thank you for giving me a chance. I'll try!

atushi-koga:
Added a test.

Messhias (approved):
It seems fine on PHP 8.0.9.

Messhias:
@barryvdh it's seems fine now :)

barryvdh (approved):

PR #1241, opened 2021-08-08
Move default models filename to config

wimski:
## Summary
The models helper filename is the only filename that is hard-coded and not configurable. This PR makes the helper filename locations consistent.

## Type of change

### Checklist


src/Console/ModelsCommand.php
```diff
@@ -131,7 +135,8 @@ public function __construct(Filesystem $files)
      */
     public function handle()
     {
-        $filename = $this->option('filename');
+        $this->filename = $this->laravel['config']->get('ide-helper.models_filename');
```
mfn, line 138: You need to provide the default here; otherwise ppl upgrade to the latest version and this returns `null`

tests/Console/ModelsCommand/AbstractModelsCommand.php
```diff
@@ -44,6 +44,9 @@ protected function getEnvironmentSetUp($app)
             'prefix' => '',
         ]);
 
+        // Set the default models helper filename
+        $config->set('ide-helper.models_filename', '_ide_helper_models.php');
```
mfn, line 48: I think for the same reason I mention this in the app code, this actually shouldn't be necessary

mfn (changes requested):
LGTM but can you validate my feedback?

tests/Console/ModelsCommand/AbstractModelsCommand.php
wimski: I thought this was weird too, but when I dump `$config->get('ide-helper')` (the entire array), I only get the values that were set explicitly. That's why I added this value as well.

barryvdh: Did you clear your config cache? Or did you override the config. A default value should be used when calling it.

wimski: I don't know what went wrong, but the tests seem to pass without adding the value explicitly now, so I've removed these lines.

mfn (approved):
LGTM, thank you!

barryvdh (approved):

PR #1242, opened 2021-08-08
Use DI for config in ModelsCommand

wimski:
## Summary
I've used dependency injection for the configuration class in order to avoid Laravel's magic. This makes the code more discoverable and will help refactoring down the line.

## Type of change

### Checklist


mfn (approved):
Nice, LGTM!

@barryvdh 👍 from me!

---
Reminds me about https://github.com/barryvdh/laravel-ide-helper/issues/756 ; I still have non-Laravel code using this and it will break but a) no problem with that, I'll just adjust and it b) I think I'm the only one out there doing that anyway ;) [j/k but probably only a low number of ppl have to do this]

src/IdeHelperServiceProvider.php
```diff
@@ -75,8 +78,11 @@ function ($app) use ($localViewFactory) {
 
         $this->app->singleton(
             'command.ide-helper.models',
-            function ($app) {
-                return new ModelsCommand($app['files']);
+            function (Container $app): ModelsCommand {
```
barryvdh, line 81: Do we even need to register it? Laravel probably does this already right?

mfn: Ah, you mean just this?
```php
$this->app->singleton('command.ide-helper.models', ModelsCommand::class);`
```
Should work as well I think!

barryvdh: Yeah or maybe not even that.

Messhias (approved):

PR #1263, opened 2021-10-07
Add support for multiple or alternately named auth providers

miken32:
## Summary
When determining return value for functions like `Auth::user()`, only the hardcoded path `auth.providers.users.model` is checked in the configuration. In the case of multiple guards (or a guard not named "users") the resulting return type is incomplete. This fix checks every entry in the `auth.providers` array to build an intersection type (e.g. `App\User|App\Admin`) for use as return.

I considered adding more checks (e.g. make sure `class_exists()` and it implements `Authenticatable`) but if you have invalid class names as part of your application config, that's not this tool's problem.

## Type of change

### Checklist
- [ n/a ] Existing tests have been adapted and/or new tests have been added
- [ n/a ] Update the README.md

### DIFF
--- a/.github/workflows/run-integration-tests.yml
+++ b/.github/workflows/run-integration-tests.yml
@@ -18,7 +18,7 @@ jobs:
             COMPOSER_NO_INTERACTION: 1
         strategy:
             matrix:
-                php: [8.0, 7.4, 7.3]
+                php: [8.1, 8.0, 7.4, 7.3]
                 lumen: [8.*]
         name: P${{ matrix.php }} - Lumen${{ matrix.lumen }}
         steps:
@@ -75,7 +75,7 @@ jobs:
             COMPOSER_NO_INTERACTION: 1
         strategy:
             matrix:
-                php: [8.0, 7.4, 7.3]
+                php: [8.1, 8.0, 7.4, 7.3]
                 laravel: [8.*]
         name: P${{ matrix.php }} - Laravel${{ matrix.laravel }}
         steps:
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04, windows-2019]
-        php: [8.0, 7.4, 7.3]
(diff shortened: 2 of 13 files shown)


jonnywilliamson:
I grabed this code change and did a very quick test with it. I got the following generated output in the ide_helper.php

```php
         * Get the currently authenticated user.
         *
         * @return \App\Models\User|App\Models\IalpaMember|null
         * @static 
         */ 
        public static function user()
        {
                        /** @var \Illuminate\Auth\SessionGuard $instance */
                        return $instance->user();
        }
```

Note the missing `\` at the start of the second User model. My IDE throws up an error.

CleanShot 2021-11-13 at 13 25 56@2x


Not sure if it's this PR or another section of the code that's causing the issue.

miken32:
@jonnywilliamson it just looks at what’s in your auth config file. Did you check that?

jonnywilliamson:
> @jonnywilliamson it just looks at what’s in your auth config file. Did you check that?

So the original `auth.php` file, fresh in a new install looks like this:

https://github.com/laravel/laravel/blob/8.x/config/auth.php#L62-L66

With no leading `\`,

So I had copied that like follows (unintentionally):

CleanShot 2021-11-15 at 14 51 32@2x


So both of them had missing `\` at the start but it seems to add it for the first one when generating the `ide_helper.php` file.

I can of course add it myself manually, no biggie but i wonder why it's done automatically for the first provider?

miken32:
I have no idea @jonnywilliamson. You can see the changes I made in this commit. I'm just taking what the `config()` helper gives me, same as the old code was. Neither does anything special with one model over the other.

That said, it should probably be making it absolute anyway, since the methods are in a namespace block at that point. I'll update the PR.

miken32:
@jonnywilliamson I'd bet what's happening is somewhere the code is assuming this is a single class name, and it's sticking a leading backslash at the beginning to make it absolute. But of course with a union type only the first class in the list will be absolute. The change I pushed out will definitely resolve that issue.

jonnywilliamson:
@miken32 Thank you michael for looking into it. I'd say you are spot on with that analysis.

Either way, I appreciate you addressing it so quickly. I hope this gets merged soon! Thanks.

PR #1285, opened 2021-12-16
Allow table inspection in several database connections at once

### LINKED ISSUE #1141: Properties are written as read or write only when just one mutator/accessor is provided
### Versions:
- ide-helper Version: 2.9.0
- Laravel Version: 8.x
- PHP Version: 8.0

### Description:

We have a multi-tenant application where models are split in 2 DB connections : system (app DB) and tenant (each customer has one)

Models on the tenant database do not have their attributes discovered from the database.

Package used to manage tenancy is hyn-multitenant which basically dynamically builds the Laravel DB connection information depending on the active tenant (which could be determined by the subdomain for instance or programmatically)

### Issue

The code linked below does not get the proper database which should match a tenant database

https://github.com/barryvdh/laravel-ide-helper/blob/f0959c1184c6f8f9e50afa6d9b6c1eb7691ae3ae/src/Console/ModelsCommand.php#L418-L436

$database stays null

### Request

We should have a mecanism to allow discovery of those attributes.

One possible way to allow that could be to have a configuration entry where we could override the database according to the connection name.

Something like

```
   'db_mapping'=> [
      // connection => database name
      'system' => 'homestead',
      'tenant' => 'my_tenant',
   ],
```

Then code to get the database could be fixed liked that:

```
       $connectionName = $model->getConnection()->getName();
       $database = config("ide-helper.db_mapping.$connectionName");
        if (strpos($table, '.')) {
            [$database, $table] = explode('.', $table);
        }
```

I have tested that on our project. There should be no difference in previous behaviour if the configuration is not specified.

Any thoughts?

### LINKED ISSUE #1280: Always generate @property when a field is a database column
### Versions:
- ide-helper Version: 2.10.0
- Laravel Version: 8.x
- PHP Version: 7.4

### Description:

We have models with property mutators. The generated code looks fine when the mutators are both created (getter + setter).

However, the comment is not really correct when only one of them is included.

Mutators can be used for instance to clean up data when saving a field, but we do not need to implement the corresponding getter.

The IDE is fooled to consider that it is not safe to write/read from those properties whereas in reality it is.

### Steps To Reproduce:

```
class Sample extends Model {

    protected $fillable = ['field_one', 'field_two', 'field_three', 'field_four'];

    // A field which only has a getter
    public function getFieldOneAttribute(): string {
        return Str::upper($this->attributes['field_one'] ?? '');
    }

    // A field which only has a setter
    public function setFieldTwoAttribute($value): string {
        $this->attributes['field_two'] = Str::lower($value ?? '');
    }

    // A field with both
    public function getFieldThreeAttribute(): string {
        return Str::upper($this->attributes['field_three'] ?? '');
    }

    public function setFieldThreeAttribute($value): string {
        $this->attributes['field_three'] = Str::upper($value ?? '');
    }
}
```

In that case, the following is generated:

```
 * @property-read $field_one
 * @property-write $field_two
 * @property $field_three
 * @property $field_four
```

PHPStorm will then issue an error if you do that:

```
$model->field_one = 'something'; // This is however correct as we want to write that value as it is to the DB
$something = $model->field_two; // This is however correct as we want to read that value as it is from the DB
```

Would it be possible to always have `@property` when the field with a getter/setter is **also** a database column? And keep the current behaviour when the implemented getter/setter is defining a new computed property (in which case, being more specific about -write or -read is indeed the expected behaviour)

vpratfr:
fixes #1141 and probably fixes #1280 too which looks like a side effect of that original issue.

Currently the package cannot handle schema discovery in more than a single database at once. In multi-tenant applications it is very frequent that there are at least two databases active at some point in time: landlord (data common to all tenants) and tenant (specific to one tenant).

When the configuration variable `db_mapping` is not specified, it defaults to null which then makes the code strictly behave as before. Hence this introduces no breaking change.

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -467,8 +467,9 @@ class ModelsCommand extends Command
             }
             $databasePlatform->registerDoctrineTypeMapping($yourTypeName, $doctrineTypeName);
         }
-
-        $database = null;
+        
+        $connectionName = $model->getConnection()->getName();
+        $database = config("ide-helper.db_mapping.$connectionName");
         if (strpos($table, '.')) {
             [$database, $table] = explode('.', $table);
         }


src/Console/ModelsCommand.php
```diff
@@ -467,8 +467,9 @@ public function getPropertiesFromTable($model)
             }
             $databasePlatform->registerDoctrineTypeMapping($yourTypeName, $doctrineTypeName);
         }
-
-        $database = null;
+        
+        $connectionName = $model->getConnection()->getName();
+        $database = config("ide-helper.db_mapping.$connectionName");
```
mfn, line 472: I have not checked the code yet, but it's wrong to use the global helper here; see all the other instances in this file how they access the config (resolve via container):

```php
$this->laravel['config']->get();
```
This is important as not everyone runs this library on a full framework installation, some might only use the illuminate/ components and not have this defined.

vpratfr: Correct, sorry about that. This is now fixed.

mfn (commented):
Although the change looks small, TBH I'm not sure it should be merged.

There's no "official" way how to achieve this, multiple 3rd party packages have different approaches and different best practices.

Is there a chance we could make this available for hooking into for 3rd party libs _to_ ide-helper?

Besides that, is there a way to even test is? I'd argue the majority is not using such an approach and we might break it in the future as we've no coverage for it

vpratfr:
Giving a way to hook into this part of the code would be fine for us. The issue we have is that there is no way to plug any logic into that part of the code as of now. 

That said, this is not even a matter of multi-tenancy: any app which uses 2 or more DB connections to fetch it's data from is exposed to the same issue. Multi-tenancy is just the most common use case for such a need.

To me, allowing to specify a physical DB name in front of a logical connection name is not specific to multi-tenancy. Again, if I store data in 2 different DBs, I'd see the same issue.

PR #1289, opened 2021-12-30
Support Laravel 8.77 Attributes

SimonJnsson:
## Summary
This PR adds support for the new Attribute syntax added in Laravel 8.77, added by https://github.com/laravel/framework/pull/40022.

Using Reflection on the get/set method provided to the Attribute, we can determine the return type of the getter and setter.

Below is an example usage (also found in added test).

```php
use Illuminate\Database\Eloquent\Casts\Attribute;

class Users extends Model
{
    public function name(): Attribute
    {
        return new Attribute(
            get: function (?string $name): ?string {
                return $name;
            },
            set: function (?string $name): ?string {
                return $name === null ? null : ucfirst($name);
            }
        );
    }
}

// => after generate models

/**
 * App\Models\Users
 * 
 * @property string|null $name
 * …
 */
```

## Type of change

This PR should be non breaking, but will however require laravel/framework >= 8.77 to work.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file.
 [Next release](https://github.com/barryvdh/laravel-ide-helper/compare/v2.10.0...master)
 --------------
 ### Added
+- Add support for Laravel 8.77 Attributes [\#1289 / SimonJnsson](https://github.com/barryvdh/laravel-ide-helper/pull/1289)
+
+### Added
 - Add support for cast types `decimal:*`, `encrypted:*`, `immutable_date`, `immutable_datetime`, `custom_datetime`, and `immutable_custom_datetime` [#1262 / miken32](https://github.com/barryvdh/laravel-ide-helper/pull/1262)
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


mfn (changes requested):
Approach overall LGTM thus I enabled running the tests, but they are failing for some of the PHP versions we need to support.

Can you please check this?

SimonJnsson:
> Approach overall LGTM thus I enabled running the tests, but they are failing for some of the PHP versions we need to support.
> 
> Can you please check this?

Thanks for taking the time to look at it! I'll get to work on the changes as soon as I get the time 👍

barryvdh:
Looks good. I would just skip the tests when the attribute doesn't exist :)

SimonJnsson:
@mfn i've updated the code to not use code that's incompatible with php 7.3. This should fix the failing unit test.

In regards to the failing integration tests, i'm a bit unsure why these are failing as it seems to somehow be happening when installing dependencies. Any pointers would be greatly appreciated.

SimonJnsson:
@barryvdh sounds good, i've added a check for existance of the Attribute class in the test :+1:

mfn (approved):
Tests green, change LGTM!

> In regards to the failing integration tests

See no failing tests? 🤔

SimonJnsson:
> 

I was referring to this https://github.com/barryvdh/laravel-ide-helper/runs/4667417056?check_suite_focus=true

But it might've been a random one-off as it sees to be gone now :+1:

PR #1292, opened 2022-01-03
Improve performance for supporting Laravel 8.77+ cast attributes

### LINKED ISSUE #1291: Latest commit causes massive slowdown on writing to models
### Versions:
- ide-helper Version: last commit #29dd57e1e27540cefea914500fc6cca3ee33dbed
- Laravel Version: 8.77.1
- PHP Version: 8.0

### Description:

Using the absolute latest commit of this package this evening when wanting to test out the new support for attributes in 8.77 (https://github.com/barryvdh/laravel-ide-helper/pull/1289), I noticed that the generation/writting to the models was SO much slower than it ever has been in the past.

I checked out multiple commits and the only one to show this issue was the one commited today (#29dd57e1e27540cefea914500fc6cca3ee33dbed)

I took a video to show the difference. Normally I'd speed up the video but in this case it makes more sense to let it run in real time. This is what I did.

* I installed the SECOND most recent commit (#6244a931a00ee313a7081a86d217fbb8e48c7aad) in my project.
* As part of my `composer.json` `post-update-cmd` is the following command

```
        "post-update-cmd": [
             <snip>
            "@php artisan ide-helper:models --smart-reset --phpstorm-noinspections --write",
             <snip>
        ],
```

Just after that gets installed you can see it writting new docblocks to all my models. It takes less than a second and whizzes through.

* Then I install the latest commit (#29dd57e1e27540cefea914500fc6cca3ee33dbed).
* Watch how it takes nearly 30secs to go through all the models.

Something in the latest commit with the new attribute support is causing a lareg slow down but I don't know where or how to debug why it's taking so much time.

I hope the video helps explain it better! It's about a minute long. 

Thank you!

https://youtu.be/c8elNXVuEiA

mfn:
## Summary
#1289 calls `getReturnType` for every method on every model, which in
turn calls `getReturnTypeFromDocBlock` which has this code:
```php
$phpDocContext = (new ContextFactory())->createFromReflector($reflection);
```
Extracting the docblock is super slow, always has been. Now that we do
this for every method, this adds up a lot.

Performance on a private commercial project _before_ #1289 was introduced:
```
$ time ./artisan ide-helper:models --write --reset >/dev/null

real	0m2.857s
user	0m1.835s
sys	0m0.129s
```
After #1289 :
```
$ time ./artisan ide-helper:models --write --reset >/dev/null

real	0m54.147s
user	0m47.132s
sys	0m1.047s
```

However, in this case we **do not need** the phpdoc fallback (which is
legitimate and by design for many other cases), because also the Laravel
implementation only works by inspecting the _actual type_, see
https://github.com/laravel/framework/blob/e0c2620b57be6416820ea7ca8e46fd2f71d2fe35/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L570-L575
```php
$returnType = (new ReflectionMethod($this, $method))->getReturnType();

return static::$attributeMutatorCache[get_class($this)][$key] = $returnType &&
            $returnType instanceof ReflectionNamedType &&
            $returnType->getName() === Attribute::class &&
            is_callable($this->{$method}()->get);
```

This side-stepping the phpdoc parsing a) still works correctly and b)
brings us back to the previous performance characteristics:
```
time ./artisan ide-helper:models --write --reset >/dev/null

real	0m2.987s
user	0m1.915s
sys	0m0.120s
```

Fixes #1291

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -562,7 +562,7 @@ class ModelsCommand extends Command
             sort($methods);
             foreach ($methods as $method) {
                 $reflection = new \ReflectionMethod($model, $method);
-                $type = $this->getReturnType($reflection);
+                $type = $this->getReturnTypeFromReflection($reflection);
                 $isAttribute = is_a($type, '\Illuminate\Database\Eloquent\Casts\Attribute', true);
                 if (
                     Str::startsWith($method, 'get') && Str::endsWith(
--- a/tests/Console/ModelsCommand/Attributes/Models/Simple.php
+++ b/tests/Console/ModelsCommand/Attributes/Models/Simple.php
@@ -20,4 +20,24 @@ class Simple extends Model
             }
         );
     }
+
+    /**
+     * ide-helper does not recognize this method being an Attribute
+     * because the method has no actual return type;
+     * phpdoc is ignored here deliberately due to performance reasons and also
+     * isn't supported by Laravel itself.
+     *
+     * @return \Illuminate\Database\Eloquent\Casts\Attribute
+     */
+    public function notAnAttribute()
+    {
+        return new Attribute(
+            function (?string $value): ?string {
+                return $value;
+            },
+            function (?string $value): ?string {
+                return $value;
+            }
+        );
+    }
 }
--- a/tests/Console/ModelsCommand/Attributes/__snapshots__/Test__test__1.php
+++ b/tests/Console/ModelsCommand/Attributes/__snapshots__/Test__test__1.php
@@ -31,4 +31,24 @@ class Simple extends Model
             }
         );
     }
+
+    /**
+     * ide-helper does not recognize this method being an Attribute
+     * because the method has no actual return type;
+     * phpdoc is ignored here deliberately due to performance reasons and also
+     * isn't supported by Laravel itself.
+     *
+     * @return \Illuminate\Database\Eloquent\Casts\Attribute
+     */
+    public function notAnAttribute()
+    {
+        return new Attribute(
+            function (?string $value): ?string {
+                return $value;
+            },
+            function (?string $value): ?string {
+                return $value;
+            }
+        );
+    }
 }


barryvdh:
Nice :)

mfn:
Pinging also @SimonJnsson who brought us #1289 🎉 

The difference _can_ be seen in the test suite execution here too but I probably would not have noticed it:
- From #1289 : https://github.com/barryvdh/laravel-ide-helper/actions/runs/1648388510
  2m8s
- This PR : https://github.com/barryvdh/laravel-ide-helper/actions/runs/1650902713
  1m18s

barryvdh (approved):

SimonJnsson:
@mfn thanks for the ping 👍 

Good catch and good solution 👍 Glad someone noticed when i didn't 😅

jonnywilliamson:
@mfn thanks so much for the quick analysis and fix!

Yes it's working just like before. Lovely and fast again!


======
Just one last (and more than likely naive) question. I assume it is a requirement that the function for `get` and `set` have a return type otherwise the docblock will not be updated?

Ie. If these two lines do NOT have a return type specifically specified, then nothing will be added to the docblock?

image

It's just because I assumed that if the function arguments had type hints then it could have been inferred and I spend 10 mins wondering why I wasn't getting anything until I noticed I didn't have a return type specified.

Thanks!

SimonJnsson:
> @mfn thanks so much for the quick analysis and fix!
> 
> Yes it's working just like before. Lovely and fast again!
> 
> 
> ======
> Just one last (and more than likely naive) question. I assume it is a requirement that the function for `get` and `set` have a return type otherwise the docblock will not be updated?
> 
> Ie. If these two lines do NOT have a return type specifically specified, then nothing will be added to the docblock?
> 
> image
> 
> It's just because I assumed that if the function arguments had type hints then it could have been inferred and I spend 10 mins wondering why I wasn't getting anything until I noticed I didn't have a return type specified.
> 
> Thanks!
> 

Yep, we look at the return type of the get and set methods, thus if these are not provided no doc blocks will be created.

PR #1297, opened 2022-01-13
added laravel 9 support

rcerljenko:
## Summary
- Added Laravel 9 Support to `composer.json` and tests

### DIFF
--- a/.github/workflows/run-integration-tests.yml
+++ b/.github/workflows/run-integration-tests.yml
@@ -77,6 +77,9 @@ jobs:
             matrix:
                 php: [8.1, 8.0, 7.4, 7.3]
                 laravel: [8.*, 9.*]
+                exclude:
+                  - php: [7.4, 7.3]
+                    laravel: 9.*
         name: P${{ matrix.php }} - Laravel${{ matrix.laravel }}
         steps:
             - name: Checkout code
--- a/.github/workflows/run-integration-tests.yml
+++ b/.github/workflows/run-integration-tests.yml
@@ -76,7 +76,7 @@ jobs:
         strategy:
             matrix:
                 php: [8.1, 8.0, 7.4, 7.3]
-                laravel: [8.*, 9.*]
+                laravel: [8.*]
                 exclude:
                   - php: 7.4
                     laravel: 9.*
--- a/.github/workflows/run-integration-tests.yml
+++ b/.github/workflows/run-integration-tests.yml
@@ -78,7 +78,9 @@ jobs:
                 php: [8.1, 8.0, 7.4, 7.3]
                 laravel: [8.*, 9.*]
                 exclude:
-                  - php: [7.4, 7.3]
+                  - php: 7.4
+                    laravel: 9.*
+                  - php: 7.3
                     laravel: 9.*
         name: P${{ matrix.php }} - Laravel${{ matrix.laravel }}
         steps:
--- a/.github/workflows/run-integration-tests.yml
+++ b/.github/workflows/run-integration-tests.yml
@@ -76,7 +76,7 @@ jobs:
         strategy:
             matrix:
                 php: [8.1, 8.0, 7.4, 7.3]
-                laravel: [8.*]
+                laravel: [8.*, 9.*]
         name: P${{ matrix.php }} - Laravel${{ matrix.laravel }}
         steps:
             - name: Checkout code
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -26,12 +26,14 @@ jobs:
         exclude:
           - php: 8.0
             dependency-version: prefer-lowest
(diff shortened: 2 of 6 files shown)


barryvdh (approved):

.github/workflows/run-integration-tests.yml
```diff
@@ -76,7 +76,7 @@ jobs:
         strategy:
             matrix:
                 php: [8.1, 8.0, 7.4, 7.3]
-                laravel: [8.*]
+                laravel: [8.*, 9.*]
```
mfn, line 79: Laravel 9 does not support PHP < 8, so those combinations need to be excluded from the matrix

.github/workflows/run-tests.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       matrix:
         os: [ubuntu-20.04, windows-2019]
         php: [8.1, 8.0, 7.4, 7.3]
-        laravel: [8.*]
+        laravel: [8.*, 9.*]
```
mfn, line 24: Same as above

mfn (changes requested):
Some combinations of PHP / Laravel are not valid in the test matrix, can you please fix them?

.github/workflows/run-tests.yml
rcerljenko: yes i see because laravel 9 is only php8+ compatible but i'm not that much familiar with github actions to split php matrix... can you help?

mfn: You've to add an `exclude` key, see https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-excluding-configurations-from-a-matrix for an example

rcerljenko: thx! ok i've updated the PR

.github/workflows/run-integration-tests.yml
```diff
@@ -76,7 +76,10 @@ jobs:
...
+                exclude:
+                  - php: [7.4, 7.3]
```
mfn, line 81: The matrix still generates jobs for PHP 7.4 and Laravel 9, see https://github.com/barryvdh/laravel-ide-helper/actions/runs/1692091626

Maybe the `[…, …]` is not supported and you've to manually list each?
```
- php 7.3
  laravel: 9.*
- php 7.4
  laravel: 9.*
```
Just a guess though

mfn (changes requested):
We're not yet there, the matrix isn't correctly generated yet, see my inline feedback.

Another thing: L9 is not yet available for the integration tests, see https://github.com/barryvdh/laravel-ide-helper/runs/4802033128?check_suite_focus=true#step:4:13

So until this works, we can't merge and probably have to wait until the actual release.

.github/workflows/run-integration-tests.yml
rcerljenko: ok i've updated matrix to suggested syntax.. sorry it's my first time with github actions syntax

mfn: np!

yeah it can be a PITA, also since you're first time contributor, we've to manually enabled running the workflows everytime 😅 (I just did, let's wait for the result)

mfn: We're one step further, next problem: orchestra/benchmark also needs to have adapted version constraints, see https://github.com/barryvdh/laravel-ide-helper/runs/4803289512?check_suite_focus=true#step:5:20

I see they have not yet made a new 7.* release at https://github.com/orchestral/testbench-core but their branch is also updated. Maybe we can use `dev-master` there or we need to wait.

FTR: If you feel overwhelmed / frustrated, don't be: I can take it from here later if you want.

rcerljenko: @mfn if it's not a problem please take it from here because i don't wan't to break something... thx for the support!  :)

mfn:
@barryvdh nice, thx; seems to work already

Now we've to wait for the composer laravel installer to switch to L9 I guess?

barryvdh:
Yeah let's skip the integration test for Laravel 9 for now, but do just the unit tests.

.github/workflows/run-integration-tests.yml
barryvdh: Let's wait for the 9.x release to run the ingration tests.

barryvdh:
The tests run :)

rcerljenko:
@barryvdh then it's safe to merge?  :)

barryvdh:
Yeah if @mfn approves

mfn (approved):
Nice work all, thanks! 🙏🏼 

@barryvdh can you add this to the changelog? I can't push unfortunately:
`- Added Laravel 9 support [#1297 / rcerljenko](https://github.com/barryvdh/laravel-ide-helper/pull/1297)`

PR #1308, opened 2022-01-25
Ignore column properties via config and allow unset properties in hooks

gabsource:
## Summary

This PR add options to ignore some properties : 
- via the config file using a new entry `ignored_models_properties` (applies only to table column)
- via the unsetProperty in a hook

It is useful to prevent PhpDoc generation for properties that should be used via a model cast only through a custom class.

## Type of change

### Checklist
- [] Add a CHANGELOG.md entry
- [x] Update the README.md

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -489,9 +489,14 @@ class ModelsCommand extends Command
             return;
         }
 
+        $ignoredProperties = array_values($this->laravel['config']->get('ide-helper.ignored_models_properties.' . get_class($model), []));
+
         $this->setForeignKeys($schema, $table);
         foreach ($columns as $column) {
             $name = $column->getName();
+            if(in_array($name, $ignoredProperties)){
+                continue;
+            }
             if (in_array($name, $model->getDates())) {
                 $type = $this->dateClass;
             } else {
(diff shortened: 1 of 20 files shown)


mfn:
Wouldn't be the point of the hooks that we don't add configs for every use-case on earth?

Like: sure, let's add the hook method but not the config?

gabsource:
> Wouldn't be the point of the hooks that we don't add configs for every use-case on earth?
> 
> Like: sure, let's add the hook method but not the config?

Sure. Found it more convenient (less verbose) but indeed it's for a specific use case and may pollute the config file.

src/Console/ModelsCommand.php
```diff
@@ -830,7 +837,7 @@ public function setMethod($name, $type = '', $arguments = [], $comment = '')
 
     public function unsetMethod($name)
     {
-        unset($this->methods[strtolower($name)]);
+        unset($this->methods[$name]);
```
mfn, line 840: Not sure if we should change this…

It's from the initial implementation https://github.com/barryvdh/laravel-ide-helper/pull/1198/files#diff-2df8ee218526540f71334c651990edfbfe5345cc53a521b0259c874ff24abd9aR766

ping @jenga201 if you remember why?

PR #1313, opened 2022-02-08
Replace composer with copy of classmap generator

barryvdh:
This replaces the dependency on composer/composer and just uses the copy of the classmap generator (slightly modified).


barryvdh:
I'm merging this already because of a conflict with Composer and Laravel 9.

barryvdh:
No indeed. See https://github.com/composer/composer/discussions/10527 for future plans :)

PR #1316, opened 2022-02-11
helps generating helper files for better code suggestion for laravel

ghost:
## Summary
Edited the template file, filtered out the "empty" classes whose methods are all inherited from parent classes and have none of their own, made createLocalViewFactory as public in IdeHelperServiceProvider so that developers would have the chance to extend it and implement custom logic.

## Type of change

### Checklist

### DIFF
--- a/tests/_ide_helper.php
+++ b/tests/_ide_helper.php
@@ -1,0 +1,19577 @@
+<?php
+
+declare(strict_types=1);
+// @formatter:off
+// phpcs:ignoreFile
+
+/**
+ * A helper file for Laravel, to provide autocomplete information to your IDE
+ * Generated for Laravel 8.83.0.
+ *
+ * This file should not be included in your code, only analyzed by your IDE!
+ *
+ * @author Barry vd. Heuvel <barryvdh@gmail.com>
+ * @see https://github.com/barryvdh/laravel-ide-helper
+ */
+
+    namespace Illuminate\Support\Facades {
+            /**
+     *
+     *
+     * @see \Illuminate\Contracts\Foundation\Application
+     */
+        class App
+        {
+            /**
+         * Get the version number of the application.
+         *
+         * @return string
+         * @static
+         */
+            public static function version()
+            {
+                /** @var \Illuminate\Foundation\Application $instance */
+                return $instance->version();
+            }
+            /**
+         * Run the given array of bootstrap classes.
+         *
+         * @param string[] $bootstrappers
+         * @return void
+         * @static
+         */
+            public static function bootstrapWith($bootstrappers)
+            {
+                /** @var \Illuminate\Foundation\Application $instance */
+                $instance->bootstrapWith($bootstrappers);
+            }
+            /**
+         * Register a callback to run after loading the environment.
+         *
+         * @param \Closure $callback
+         * @return void
+         * @static
+         */
+            public static function afterLoadingEnvironment($callback)
+            {
+                /** @var \Illuminate\Foundation\Application $instance */
+                $instance->afterLoadingEnvironment($callback);
+            }
+            /**
+         * Register a callback to run before a bootstrapper.
+         *
+         * @param string $bootstrapper
+         * @param \Closure $callback
+         * @return void
+         * @static
+         */
+            public static function beforeBootstrapping($bootstrapper, $callback)
+            {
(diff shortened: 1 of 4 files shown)


mfn:
None of this is covered in tests so we've to be careful here. I read your description but I'm not sure: what does this change "fix" or "improve"?

zbcchen:
I think it should be a `fix`, it maybe doesn't need the test, because I had not edited your logic code, let me tell you why I edit the `helper.php` template file instead of other files.
1. when I add a `macro` to the `Illuminate\Console\Command` "class" in one of `ServiceProvider`, such as below 

`Command::macro('info', function ($output, $message) {
            $output->writeln("<info>$message</info>");
        });`
then you can call the `ide-helper:generate` command, you will get the _ide_helper.php that contain some `empty` Command classes. but they are not In the expectation of developers. so I think the `empty` classes should not appear in the `_ide_helper.php`.
also, I think I can declare your ServiceProvider, but I found the main method of generating helper file is `private`.so I push this pr.

zbcchen:
> I think it should be a `fix`, it maybe doesn't need the test, because I had not edited your logic code, let me tell you why I edit the `helper.php` template file instead of other files.
> 
> 1. when I add a `macro` to the `Illuminate\Console\Command` "class" in one of `ServiceProvider`, such as below
> 
> `Command::macro('info', function ($output, $message) { $output->writeln("<info>$message</info>"); });` then you can call the `ide-helper:generate` command, you will get the _ide_helper.php that contain some `empty` Command classes. but they are not In the expectation of developers. so I think the `empty` classes should not appear in the `_ide_helper.php`. also, I think I can declare your ServiceProvider, but I found the main method of generating helper file is `private`.so I push this pr.



> None of this is covered in tests so we've to be careful here. I read your description but I'm not sure: what does this change "fix" or "improve"?

I think it should be a fix, it maybe doesn't need the test, because I had not edited your logic code, let me tell you why I edit the helper.php template file instead of other files.

when I add a macro to the Illuminate\Console\Command "class" in one of ServiceProvider, such as below
Command::macro('info', function ($output, $message) { $output->writeln("<info>$message</info>"); });
then you can call the ide-helper:generate command, you will get the _ide_helper.php that contain some empty Command classes. but they are not In the expectation of developers. so I think the empty classes should not appear in the _ide_helper.php.
also, I think I can declare your ServiceProvider, but I found the main method of generating helper file is private.so I push this pr.

zbcchen:
The scenario I am trying to improve is, when a parent class uses the `Macroable` Trait if we run the `ide-helper:generate` command, the _ide_helper.php we get contains some 'empty' child classes without any self-owned methods that are not intended to be in the helper file.

I am considering improving it in a less invasive way: when we filter the `getMacroableClass`, we could remove the classes that do not use the `Macroable` Trait by itself.

The attachments are the files generated before and after this change was applied.

[_ide_helper_improved.txt](https://github.com/barryvdh/laravel-ide-helper/files/8067500/_ide_helper_improved.txt)
[_ide_helper_origin.txt](https://github.com/barryvdh/laravel-ide-helper/files/8067505/_ide_helper_origin.txt)

zbcchen:
I have added the appropriate tests. now you can run the test to view different~

zbcchen:
I have added tests for my changes. If there are still other problems, please let me know and I will revise them immediately. Thank you

zbcchen:
@mfn

tests/_ide_helper.php
```diff
@@ -0,0 +1,19577 @@
+<?php
```
mfn, line 1: Was this file commited on purpose?

It seems `generateAndGetIdeHelperFileContent` generates it all the time anyway?

zbcchen: yes，it's a test temp file. it will provides its content to test.

src/IdeHelperServiceProvider.php
```diff
@@ -41,11 +41,6 @@ public function boot()
             });
         }
 
-        if ($this->app->has('view')) {
-            $viewPath = __DIR__ . '/../resources/views';
-            $this->loadViewsFrom($viewPath, 'ide-helper');
-        }
```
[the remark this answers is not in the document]
zbcchen: I found the same effect code in the `register`
`
$localViewFactory = $this->createLocalViewFactory();
`
so I removed this.

PR #1319, opened 2022-02-14
Fix BadMethodCallException when no fake method is available.

TiiFuchs:
## Summary
When executing `php artisan ide-helper:generate` it can occur that an object behind an Facade that does not implement the fake() method can trigger an BadMethodCallException in Alias.php.

The corresponding try-catch Block has no catch, so I guess it does not catch anything.
I added a catch for this case.

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/src/Alias.php
+++ b/src/Alias.php
@@ -222,6 +222,8 @@ class Alias
             if ($fake !== $real) {
                 $this->addClass(get_class($fake));
             }
+        } catch (\Exception $e) {
+            //
         } finally {
             $facade::swap($real);
         }


mfn:
> it can occur that an object behind an Facade that does not implement the fake() method can trigger an BadMethodCallException in Alias.php.

But why is this not caught by this code then, just a few lines above? https://github.com/barryvdh/laravel-ide-helper/blob/c5e18beff1a2933913f86d56281581566e5c8705/src/Alias.php#L213-L215

TiiFuchs:
In my case the problem was the Http Facade... The Http Facade has a fake() static method (this one gets checked and exists). But later in the call stack Illuminate\Http\Client\PendingRequest has no fake() method.

you can take a look at my stack trace here: https://flareapp.io/share/J7ovApkP#F23

mfn (commented):
Oh my, thanks that helps.

I've mixed feelings about swallowing this completely, maybe use `$this->error` as done in other cases?

Also I would suggest to catch `\Throwable` while we're at it.

Do you think you can add a test for this?

TiiFuchs:
With that new knowledge, I'm not sure anymore if we need a change at all...
Since the signature of that object gets through my own changes slightly different.

I guess it would not harm to make this more robust at this point... What do you say, @mfn?

mfn:
I tried your code but it doesn't break for me, using ide-helper 2.12.2

For simplicity I put this in my `AppServiceProvider::register():
```php
        $this->app->extend(Factory::class, function (Factory $http) {
            return $http->withOptions([
                    'proxy'  => '',
                    'verify' => false,
                ]);
        });
```
It doesn't break for me, no errors, ide_helper is generated.

However I'm not sure, is this correct?

I've never used the Http client, but it seems `->withOptions() does merge the values, i.e. mutates the state, so you don't need to return it's value?

Can you try this instead?
```php
$this->app->extend(Factory::class, function (Factory $http, $app) {
    if (config('app.request_proxy')) {
        $http->withOptions([
            'proxy'  => config('app.request_proxy'),
            'verify' => false,
        ]);
    }

    return $http;
});
```
(the first `return` is removed)

TiiFuchs:
Your proposal would not work, because $http is the Factory and with calling withOptions, the __call() method automatically creates a PendingRequest and calls the withOptions on that... The return type of $http->withOptions is a PendingRequest and the provided options only apply to this specific object, not to the Http\Client\Factory itself. So it would not solve the initial intent of setting a proxy.

I tried it in a completely fresh installation of Laravel 9 with the newest laravel-ide-helper. Changed the register() method in AppServiceProvider like this:
```php
    public function register()
    {
        $this->app->extend(\Illuminate\Http\Client\Factory::class, function (\Illuminate\Http\Client\Factory $http) {
            return $http->withOptions([]);
        });
    }
```

and then called `php artisan ide-helper:generate`

PR #1323, opened 2022-02-25
add topclaudy/compoships support

alruhmi:
## Summary
support [topclaudy/compoships](https://github.com/topclaudy/compoships) by adding check in method `ModelsCommand::isRelationNullable()`.

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -777,6 +777,14 @@ class ModelsCommand extends Command
         $fkProp->setAccessible(true);
 
         if ($relation === 'belongsTo') {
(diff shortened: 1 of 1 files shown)


mfn:
Is there a way to avoid having this directly in the library? Like, using hooks or making the code more open/extendable so this can be maintained outside?

### Why?

It's already not easy to handle all the variants you can with models just with laravel and multiple versions and on top of that supporting 3rd, no matter how popular, won't make it easier and it's prune to break stuff. Also it makes the code harder to reason for things barely used.

PR #1330, opened 2022-03-11
Refactor resolving of null information for custom casted attribute types

wimski:
## Summary

Let's say we have the following custom cast:

```php
class MyCustomCast implements CastsAttributes
{
    /**
     * @param Model                $model
     * @param string               $key
     * @param mixed                $value
     * @param array<string, mixed> $attributes
     * @return SomeObject|null
     */
    public function get($model, string $key, $value, $attributes): ?SomeObject
    {
        if ($value === null) {
            return null;
        }

        return new SomeObject($value);
    }

    // setter omitted from example for brevity
}
```

 Now suppose we have two pretty much identical models which use this cast:

```php
class ModelA extends Model
{
    protected $table = 'models_a';

    /**
     * @var array<string, string>
     */
    protected $casts = [
        'casted_column' => MyCustomCast::class,
    ];
}
```

```php
class ModelB extends Model
{
    protected $table = 'models_b';

    /**
     * @var array<string, string>
     */
    protected $casts = [
        'casted_column' => MyCustomCast::class,
    ];
}
```

These models have tables created with the following migrations:

```php
Schema::create('models_a', function (Blueprint $table) {
	$table->id();
	$table->string('casted_column');
	$table->timestamps();
});
```

```php
Schema::create('models_b', function (Blueprint $table) {
	$table->id();
	$table->string('casted_column')->nullable();
	$table->timestamps();
});
```

The only difference between A and B is the `nullable()` setting on the column. Based on that difference I would expect the following properties to be written by the IDE helper:

```php
/**
 * @property SomeObject $casted_column
 */
class ModelA extends Model
```

```php
/**
 * @property SomeObject|null $casted_column
 */
class ModelB extends Model
```

However, both models will receive `SomeObject|null` as the property type, because currently the IDE helper only check the cast class without taking the migration into account. In my opinion the migration should inform the possibility of a `null` value and not the cast, because the cast doesn't know anything about the column it's going to be used for and should always implement `X|null`.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

In my opnion this is a fix for behaviour that was always incorrect, but because it changes behaviour (and outcome), I do consider it a breaking change. I'll let it up to the maintainers how they want to deal with this.

### Checklist
- [ ] Update the README.md

### DIFF
--- a/tests/Console/ModelsCommand/LaravelCustomCasts/__snapshots__/Test__test__1.php
+++ b/tests/Console/ModelsCommand/LaravelCustomCasts/__snapshots__/Test__test__1.php
@@ -28,8 +28,9 @@ use Illuminate\Database\Eloquent\Model;
  * @property \Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\LaravelCustomCasts\Casts\CastedProperty $casted_property_with_return_docblock
  * @property \Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\LaravelCustomCasts\Casts\CastedProperty $casted_property_with_return_docblock_fqn
  * @property array $casted_property_with_return_primitive
- * @property array|null $casted_property_with_return_primitive_docblock
- * @property array|null $casted_property_with_return_nullable_primitive
+ * @property array $casted_property_with_return_primitive_docblock
+ * @property array $casted_property_with_return_nullable_primitive
+ * @property array|null $casted_property_with_return_nullable_primitive_and_nullable_column
  * @property $casted_property_without_return
  * @property \Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\LaravelCustomCasts\Casts\CastedProperty $casted_property_with_param
  * @property SelfCastingCasterWithStaticDocblockReturn $casted_property_with_static_return_docblock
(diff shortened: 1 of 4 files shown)


tests/Console/ModelsCommand/LaravelCustomCasts/__snapshots__/Test__test__1.php
```diff
@@ -28,8 +28,9 @@
  * @property \Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\LaravelCustomCasts\Casts\CastedProperty $casted_property_with_return_docblock
  * @property \Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\LaravelCustomCasts\Casts\CastedProperty $casted_property_with_return_docblock_fqn
  * @property array $casted_property_with_return_primitive
- * @property array|null $casted_property_with_return_primitive_docblock
- * @property array|null $casted_property_with_return_nullable_primitive
+ * @property array $casted_property_with_return_primitive_docblock
+ * @property array $casted_property_with_return_nullable_primitive
```
mfn, line 32: Why is this changed? Is it intentional based on the fix? Thanks

wimski: Yes, this is intentional. I didn't change the migration for these properties and they weren't marked as `nullable()`, so the `|null` gets removed by the fix in this PR.

daniel-de-wit:
@mfn Can you please take another look at this?

barryvdh:
Added to changelog in https://github.com/barryvdh/laravel-ide-helper/commit/aa1aa66ee6cd70acb6c8be6cc0fc2bdee3dc76ba
Thanks!

PR #1338, opened 2022-04-06
Add support for non type-hinted attribute accessors with no backed property

pindab0ter:
## Summary
Even though the documentation doesn't state this explicitly at the time of writing, it is possible to use the new Attribute accessor to create a calculated property where there is no backing property.

This solves #1315.

I added a check for if the Attribute accessor function has no specified type, and then add it without type.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

### Checklist
- [x] Existing tests have been adapted and/or new tests have been added
- [x] Add a CHANGELOG.md entry
- [x] Code style has been fixed via `composer fix-style`

Issue #1293, opened 2022-01-05
Protected/Private Attributes marked methods are not included in the model docBlock


viicslen:
### Versions:
- ide-helper Version: dev-master
- Laravel Version: 8.78.1
- PHP Version: 8.0.13

### Description:
As described in the [pull request](https://github.com/laravel/framework/pull/40022) attribute marked methods should be protected which makes them not discoverable by the the ModelsCommand.

### Steps To Reproduce:

- Create a protected attribute marked method
- Run the `ide-helper:models` command

viicslen:
While tinkering a bit with how Laravel handles these methods I found that we can get all methods even non-public by using `(new ReflectionClass($instance))->getMethods()` instead of `get_class_methods()`

pixelcreation-beheer:
I would love to see this improvement made to be able to adhere to the recommended code style.

pindab0ter:
I would love to see this improvement made to be able to adhere to the recommended code style.

mridul89:
I created PR for this around a month ago- https://github.com/barryvdh/laravel-ide-helper/pull/1325

mfn:
@mridul89 seems I've forgotten about it; my first reaction: 0 tests. Probably that's the reason I wanted to queue it for later and then forgot about it…

pindab0ter:
In general, if you add a feature or fix a bug, you write tests for it to prove that it works since the existing tests won't point that out.

PR #1325, opened 2022-03-04
handle protected methods with the attribute syntax

mridul89:
## Summary

Laravel introduced a new syntax to define attribute "accessors / mutators" in https://github.com/laravel/framework/pull/40022  
Support for this new syntax was added in https://github.com/barryvdh/laravel-ide-helper/pull/1289

However, laravel documentation asks to use protected methods, but #1289 only looked at the public methods. This will also include protected methods.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

### Checklist
- [ ] Add a CHANGELOG.md entry
- [ ] Code style has been fixed via `composer fix-style`

PR #1339, opened 2022-04-06
Add support for protected Attribute accessors

pindab0ter:
## Summary
The [documentation for defining accessors](https://laravel.com/docs/9.x/eloquent-mutators#defining-an-accessor) states that “To define an accessor, create a _protected method_ on your model to represent the accessible attribute.” (emphasis mine).

This is currently not supported by IDE Helper as noted in #1293.

The problem is that `get_class_methods` doesn't return protected or private functions. The solution is to use Reflection instead and filter out any unwanted methods:
- Methods marked as `private` since marking an accessor as `private` breaks the accessor and the two methods defined in 
- Two methods defined in each model through the `\Illuminate\Database\Eloquent\Concerns\HasAttributes` trait.

Any accessors marked as private will now not show up in the IDE Helper files. I chose to not mark this as a breaking change since those accessors would not have worked anyway.

This PR conflicts with #1338.

## Type of change

### Checklist


src/Console/ModelsCommand.php
```diff
@@ -1096,6 +1096,9 @@ protected function hasCamelCaseModelProperties()
 
     protected function getAttributeReturnType(Model $model, \ReflectionMethod $reflectionMethod): Collection
     {
+        // Private/protected ReflectionMethods require setAccessible prior to PHP 8.1
+        $reflectionMethod->setAccessible(true);
```
pindab0ter, line 1100: I chose not to wrap this in an if-statement checking for `PHP_VERSION_ID` since according to the docs from PHP 8.1 onward this method call is a noop.

pindab0ter:
Can this PR be looked at? Using `protected` rather than public is the documented way of using this functionality and prevents accidentally calling the method instead of the magic property.

da-mask (commented):
Looks good to me, but I'm a nobody.

da-mask (approved):
Looks good to me, but I'm a nobody.

mfn:
Will soon check it out; this competes with https://github.com/barryvdh/laravel-ide-helper/pull/1325 right?

pindab0ter:
> Will soon check it out; this competes with #1325 right?

Correct.

pindab0ter:
I would like to ask if someone could look into this again. The summary states the case/necessity for this PR.

There are some fairly major changes to `ModelsCommand.php`, but all tests are green. They also result in that flow being more consistent: Everything now uses `ReflectionClass`/`ReflectionMethod` instead of only the parts that explicitly require that.

danieleambrosino (approved):
Looks perfectly fine to me, hope this will get reviewed and merged by a maintainer soon.
Very good job @pindab0ter!

pindab0ter:
I would like to update this PR, but it relies on #1338. I would like to note that this PR is required to support the Attribute accessors that are new since Laravel 9. Without this PR, they are not recognised when following Laravel's documentation in implementing those.

matiaslauriti:
I am currently using `Attribute`  in some places and it is making me very mad. I truly don't want to make the methods `public`, because that is not how the framework works and just changing them to `public` because we still don't have this fix merged (reported a long time ago) makes me really mad, as if anyone runs the command, we will lose those properties unless they are aware of this "bug" and they change the properties to `public`, run the command, and move them back to `protected`...

That would be unmaintainable... so please, @mfn @barryvdh or any collaborator, member or admin, can this be looked at and merged ASAP?

@pindab0ter I do see there are conflicts (but I can't see them), could you fix them so this can be taking care of immediately?

Would appreciate having this merged and release ASAP because it is a must (`required-dev`) for any Laravel project I work on, so this is truly a gem package for devs

pindab0ter:
I rebased this PR onto #1338 which is in turn rebased on the most recent commit on `main`.

Please review #1338 as this PR relies on it. This is not a massive change and all changes are covered by tests. It would mean a lot to a lot of people if this was accepted.

pindab0ter:
* I changed the visibility of the test methods (I assume this is what you're referring to) because this is what the [Laravel Docs on Attributes](https://laravel.com/docs/10.x/eloquent-mutators#defining-an-accessor) states; "To define an accessor, create a protected method on your model […]." What reason is there for not wanting to change the visibility to what it should be from now on?
* The merge conflicts were resolved and I changed variable usage to reduce clutter as much as possible. I would suggest renaming `method` to `methodName` to prevent possible confusion about the type of the variable once you have OK'd this PR, if that's okay with you.
* I made a few style changes because I felt the code style wasn't in line with the style of the rest of the code in those places. If you disagree, I can revert those changes.

pindab0ter:
Awesome, thanks!

What are your thoughts on this?

> I would suggest renaming `$method` to `$methodName` to prevent possible confusion about the type of the variable […].

Can you do #1338 next? It adds better support for (missing) type hinting of Attributes.

Issue #1120, opened 2020-12-08
Database fields not written to _ide_helper_models.php


ionesculiviucristian:
### Versions:
- ide-helper Version: 2.8.2
- Laravel Version: 8.18.1
- PHP Version: 8.0.0
- MySQL Version: 8+

(using Docker and Laravel Passport, if it matters)
### Description:

When running ```php artisan ide-helper:models --nowrite``` the model ```User.php``` located in ```app/Models``` doesn't write the database fields, only the default properties/methods. 
In ```Barryvdh\LaravelIdeHelper\Console\ModelsCommand.php``` the lines:

```php
        if (strpos($table, '.')) {
            [$database, $table] = explode('.', $table);
        }
```
expect the ```$table``` to have the database name in but for me ```$table``` is ```'users'``` and as such returns here

```php
        $columns = $schema->listTableColumns($table, $database);

        if (!$columns) {
            return;
        }
```

mfn:
@ionesculiviucristian is this a recent regression for your? Or did it newer work / new project?

If a regression, can you check out older ide-helper / Laravel versions?

thank you!

ionesculiviucristian:
Tried with 2.8.1 (lowest with Laravel 8 support) and 2.8.2. Same result.

mfn:
Your config looks exactly like mine, I'm on PHP 7.4.13 with Laravel 8.17.2 and ide-helper 2.8.1/2.8.2 but work with and without `--nowrite`.

mfn:
Do you have doctrine/dbal installed?

ionesculiviucristian:
@mfn No. i don't but for whatever reason, today, after doing a composer update, it just worked. I guess the issue can be closed.

I see this package requires ```"doctrine/dbal": "~2.3"```. Is that not for getting the database fields?

mfn:
It is, basically doctrine/dbal is required for the feature you expected to work :}

Do you still feel we can close this? I'm not sure I entirely follow how it was suddenly fixed for your and we may still have something to improve?

ossicoinc:
I can confirm this works - it doesn't support tables that are in a different schema than dbo

In model: 
`$table = 'sc.MyTable';`

Thanks for the workaround!

SimplyCorey:
https://github.com/barryvdh/laravel-ide-helper/pull/1349 should fix the schema issue.

developh:
I have tried to add schema to table name in model `__construct`
`
```
[...]

PR #974, opened 2020-06-25
Support database.schema.table in \Illuminate\Database\Eloquent\Model::$table

mfn:
## Summary

The current code handles `database.table` but fails when
`database.schema.table` is present, as it populates
`$database = 'database'; $table = 'schema';` and "drops"
 the table entirely.

There were two previous attempts for PgSQL [1] and MsSQL [2] with two
different approaches; this is yet another one with the goal to be simple
and not being too dependent on which driver is used.

Other databases don't support this feature (AFAIK) but IMHO a driver
specific check shouldn't be necessary; KISS.

### Links
- [1] https://github.com/barryvdh/laravel-ide-helper/pull/392
- [2] https://github.com/barryvdh/laravel-ide-helper/pull/781
- Fixes https://github.com/barryvdh/laravel-ide-helper/issues/780
- Fixes https://github.com/barryvdh/laravel-ide-helper/issues/483

PR #1349, opened 2022-05-06
Allow for proper database resolve with schemas

SimplyCorey:
## Summary
Hello,

Thanks for the package.  It's very helpful.

This PR fixes an issue with the model IDE helper generation command explained in https://github.com/barryvdh/laravel-ide-helper/issues/1120.  The command does not work with databases that use schemas such as postgres.  This is from the way the `explode` is computed to calculate the `database` and `table` name.  The `explode` needs to be restricted to the first instance of `.`.

Old Behavior (Bugged)

| Table String         | Database | Table     |
|----------------------|----------|-----------|
| `cats`               | null     | `cats`    |
| `mysql.cats`         | `mysql`  | `cats`    |
| `pgsql.animals.cats` | `pgsql`  | `animals` |

New Behavior:

| Table String         | Database | Table          |
|----------------------|----------|----------------|
| `cats`               | null     | `cats`         |
| `mysql.cats`         | `mysql`  | `cats`         |
| `pgsql.animals.cats` | `pgsql`  | `animals.cats` |

I'm not sure how to write a test for this as the testing suite uses sqlite and I don't believe schemas operate like this there.

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -484,7 +484,7 @@ class ModelsCommand extends Command
 
         $database = null;
         if (strpos($table, '.')) {
-            [$database, $table] = explode('.', $table);
+            [$database, $table] = explode('.', $table, 2);
         }
 
         $columns = $schema->listTableColumns($table, $database);
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -463,6 +463,7 @@ class ModelsCommand extends Command
      */
     public function getPropertiesFromTable($model)
     {
+        $database = $model->getConnection()->getName();
         $table = $model->getConnection()->getTablePrefix() . $model->getTable();
         $schema = $model->getConnection()->getDoctrineSchemaManager();
         $databasePlatform = $schema->getDatabasePlatform();
@@ -480,11 +481,6 @@ class ModelsCommand extends Command
                 throw $exception;
             }
             $databasePlatform->registerDoctrineTypeMapping($yourTypeName, $doctrineTypeName);
-        }
-
-        $database = null;
-        if (strpos($table, '.')) {
-            [$database, $table] = explode('.', $table, 2);
         }
 
         $columns = $schema->listTableColumns($table, $database);


src/Console/ModelsCommand.php
```diff
@@ -484,7 +484,7 @@ public function getPropertiesFromTable($model)
 
         $database = null;
         if (strpos($table, '.')) {
-            [$database, $table] = explode('.', $table);
+            [$database, $table] = explode('.', $table, 2);
```
mfn, line 487: So, my nemesis (the code, not you) is appearing again.

I had a fix for this some years ago, but it failed properly work in all environments -> https://github.com/barryvdh/laravel-ide-helper/pull/974/files

Although my code looks, err..., weird, it actually does the same 😬

Yours:
```
>>> explode('.', 'foo.bar', 2);
=> [
     "foo",
     "bar",
   ]
>>> explode('.', 'foo.bar.baz', 2);
=> [
     "foo",
     "bar.baz",
   ]
>>>
```
Mine:
```
>>> $table = 'foo.bar'; strtok($table, '.');
=> "foo"
>>> strtok('');
=> "bar"
>>>
>>> $table = 'foo.bar.baz'; strtok($table, '.');
=> "foo"
>>> strtok('');
=> "bar.baz"
```

Obviously your solution is smarter 💪🏼 

But in the end, it's the same outcome and [then we had the issue reported it does not work with MSSQL](https://github.com/barryvdh/laravel-ide-helper/pull/974#issuecomment-659565923) (which I don't have to test), so I had to drop it.

SimplyCorey: hmm, wasn't aware of the MSSQL issue 😕 .  I took another stab at this and I think it's not necessary to even attempt to compute the `$database` and `$table` from exploding but rely on the connection information altogether.  I just pushed what I think might be a better approach to the PR.

I don't have access to MSSQL to test this.  This seems to work fine on pgsql/mysql/sqlite.

thijsvdanker: I can confirm it works great on mssql!

SimplyCorey: @mfn Sorry to bother, but anything else you want me to do here?

mfn: @SimplyCorey I don't think so, but I haven't had time to come back to the issue!

barryvdh:
So it looks like the explode isn't needed anymore?

barryvdh (approved):

PR #1352, opened 2022-05-22
Fix issue where \Eloquent is not included when using write_mixin

### LINKED ISSUE #1299: Removing extends from generated class breaks autocomplete
### Versions:
- ide-helper Version: 2.11.0
- Laravel Version: 8.79.0
- PHP Version: 8.0

### Description:

https://github.com/barryvdh/laravel-ide-helper/pull/1268 changed mixin mode to not add an extends to the generated class. This breaks the autocomplete for functions like `Model->find()` as my classes no longer extend `\Eloquent`

### Steps To Reproduce:

- Run `ide-helper:models -M`
- Try to use the `find()` function on any model and see that the return type is listed as mixed instead of the proper class

### LINKED ISSUE #1342: Autocompletion for Model methods is not avaliable
### Versions:
- ide-helper Version: 2.12.2
- Laravel Version: 9.4.1
- PHP Version: 8.1

### Description:
Autocompletion is not available for some methods.

image

Commands used to generate helpers:
```
  artisan ide-helper:generate
  artisan ide-helper:models -M
  artisan ide-helper:meta
```

### LINKED ISSUE #1343: Model methods are missing with `--write-mixin`
### Versions:
- ide-helper Version: 2.12.3
- Laravel Version: 9.9.0
- PHP Version: 8.1.3

### Description:

Model methods are missing when using `--write-mixin`

| With `--write-mixin` | Without `--write-mixin` |
|---|---|
|  |  |

Basic model methods like `find` appear to be missing when using `--write-mixin`. They work correctly without `--write-mixin`

I've reproduced the following steps in this repo: https://github.com/ptrkcsk/laravel-ide-helper-models

There are two differences in `_ide_helper_models.php` between using `--write-mixin` and not, as seen here: https://github.com/ptrkcsk/laravel-ide-helper-models/compare/write-mixin...no-write-mixin

1. The generated class has `@mixin IdeHelperUser`. This looks like a mistake to me because `IdeHelperUser` appears to not exist
1. The generated class extends `\Eloquent`. I suspect this is the problem, because with `--write-mixin`, the generated class is missing the methods provided by `\Eloquent`

### Steps To Reproduce:

1. Generate facades helper and models helper with `--write-mixin`:

    ```sh
    vendor/bin/sail artisan ide-helper:generate
    vendor/bin/sail artisan ide-helper:models --write-mixin
    ```
   
1. Try to use a model's `find` method, e.g. `User::find()`. Notice "Method 'find' not found..." warning in IDE

1. Re-generate models helper without `--write-mixin`:

    ```sh
    vendor/bin/sail artisan ide-helper:models
    ```
   
1. Try to use a model's `find` method again. Notice `find` method is now found

Jefemy:
## Summary
If you write your models using the `--write-mixin` option the included docblock generated on the model will include a mixin. Previously the code replaced here would ignore the `@mixin \Eloquent` addition if any mixin existed on the model. This would break the code suggestions as on first write the Eloquent mixin would be added but after that it wouldn't be generated as its detected as 'already existing'. Info including the lines and commits where I found this are discussed in #1299 

This fix changes it to a loop which removes any mixins that matches the Eloquent class before appending it.

Also fixes #1342, fixes #1343, fixes #1299

## Type of change

### Checklist
- [ ] Update the README.md


FatBoyXPC:
Will this end up writing `@mixin \Eloquent` to the model class, or to the generated `_ide_helper_models.php` classes?

Jefemy:
> Will this end up writing `@mixin \Eloquent` to the model class, or to the generated `_ide_helper_models.php` classes?

It will put it in the generated `_ide_helper_models` class. This pull request changes no base functionality in the package, it just fixes an issue where on a second call of the `ide-helper:models --write-mixin` it will create a different output.

Jefemy:
> For the temporary work around, do you mean delete the `@mixin IdeHelperFoo` from the `Foo` model?

Yes. In the current version any mixins existing on the original model will cause `\Eloquent` to be excluded from the generated class.

dnmarques:
@barryvdh any plans to merge this pull request?

barryvdh (approved):

PR #1388, opened 2022-11-09
Fix Castables issue #1312

kwarcu:
## Summary
Fix for issue with Castables described in https://github.com/barryvdh/laravel-ide-helper/issues/1312
If type of a `Castable` can not be determined from return type of get method from `castUsing()`->`CastsAttributes` then the original `Castable` class should be used.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

### Checklist
- [x] Existing tests have been adapted and/or new tests have been added
- [x] Code style has been fixed via `composer fix-style`
- [x] Add a CHANGELOG.md entry

Issue #1312, opened 2022-02-07
Around v2.12.0 @property for custom Castables stopped working


kwarcu:
In v2.11.0, model like this

```php
class Job extends Model
{
    protected $fillable = [
        'status',
    ];

    protected $casts = [
        'status' => JobStatus::class,
    ];
[...]
```

would generate  phpdoc `@property JobStatus $status`, but since v2.12.0 it keeps generating `@property mixed $status`

Exact command I'm using is `php artisan ide-helper:models --write --smart-reset`

It was fine in v2.11.0, currently using this exact version locked

innocenzi:
Not sure if it's related, but using https://github.com/spatie/laravel-data, a similar issue happens:

```php
protected $casts = [
    'legacy' => LegacyUserData::class,
];
```

The generated property is `@property \Spatie\LaravelData\Data|null|null $legacy` instead of `@property Domain\Legacy\DTO\LegacyUserData|null $legacy`

seanmtaylor:
Same here. We're still on `v2.11.0` for this very reason. All of our custom casts are being changed to `mixed`.

mfn:
Is this still an issue with the latest version?

I think https://github.com/barryvdh/laravel-ide-helper/pull/1306 is the reason, but there was a hotfix release shortly afterwards…

kwarcu:
> Is this still an issue with the latest version?

Just checked v2.12.3 and issue still exists.

kwarcu:
v2.12.3 is still the latest version and issue still exists for latest Laravel version.

aaronflorey:
I'm not quite sure on the rules of accepting code here, but would the maintainers accept a PR that hard codes in a check for LaravelData?

```
        if(get_parent_class($type) === 'Spatie\LaravelData\Data') {
            return $type;
        }
```

kwarcu:
> Anybody found a way to fix that?

Still nothing. I never got to debugging the lib itself though. FWIW problem still exists on latest version 🤷‍♂️

PR #1400, opened 2023-01-02
Property type isn't detected when casting Eloquent attributes to `spatie/laravel-data` Data objects

code-distortion:
## The problem

When casting Eloquent attributes to `spatie/laravel-data` Data objects, ide-helper doesn't pick the correct class. For example, the example from Spatie's documentation https://spatie.be/docs/laravel-data/v2/advanced-usage/eloquent-casting:

```php
<?php
# app/Models/Song.php
namespace App\Models;

use App\SpatieData\ArtistData;

class Song extends Model
{
    protected $casts = [
        'artist' => ArtistData::class,
    ];
}
```

```php
<?php
# app/SpatieData/ArtistData.php
namespace App\SpatieData;

use Spatie\LaravelData\Data;

class ArtistData extends Data
{
    public function __construct(
        public string $name,
        public int $age,
    ) {
    }
}
```

Running `php artisan ide-helper:models --dir='app' --write` generates this DocBlock:

```php
/**
 * App\Models\Song
 *
 * @property \Spatie\LaravelData\Contracts\BaseData|null $artist // ❌ not the desired prop definition
 * @method static \Illuminate\Database\Eloquent\Builder|Song newModelQuery()
 * @method static \Illuminate\Database\Eloquent\Builder|Song newQuery()
 * @method static \Illuminate\Database\Eloquent\Builder|Song query()
 * @mixin \Eloquent
 */
```

This means that we don't get the benefit of type-hinting for casted attributes like this:

```php
$songModel->artist->name;
```

---

## Why does this happen?

This happens because `src/Support/EloquentCasts/DataEloquentCast.php` from Spatie's package specifies a return type. *This* is what ide-helper picks up:

```php
# src/Support/EloquentCasts/DataEloquentCast.php in spatie/laravel-data
…
    public function get($model, string $key, $value, array $attributes): ?BaseData
    {
…
```

---

## Proposed change

I put this together when looking through the code, trying to work out how to fix it in a project of mine. I thought I'd turn it into a PR for discussion…

This PR lets you instruct ide-helper to use the class *that was specified*, by adding the DocBlock tag `@ide-helper-eloquent-cast-to-specified-class` to the class being cast to. e.g.

```php
<?php
# app/SpatieData/ArtistData.php

/**
 * @ide-helper-eloquent-cast-to-specified-class
 */
class ArtistData extends Data
{
…
```

From which, ide-helper now generates:

```php
/**
 * App\Models\Song
 *
 * @property ArtistData $artist // ✅ the desired prop definition
 * @method static \Illuminate\Database\Eloquent\Builder|Song newModelQuery()
 * @method static \Illuminate\Database\Eloquent\Builder|Song newQuery()
 * @method static \Illuminate\Database\Eloquent\Builder|Song query()
 * @mixin \Eloquent
 */
```

---

## Notes

- The `@ide-helper-eloquent-cast-to-specified-class` tag can be added to a class, or any of it's parents.
- `spatie/laravel-data` could add this tag to their `Spatie\LaravelData\Data` class, so it's always applied to Data classes that use it. But if they don't, it can just as easily be added to Data classes (or a parent class) in individual projects.
- This solution isn't specific to `spatie/laravel-data`. It can be used for any classes being cast to in this manner.

## Thoughts

- I couldn't think of a way for it to happen *automatically* without forcing it in *every* case. Which I don't think is 100% desired.
- I don't think having to add a @tag is a "nice" solution, however it does give control over when it happens.
- An alternative I considered was to specify the relevant classes inside ide-helper's config file. But this doesn't seem very flexible.
- I'd be interested to hear if anybody has other ideas about how to detect when it should occur.

## Other issues and PRs

- This PR may help address this issue https://github.com/barryvdh/laravel-ide-helper/issues/1312
- The https://github.com/barryvdh/laravel-ide-helper/pull/1388 PR addresses a slightly different situation. It falls back to the specified-class when none was specified by the casting class. It is short and elegant. However `spatie/laravel-data` *does* specify a return type. So whilst it might be a useful change, it doesn't affect *this* situation.
- The comment https://github.com/barryvdh/laravel-ide-helper/issues/1312#issuecomment-1221270878 suggests hard-coding a check for `spatie/laravel-data`. This seems inflexible to me as it wouldn't allow for other packages or custom classes to benefit.

---

## Type of change

### Checklist
- [X] Add a CHANGELOG.md entry
- [ ] Update the README.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 - Add support for custom casts that implement `CastsInboundAttributes` [#1329 / sforward](https://github.com/barryvdh/laravel-ide-helper/pull/1329)
+- A solution to detect property types when casting Eloquent attributes to objects. Add the `@ide-helper-eloquent-cast-to-specified-class` tag to the class being cast to
 
 2022-03-06, 2.12.3
 ------------------
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -935,14 +935,14 @@ class ModelsCommand extends Command
 
             // remove the already existing tag to prevent duplicates
             foreach ($phpdoc->getTagsByName('mixin') as $tag) {
-                if($tag->getContent() === $eloquentClassNameInModel) {
+                if ($tag->getContent() === $eloquentClassNameInModel) {
                     $phpdoc->deleteTag($tag);
                 }
             }
 
             $phpdoc->appendTag(Tag::createInstance('@mixin ' . $eloquentClassNameInModel, $phpdoc));
         }
-        
+
         if ($this->phpstorm_noinspections) {
             /**
              * Facades, Eloquent API
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -1211,6 +1211,30 @@ class ModelsCommand extends Command
         return $type;
     }
 
+    /**
+     * Check to see if a class (or its parents) has a @ide-helper-eloquent-cast-to-specified-class tag
+     *
+     * @param \Reflector $reflector
+     *
+     * @return boolean
+     */
+    protected function classHasCastSpecifiedTag(\Reflector $reflector = null): bool
+    {
+        $phpDocContext = (new ContextFactory())->createFromReflector($reflector);
+        $context = new Context(
+            $phpDocContext->getNamespace(),
+            $phpDocContext->getNamespaceAliases()
+        );
+        $phpdoc = new DocBlock($reflector, $context);
+
+        if ($phpdoc->hasTag('ide-helper-eloquent-cast-to-specified-class')) {
+            return true;
+        }
+
+        $parentReflector = $reflector->getParentClass();
+
+        return $parentReflector ? $this->classHasCastSpecifiedTag($parentReflector) : false;
+    }
 
     /**
      * Generates methods provided by the SoftDeletes trait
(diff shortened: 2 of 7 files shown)


jameshulse:
Fantastic. We are running into the same issue now. I guess the other alternative would be to give laravel-ide-helper special handling for the `spatie/laravel-data` package. But I think your solution is still workable.

barryvdh:
Can we not check if the actual cast is extending the type, and in that case return actual cast (same as fallback $type from #1312?

rolandverner:
any updates?

scajal:
Bump

igorleszczynski:
Bump

igorleszczynski:
Temporary solution is that you should make custom cast like this:

```php
<?php

namespace App\Modules\Bdo\Mutators;

use App\Modules\Bdo\Data\BdoLastUserSyncFilters;
use Illuminate\Contracts\Database\Eloquent\CastsAttributes;

class BdoLastUserSyncFiltersCast implements CastsAttributes {

    public function get(Model|BdoLastUserSyncFilters $model, string $key, mixed $value, array $attributes): null|BdoLastUserSyncFilters
    {
        if($value instanceof BdoLastUserSyncFilters) {
            return $value;
        }
        if (!$value) {
            return null;
        }
        return BdoLastUserSyncFilters::from($value);
    }

    public function set(Model|BdoLastUserSyncFilters $model, string $key, mixed $value, array $attributes)
    {
        return $model->toJson();
    }
}
```


Add this into your model:

```php
    protected function casts(): array
    {
        return [
            'filters' => BdoLastUserSyncFiltersCast::class,
        ];
    }
```

Gregory-Gerard:
If anyone using `@mixin` (if not, it may override the doc block, but I didn't tested), you can just add a `@property` to override with the good one:

```php
<?php

/**
 * @property SettingsData $settings
 * @mixin IdeHelperSetting
 */
final class Setting extends Model { }
```

PR #1432, opened 2023-03-24
Add Throwable exception catch

### LINKED ISSUE #1431: Alias loading Missing Throwable exception catch
### Versions:
- ide-helper Version: 2.3.10
- PHP Version: 8.1

### Description:

At some point, loading Alias may cause a Throwable error due to some incorrect configuration. I recommend adding the catch of Throwable when Alias' detectRoot catches exceptions so that helper files can be generated properly

dongm2ez:
## Summary
fix #1431

## Type of change


barryvdh:
I cannot do that anymore. is it is deleted.

PR #1446, opened 2023-06-12
fix: cant auto complete method chaining

SocolaDaiCa:
## Type of change

### Checklist
- [ ] Update the README.md

Before:

image

image

After:

image

image


SocolaDaiCa:
@mfn yup, The difference being the static, because `::` only hint static method, So for each method I add a corresponding static method

SocolaDaiCa:
@mfn I'm sorry I was a little confused in the previous explanation, I'll explain again

I define scopeActive in the User model
Currently ide-helper is `@anotation` scope method as static
when i type `User::ac` => it suggests me `User::active`
but when I type `User::query()->ac` it won't suggest me the `active` method, or when I type `User::query()->where(...)->ac` it won't suggest the method either activate me
Therefore, I added a non-static method for better suggestion and combined with method chaining


Perhaps the best way is to remove static in the ide-helper `@anotation` instead of adding both static and non-static methods

SocolaDaiCa:
> I don't see how we can do this in the current form without causing collateral issues.

Because all the @anotations that ide-helper creates for the model are scope and where, they certainly work in both static and non-static forms.
I can't think of a case where it only works in static form

and laravel also recommends using the form `Model::query()-><method1>()-><method2>()` instead of the form `Model::<method1>()-><method2>()`.

So I think changing `@anotation` scope and where from static to non static is reasonable and safe

SocolaDaiCa:
@mfn Please help me check again

I've fixed it, non-static is now only applied to function `scope` and `where<column>`

If anything is not right, please let me know

image

barryvdh:
Can you try with the latest dev-version? `3.4.x@dev` or dev-master

Seems to work fine for me.

Screenshot 2024-12-29 at 11 56 23@2x

SocolaDaiCa:
@barryvdh It not work for me, whereName is still static
image

barryvdh:
WhereName is supposed to be static

SocolaDaiCa:
@barryvdh I know, but I think WhereName should support both static and non static. Example

```php
User::whereName()->get()


User::whereEmail()->whereName()->get()
```

GNfsys:
Laravel Idea does the following when generating model helper files, I don't know if it's helpful in this case:

```
class _BaseBuilder extends \Illuminate\Database\Eloquent\Builder {}

/**
* @method _IH_User_QB active()
*/
class _IH_User_QB extends _BaseBuilder {}
```

Which then allows to use both:
`User::active()` and `User::query()->active();`

PR #1461, opened 2023-08-21
fix belongsTo relation is detected as nullable.

snmatsui:
## Summary

In ModelsCommand, belongs to relations are detected as nullable when this column is nullable or has not foreign key constraint.

https://github.com/barryvdh/laravel-ide-helper/blob/master/src/Console/ModelsCommand.php#L841-L842

Now read foreign keys is not reset on model loop.
For example. when column x of model A has no foreign key and column x of model B has foreign key, column x of model A may marked as not nullable because column x is recoreded as foreign key by model B.

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -266,6 +266,7 @@ class ModelsCommand extends Command
             }
             $this->properties = [];
             $this->methods = [];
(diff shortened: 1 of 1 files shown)


barryvdh (approved):

barryvdh:
Thanks, seems like an oversight indeed.

PR #1464, opened 2023-08-28
Add support for enum default arguments using enum cases

### LINKED ISSUE #1463: Default values as enums cases
### Versions:
- ide-helper Version: 2.12.3
- Laravel Version: 10.19.0
- PHP Version: 8.2.8

### Description:

In case we want to add a default value using an enum typed argument on a model query scope like so:

```php
<?php

class MyModel extends Model
{
    public function scopeHasAbility(Builder $query, Authenticatable $user, ?CrudAction $ability = CrudAction::Read)
    {
        // code here...
    }
}
```

This will run fine in PHP 8+ but IDE helper models docblock generator command will complain with the following error:

image

Error comes from this line: https://github.com/barryvdh/laravel-ide-helper/blob/0801d002e48d7ead2b0a0169c04f44f72800738a/src/Console/ModelsCommand.php#L1103

### Steps To Reproduce:

- Add code from description
- Run `php artisan ide-helper:models -w`

d8vjork:
## Summary

Fixes #1463 

## Type of change

### Checklist
- [x] Update the README.md


d8vjork:
@mfn Will try, don't know if there are tests for these default values, I was looking around and didn't found something I can adapt to this, or do you think I should add a new folder instead?

d8vjork:
Ok I reply myself, I think this should go in a new folder so we can mark it as PHP 8+ only as these enums won't work on PHP 7 tests

.github/workflows/run-static-analysis.yml
```diff
@@ -17,7 +17,7 @@ jobs:
             - name: Setup PHP
               uses: shivammathur/setup-php@v2
               with:
-                php-version: 8.0
+                php-version: 8.1
```
barryvdh, line 20: Otherwise the UnitEnum does not exist

barryvdh (approved):

PR #1472, opened 2023-10-05
Fix flag reference in README

gregkos:
## Summary
Following the README and trying to set up a config with mixins I got an error. After digging through I realized the option is actually `--write_mixins` and not `--write-mixin` as suggested by the README. I've updated the file to reflect that.

## Type of change
- [X] Misc. change (internal, infrastructure, maintenance, etc.)

### Checklist
- [X] Update the README.md


gregkos:
Continuing to dig through, the confusion appears to stem from the generate command having `--write_mixins` available and the models command having `--write-mixin` available. I think this PR is wrong, but I believe a fix is warranted. I'd be happy to implement but I am not sure what the correct path is at this point.

Issue #838, opened 2019-10-23
error when using composite key package


MartinP7r:
#706 seems related

I'm using `awobaz/compoships` (https://github.com/topclaudy/compoships) for a relationship that uses composite keys in one of my models, which seems to break the generation script for models as it throws the following error:

```
Exception: Illegal offset type in isset or empty
Could not analyze class App\ClassName.
```

the breaking composite key relationship looks like this:

```php
public function department()
{
    return $this->hasOne(
        Department::class,
        ['company_id', 'department_id'],
        ['company_id', 'department_id']
    );
}
```

I realize that you can't account for 3rd party packages, but would there be a workaround way to at least exclude the relationship from scanning?

mpyw:
@MartinP7r 

Quick fix for this issue (requires `barryvdh/laravel-ide-helper^2.7` `php>=7.4`):

```php
<?php

namespace App\Console\Commands\IdeHelper;

use Barryvdh\LaravelIdeHelper\Console\ModelsCommand as BaseModelsCommand;
use Illuminate\Support\\Arr;
use Illuminate\Database\Eloquent\Relations\Relation;

class ModelsCommand extends BaseModelsCommand
{
    /**
     * @param  string   $relation
     * @param  Relation $relationObj
     * @return bool
     */
    protected function isRelationNullable(string $relation, Relation $relationObj): bool
    {
        $reflectionObj = new \ReflectionObject($relationObj);

        if (in_array($relation, ['hasOne', 'hasOneThrough', 'morphOne'], true) || !$reflectionObj->hasProperty('foreignKey')) {
            return parent::isRelationNullable($relation, $relationObj);
        }

        $fkProp = $reflectionObj->getProperty('foreignKey');
        $fkProp->setAccessible(true);

        return (bool)Arr::first(
            (array)$fkProp->getValue($relationObj),
            fn (string $value) => isset($this->nullableColumns[$value])
        );
    }
}

```

burut13:
Spostovani,
do torka, 4.8.2020 sem odsoten. Na posto bom odgovoril glede na pomembnost zadeve takoj, ko bo mogoce.
Za nujne zadeve pisite na urgent@u-cart.si ali poklicite +386 40 847 061, Damijan.
Hvala za razumevanje in lep pozdrav,
Borut


Hi there,
I'll be out of office until August 4th 2020. I will reply to your e-mail when I get a chance.
[...]

Issue #706, opened 2018-09-12
Support for composite primary keys


sptdigital:
When declaring an Eloquent model with a composite primary key the following error is always returned upon running of the Models helper ...

```
Exception: Illegal offset type in isset or empty
```

Do you plan on adding support for composite primary keys?

isclopezm:
I have the same problem. I'm using the package _awobaz/compoships_, which allows using composite primary keys and when I run the _ide-helper:models_ command I get that error.

Mochaka:
I'm also getting this error on the same package, is there any fix?

Mochaka:
i patched it by modifying a line in `vendor/barryvdh/laravel-ide-helper/src/Console/ModelsCommand.php:552`

change

`return isset($this->nullableColumns[$fkProp->getValue($relation)]);`

to 

` return isset($this->nullableColumns[is_array($fkProp->getValue($relation)) ? $fkProp->getValue($relation)[0] : $fkProp->getValue($relation)]);`

it doesn't really fix it, it just stops the error coming up

burut13:
Anything new on this issue?

I'd like to mention that it only occurs with `belongsTo` relationship, while working just fine with `hasMany`. Haven't tested other relationship types.

deare:
The error occurs when you use a model that has a composite primary key.
In the case of composite primary keys, also write in your model:
_public $incrementing = false;_

mpyw:
@sptdigital @isclopezm @aaronflorey @burut13 @deare (CC @barryvdh)

Quick fix for this issue (requires `barryvdh/laravel-ide-helper^2.7` `php>=7.4`):

```php
<?php

namespace App\Console\Commands\IdeHelper;

use Barryvdh\LaravelIdeHelper\Console\ModelsCommand as BaseModelsCommand;
use Illuminate\Support\Arr;
use Illuminate\Database\Eloquent\Relations\Relation;

class ModelsCommand extends BaseModelsCommand
{
    /**
     * @param  string   $relation
     * @param  Relation $relationObj
     * @return bool
     */
    protected function isRelationNullable(string $relation, Relation $relationObj): bool
    {
        $reflectionObj = new \ReflectionObject($relationObj);

        if (in_array($relation, ['hasOne', 'hasOneThrough', 'morphOne'], true) || !$reflectionObj->hasProperty('foreignKey')) {
            return parent::isRelationNullable($relation, $relationObj);
        }

[...]

PR #1479, opened 2023-11-06
Add support for composite keys

### LINKED ISSUE #1474: Incorrect PHPDoc being written to model for BelongsTo relation
### Versions:
- ide-helper Version: 2.13.0
- Laravel Version: 10.x
- PHP Version: 8.2.x

### Description:
Hello! 

I'm currently writing the PHPDocs to a few models and I noticed the wrong doc was being generated for the `BelongsTo` relation (and possibly other relations that return single models):

```php
/** @property-read \Illuminate\Database\Eloquent\Collection<int, \App\Models\User> $createdByUser */
```

However, a `BelongsTo` relation *never* returns a collection so this instead should be:

```php
/** @property-read null|\App\Models\User $createdByUser */
```

This is causing PHPStan Level 2 to fail with:

```
  27     Access to an undefined property Illuminate\Database\Eloquent\Collection<int, App\Models\User>::$name.                                   
```
when trying to call `$model->createdByUser->name`

### Steps To Reproduce:

- Create model with `BelongsTo` relation defined
- Write PHPDocs to model

Thanks!

calebdw:
Fixes #1474, https://github.com/topclaudy/compoships/issues/77

## Summary

Hello!

This PR adds support for relationships with composite keys. There have been a couple of issues over the years about this but this still has not been resolved (I don't want to have to extend the Artisan command in my app):

- https://github.com/barryvdh/laravel-ide-helper/issues/838
- https://github.com/barryvdh/laravel-ide-helper/issues/706

There's even an open issue on the Compoship side of things (https://github.com/topclaudy/compoships/issues/77) but as the maintainer stated, there really nothing to do on that end.

I understand Laravel does not support composite keys, however, people do use them and this package already provides the ability to define non-Laravel relations so it would be much easier to support this rather than to expect everyone who uses composite keys to have to extend the Artisan command with the same custom logic.

Thanks!

## Type of change

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 - Add support for enum default arguments using enum cases. [#1464 / d8vjork](https://github.com/barryvdh/laravel-ide-helper/pull/1464)
 - Add support for real-time facades in the helper file. [#1455 / filipac](https://github.com/barryvdh/laravel-ide-helper/pull/1455)
+- Add support for relations with composite keys. [#1479 / calebdw](https://github.com/barryvdh/laravel-ide-helper/pull/1479)
 
 2024-02-05, 2.14.0
 ------------------
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -322,7 +322,8 @@ return [
     |
     | When using custom relation types its possible for the class name to not contain
     | the proper return type of the relation. The key of the array is the relationship
-    | method name. The value of the array is the return type of the relation.
+    | method name. The value of the array is the return type of the relation ('many'
+    | or 'morphTo').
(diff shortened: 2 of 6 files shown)


calebdw:
@mfn, would it be possible to review this PR please?

Thanks!

barryvdh:
Can we have a test for this?

barryvdh (changes requested):
Can you add some tests?

barryvdh:
Do we need that package to test it? Or can we replicate a simple case here?

calebdw:
I'm working on a test at the moment, I don't think we need the package

calebdw:
Test has been added

barryvdh:
Thanks!

calebdw:
Thank you!!

PR #1486, opened 2023-11-28
disable inspections of helper files

eidng8:
## Summary
Currently PS inspects all helper files and screems... This PR makes sure it doesn't inspect helper files.

## Type of change

### DIFF
--- a/resources/views/helper.php
+++ b/resources/views/helper.php
@@ -8,6 +8,7 @@
  */
 ?>
 
+/* @noinspection ALL */
 // @formatter:off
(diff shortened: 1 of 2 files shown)


barryvdh:
But you need the files to be inspected? Or which tools does this block?

eidng8:
It doesn't break anything. It's just that when performing inspection only whole project using PHPStorm, these helper files are insepcted too. Although inspecting the whole project this way results in inspection issues, it doesn't break anything in anyway.

barryvdh (approved):

PR #1510, opened 2024-02-07
Add support for Laravel 11

### LINKED ISSUE #1509: Modifications required to support Laravel 11
Hello, Laravel 11 will be released soon.

## Overview

I am using the `11.x-dev` branch under development to check the operation of Laravel IDE Helper, but there are some places where it does not work due to breaking changes on the Laravel side.

This version update will require modifications not only to `composer.json` and CI, but also to the implementation. Here is a summary of them.

### PHPUnit version upgrade

`orchestra/testbench: ^9`, which is required for testing with Laravel 11, requires `phpunit/phpunit: ^10.5`. We need to widen the version range with `require-dev`.

### Support for Doctrine DBAL removal

Related: https://github.com/laravel/framework/pull/48864

All functionality related to Doctrine DBAL has been removed. Not only does Laravel itself not use it, **public methods such as `getDoctrineSchemaManager()` have been removed**, and classes related to `Connection` have also been significantly refactored.

Laravel IDE Helper depends on many of these features, so when supporting Laravel 11 we will need to re-implement it using a different method.

### Change migration method for floating-point types

Related: https://github.com/laravel/framework/pull/48861

Some of migration methods for setting columns using a floating-point type has been changed.

The Laravel IDE Helper implementation code is not involved in the migration method, but it is referenced from the test code. Correcting is required for the relevant parts.

### Adding methods to `Authenticatable` contract

Related: https://github.com/laravel/framework/pull/48665

Implementation of `rehashPasswordIfRequired()` is now required. It is referenced as `User implements Authenticatable` in the test code, so it needs to be modified to be contract compliant.

KentarouTakeda:
## Summary

This pull request resolves #1509.

## Type of change

### Checklist
- [ ] Update the README.md

## Overview of responses to each issue

* [x] PHPUnit version upgrade

  I have expanded the version range accordingly. There was no corresponding breaking change in PHPUnit 10, so there is no need to modify the test code.

* [x] Support for Doctrine DBAL removal

  I have migrated the minimal necessary code for the Laravel IDE Helper from the implementation existing in Laravel 10, which was removed in Laravel 11.

  Ideally, it should be implemented in a newly created Laravel API with the removal of Doctrine DBAL, but the version with that modification would be `laravel/framework > 10` only. We'll need to support `<=10` for a while, so this is the way to go for now.

* [x] Migration method changes for floating-point types

  I have removed the corresponding migration method from the test. Only `$table->unsignedDecimal()` was applicable.

  I don't think there is any need to worry about narrowing the scope of the test.

  * The migration method itself is not the target of the test; it is simply used to create the test table.
  * This feature has already been deprecated and has been removed in the latest version.
  * Tests where the database-side floating-point type is converted to `string` are covered by other tests such as `$table->decimal()`.

* [x] Adding methods to `Authenticatable` contract

  I have implemented a new method in the Authenticatable contract. Like any other request method, it is simply declared and empty.

## Test Results

* In my repository from which this pull request is based, I have confirmed that all existing tests pass with newer versions including php-8.3 and Laravel 11. CI results can be seen from [here](https://github.com/KentarouTakeda/barryvdh-laravel-ide-helper/actions/runs/7813420496).
* CI testing is done with SQLite only. I set up PostgreSQL and MySQL in my local environment and confirmed that the `User` class (`users` table and its migration) in the initial state of Laravel produced the same execution results as before.
   * SQL Server was not confirmed as it was difficult to set up.

I'm looking forward to continuing to use the great tools you've created in Laravel 11. Please let us know if you have any feedback regarding this pull request. Thank you for confirmation.

Regards.

### DIFF
--- a/.github/workflows/run-integration-tests.yml
+++ b/.github/workflows/run-integration-tests.yml
@@ -20,10 +20,14 @@ jobs:
             fail-fast: false
             matrix:
                 php: [8.3, 8.2, 8.1, 8.0]
-                laravel: [10.*, 9.*]
+                laravel: [11.*, 10.*, 9.*]
                 exclude:
                   - php: 8.0
                     laravel: 10.*
+                  - php: 8.0
+                    laravel: 11.*
+                  - php: 8.1
+                    laravel: 11.*
                   - php: 8.3
                     laravel: 9.*
         name: P${{ matrix.php }} - Laravel${{ matrix.laravel }}
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -21,13 +21,17 @@ jobs:
       matrix:
         os: [ubuntu-20.04, windows-2019]
         php: [8.3, 8.2, 8.1, 8.0]
-        laravel: [9.*, 10.*]
+        laravel: [9.*, 10.*, 11.*]
         dependency-version: [prefer-lowest, prefer-stable]
         exclude:
           - php: 8.0
             laravel: 10.*
           - php: 8.0
             dependency-version: prefer-lowest
+          - php: 8.0
+            laravel: 11.*
+          - php: 8.1
+            laravel: 11.*
     steps:
     - name: Set git to use LF
       if: ${{ matrix.os == 'windows-2019' }}
--- a/composer.json
+++ b/composer.json
@@ -25,25 +25,25 @@
         "barryvdh/reflection-docblock": "^2.0.6",
         "composer/class-map-generator": "^1.0",
         "doctrine/dbal": "^2.6 || ^3",
-        "illuminate/console": "^9 || ^10",
-        "illuminate/filesystem": "^9 || ^10",
-        "illuminate/support": "^9 || ^10",
+        "illuminate/console": "^9 || ^10 || ^11",
+        "illuminate/filesystem": "^9 || ^10 || ^11",
+        "illuminate/support": "^9 || ^10 || ^11",
         "nikic/php-parser": "^4.18 || ^5",
         "phpdocumentor/type-resolver": "^1.1.0"
     },
     "require-dev": {
         "ext-pdo_sqlite": "*",
         "friendsofphp/php-cs-fixer": "^3",
-        "illuminate/config": "^9 || ^10",
-        "illuminate/view": "^9 || ^10",
+        "illuminate/config": "^9 || ^10 || ^11",
+        "illuminate/view": "^9 || ^10 || ^11",
         "mockery/mockery": "^1.4",
-        "orchestra/testbench": "^7 || ^8",
-        "phpunit/phpunit": "^8.5 || ^9",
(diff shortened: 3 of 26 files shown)


hafezdivandari:
Why not remove `doctrine/dbal` in favor of native schema methods? You are using Doctrine DBAL to inspect database tables on [`ModelsCommand`](https://github.com/barryvdh/laravel-ide-helper/blob/a8b34ad228f3471d25a9789f4de62817a646af10/src/Console/ModelsCommand.php), instead of bringing `DoctrineConnection` back, you may use `Schema::getTables()`, `Schema::getColumns()`, `Schema::getForeignKeys()`, etc native methods.

Therefore, there's no need to register custom column types and you will have full support for all native column types (e.g. `enum`, `geometry`, etc.). please let me know if you have any question on using this methods.

barryvdh:
Thanks! Yeah I heard at Laracon that they removed doctrine altogether, and use their own functions. See https://github.com/laravel/framework/pull/48864

And now that I open it again, I see that @hafezdivandari did that PR, so good job :)

But yeah, my thought was also to drop doctrine and use that, so only support Laravel 11.

We released 2.14 yesterday, so I think we could probably skip directly to Laravel 11 for 2.15? Maybe a small maintenance release in the 2.14.x range, but removing doctrine would be a nice change to aim for. Perhaps @hafezdivandari would be open to help with this?

barryvdh:
@mfn what do you think about the Laravel 11 only support voor 2.15?

hafezdivandari:
Thank you @barryvdh

I added most of the new schema methods on Laravel 10 (you may check the list of PRs on laravel/framework#48864), then removed Doctrine DBAL on Laravel 11. So the schema methods I mentioned are also available on L10.

I'm ready to help; I think supporting Laravel 11 on `laravel-ide-helper` would be a good example for other packages.

barryvdh:
Yeah we could wait for L11.
See work in progress here: https://github.com/barryvdh/laravel-ide-helper/pull/1512
Probably missed some things, but we could do a L10.x release if we figure out what the minimum version is.

barryvdh:
So if we look at the release schedule, Laravel 9 is not supported since yesterday:
image

And looking at #1512 we can actually still support Laravel 10 (at least the latest minor versions). And it doesn't seem to change that much actually for the user (at least, no test changes required). PHPunit also doesn't impact te users.

I kinda prefer not bumping the major version, albeit just because it makes it easier for users to upgrade to Laravel 11. 
Laravel 11 should drop in about 4 weeks, so no real rush. But there is also no reason to keep supporting Laravel 9 or older 10.x releases. But supporting 10.x is a big pro, as long as it's officially supported by Laravel.

barryvdh:
I've cherry-picked your commits in https://github.com/barryvdh/laravel-ide-helper/pull/1520
Together with https://github.com/barryvdh/laravel-ide-helper/pull/1512 this replaces this PR

PR #1512, opened 2024-02-07
Remove doctrine

barryvdh:
See https://github.com/barryvdh/laravel-ide-helper/pull/1510

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,9 +25,9 @@
         "barryvdh/reflection-docblock": "^2.0.6",
         "composer/class-map-generator": "^1.0",
         "illuminate/console": "^10",
+        "illuminate/database": "^10.30",
         "illuminate/filesystem": "^10",
         "illuminate/support": "^10",
-        "illuminate/database": "^10.30",
         "nikic/php-parser": "^4.18 || ^5",
         "phpdocumentor/type-resolver": "^1.1.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -24,9 +24,10 @@
         "ext-json": "*",
         "barryvdh/reflection-docblock": "^2.0.6",
         "composer/class-map-generator": "^1.0",
-        "illuminate/console": "^9 || ^10",
-        "illuminate/filesystem": "^9 || ^10",
-        "illuminate/support": "^9 || ^10",
+        "illuminate/console": "^10",
+        "illuminate/filesystem": "^10",
+        "illuminate/support": "^10",
+        "illuminate/database": "^10.30",
         "nikic/php-parser": "^4.18 || ^5",
         "phpdocumentor/type-resolver": "^1.1.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "barryvdh/reflection-docblock": "^2.0.6",
         "composer/class-map-generator": "^1.0",
         "illuminate/console": "^10",
-        "illuminate/database": "^10.36",
+        "illuminate/database": "^10.38",
         "illuminate/filesystem": "^10",
         "illuminate/support": "^10",
         "nikic/php-parser": "^4.18 || ^5",
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     "require": {
         "php": "^8.0",
         "ext-json": "*",
-        "barryvdh/reflection-docblock": "^2.0.6",
+        "barryvdh/reflection-docblock": "^2.1.1",
         "composer/class-map-generator": "^1.0",
         "illuminate/console": "^10",
         "illuminate/database": "^10.38",
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
         "illuminate/config": "^9 || ^10",
         "illuminate/view": "^9 || ^10",
         "mockery/mockery": "^1.4",
-        "orchestra/testbench": "^7 || ^8",
+        "orchestra/testbench": "^8",
         "phpunit/phpunit": "^9",
         "spatie/phpunit-snapshot-assertions": "^4",
         "vimeo/psalm": "^5.4"
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,6 @@
         "ext-json": "*",
         "barryvdh/reflection-docblock": "^2.0.6",
         "composer/class-map-generator": "^1.0",
-        "doctrine/dbal": "^2.6 || ^3",
         "illuminate/console": "^9 || ^10",
         "illuminate/filesystem": "^9 || ^10",
         "illuminate/support": "^9 || ^10",
--- a/composer.json
+++ b/composer.json
@@ -38,8 +38,8 @@
         "illuminate/view": "^9 || ^10",
         "mockery/mockery": "^1.4",
         "orchestra/testbench": "^7 || ^8",
-        "phpunit/phpunit": "^8.5 || ^9",
-        "spatie/phpunit-snapshot-assertions": "^3 || ^4",
+        "phpunit/phpunit": "^9",
+        "spatie/phpunit-snapshot-assertions": "^4",
         "vimeo/psalm": "^5.4"
     },
     "suggest": {
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "barryvdh/reflection-docblock": "^2.0.6",
         "composer/class-map-generator": "^1.0",
         "illuminate/console": "^10",
-        "illuminate/database": "^10.30",
+        "illuminate/database": "^10.36",
         "illuminate/filesystem": "^10",
         "illuminate/support": "^10",
         "nikic/php-parser": "^4.18 || ^5",
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -1080,6 +1080,8 @@ class ModelsCommand extends Command
                     $default = 'null';
                 } elseif (is_int($default)) {
                     //$default = $default;
+                } elseif ($default instanceof \UnitEnum) {
+                    $default = '\\' . get_class($default) . '::' . $default->name;
                 } else {
                     $default = "'" . trim($default) . "'";
                 }
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -496,7 +496,7 @@ class ModelsCommand extends Command
         $table = $model->getTable();
         $schema = $model->getConnection()->getSchemaBuilder();
         $columns = $schema->getColumns($table);
-        $platformName = $model->getConnection()->getConfig('driver');
+        $driverName = $model->getConnection()->getDriverName();
 
         if (strpos($table, '.')) {
             [$database, $table] = explode('.', $table);
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -494,6 +494,7 @@ class ModelsCommand extends Command
         $table = $model->getTable();
         $schema = $model->getConnection()->getSchemaBuilder();
         $columns = $schema->getColumns($table);
+        $platformName = $model->getConnection()->getConfig('driver');
 
         if (!$columns) {
             return;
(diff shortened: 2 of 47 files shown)


KentarouTakeda:
Added fix to #1510,

* Remove Doctrine DBAL
* All tests pass with Laravel `>= ^10.38 || 11`

The completed code is in my repository.

There are still some points that I would like to carefully check, so it will be a while before I can say "ready for merge", but would you like to submit a draft pull request?

Change set: https://github.com/KentarouTakeda/barryvdh-laravel-ide-helper/compare/laravel-11...laravel-11-remove-doctrine-dbal
CI result: https://github.com/KentarouTakeda/barryvdh-laravel-ide-helper/actions/runs/7817602705/job/21325891898

From what I've tried, the `Schema::getXxx()` required by the Laravel IDE Helper seems to work starting with Laravel 10.38. The corresponding version in the change set above also complies with that.

barryvdh:
Sorry didn't see your commit until I fixed some changes. Will compare and check, thanks!

barryvdh:
I dont think the minimum stability was working before..

barryvdh:
So what do you think @mfn 
Should we target a v3 for this? It doesn't seem too much impact though.

jameshulse:
Awesome change thanks! I have tested this on a relatively complex Postgres database and it fixes a major issue we had with Doctrine that affected both Laravel and ide-helper previously.

We have a table in two schemas with the same name, but differing types on the `id` column. Previously Doctrine didn't handle schemas well and would return all columns for all tables matching the name, regardless of schema. So ide-helper ended up with two `id` columns, one a string, and one an integer. We would sometimes end up with the wrong type information.

I'll be using this dev version until it is fully released.

barryvdh:
Yes added that now

composer.json
```diff
@@ -22,12 +22,12 @@
...
+        "barryvdh/reflection-docblock": "^2.1.1",
         "composer/class-map-generator": "^1.0",
-        "doctrine/dbal": "^2.6 || ^3.1.4",
-        "illuminate/console": "^9 || ^10",
-        "illuminate/filesystem": "^9 || ^10",
-        "illuminate/support": "^9 || ^10",
+        "illuminate/console": "^10",
+        "illuminate/database": "^10.38",
+        "illuminate/filesystem": "^10",
+        "illuminate/support": "^10",
```
KentarouTakeda, line 30: Since we removed the dependency on Laravel 9, we can remove `^7` from `orchestra/testbench` located in `require-dev`.

src/Console/ModelsCommand.php
```diff
@@ -502,47 +490,29 @@ protected function getTypeOverride($type)
...
 
         $this->setForeignKeys($schema, $table);
         foreach ($columns as $column) {
-            $name = $column->getName();
+            $name = $column['name'];
             if (in_array($name, $model->getDates())) {
                 $type = $this->dateClass;
             } else {
-                $type = $column->getType()->getName();
+                $type = $column['type_name'];
```
hafezdivandari, line 515:
```suggestion
                $type = $this->laravel['config']->get("ide-helper.custom_db_types.{$driverName}.{$column['type_name']}", null);
                $type ??= $column['type'] === 'tinyint(1)' ? 'boolean' : null;
                $type ??= match ($column['type_name']) {
                    'varchar', 'nvarchar',
                    'char', 'bpchar', 'nchar',
                    'varbinary', 'blob', 'bytea',
                    'decimal', 'numeric',
                    'date', 'time', 'timetz', 'datetime', 'datetime2', 'datetimeoffset', 'timestamp', 'timestamptz', 'year',
                    'uuid', 'uniqueidentifier',
                    'json', 'jsonb',
                    'macaddr', 'inet',
                    'enum', 'set',
                    'geography', 'geometry', 'geometrycollection', 'linestring', 'multilinestring', 'multipoint', 'multipolygon', 'point', 'polygon',
                    'text', 'tinytext', 'longtext', 'mediumtext' => 'string',

                    'tinyint',
                    'integer', 'int', 'int4',
                    'smallint', 'int2',
                    'mediumint',
                    'bigint', 'int8' => 'integer',

                    'boolean', 'bool', 'bit' => 'boolean',

                    'float', 'real', 'float4',
                    'double', 'float8' => 'float',

                    default => 'mixed',
                };
```

src/Console/ModelsCommand.php
```diff
@@ -502,47 +490,29 @@ protected function getTypeOverride($type)
...
     {
-        $database = $model->getConnection()->getDatabaseName();
-        $table = $model->getConnection()->getTablePrefix() . $model->getTable();
-        $schema = $model->getConnection()->getDoctrineSchemaManager();
-        $databasePlatform = $schema->getDatabasePlatform();
-        $databasePlatform->registerDoctrineTypeMapping('enum', 'string');
+        $table = $model->getTable();
+        $schema = $model->getConnection()->getSchemaBuilder();
+        $columns = $schema->getColumns($table);
+        $platformName = $model->getConnection()->getConfig('driver');
```
hafezdivandari, line 499:
```suggestion
        $driverName = $model->getConnection()->getDriverName();
```

hafezdivandari (changes requested):

src/Console/ModelsCommand.php
hafezdivandari: Almost all possible types for `mysql`, `mariadb`, `pgsql`, `sqlsrv`, and `sqlite`.

barryvdh: Ah yes, but I think tinyint(1) is an integer for most drivers still? Only after casting it becomes a bool

hafezdivandari: on MySQL, `bool` and `boolean` are synonyms for `tinyint(1)`, PHP also accepts `1` and `0` as boolean. But the mapping for all these edge cases (e.g. `year`, `binary`, `timestamp`, etc.) are a matter of personal preference IMO, it's up to you.

barryvdh: Yeah but I mean, if you would `dump()` the attribute, it would show as integer, not as bool. Best practice would be to cast it to bool, but don't think Laravel does that for you.

hafezdivandari: You're right, AFAIK Laravel doesn't cast it to bool.

hafezdivandari:
You may also change `config/ide-helper.php` to something like this:

    /*
    |--------------------------------------------------------------------------
    | Custom DB type mappings
    |--------------------------------------------------------------------------
    |
    | This setting allow you to map any database type, including custom types that you may have
    | created using CREATE TYPE statement or imported using database plugin
    | / extension to a PHP type.
    |
    | Each key in this array is a `connections.driver` of the `database` config file. Currently valid names are:
    | 'pgsql', 'mysql', 'mariadb', 'sqlite', 'sqlsrv'
    | 
    | The value of the array is an array of type mappings. Key is the name of the column type,
    | (for example, `jsonb` from Postgres 9.4) and the value is the name of the corresponding PHP type.
    |
    | So to map `jsonb` to `array` in your models when working with Postgres, just add the following entry to the array below:
    |
    | "pgsql" => [
    |       "jsonb" => "array",
    |  ],
    |
    */
    'custom_db_types' => [

    ],

src/Console/ModelsCommand.php
```diff
@@ -508,86 +496,65 @@ protected function getTypeOverride($type)
...
-        $databasePlatform = $schema->getDatabasePlatform();
-        $databasePlatform->registerDoctrineTypeMapping('enum', 'string');
+        $table = $model->getTable();
+        $schema = $model->getConnection()->getSchemaBuilder();
+        $columns = $schema->getColumns($table);
+        $driverName = $model->getConnection()->getDriverName();
 
         if (strpos($table, '.')) {
             [$database, $table] = explode('.', $table);
         }
```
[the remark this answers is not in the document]
hafezdivandari: Let Laravel handle schema name (e.g `schema.table`)

src/Console/ModelsCommand.php
```diff
@@ -508,86 +496,65 @@ protected function getTypeOverride($type)
...
-                        break;
-                    default:
-                        $type = 'mixed';
-                        break;
-                }
+                // Allow to register custom types
+                $type = $this->laravel['config']->get("ide-helper.custom_db_types.{$driverName}.{$column['type_name']}", null);
+
+                // mysql and sqlite drivers return tinyint(1) as integer, not as bool
+                $type ??= $column['type'] === 'tinyint(1)' && !in_array($driverName, ['mysql', 'sqlite']) ? 'boolean' : null;
```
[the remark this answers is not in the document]
hafezdivandari: You should either map `tinyint(1)` to `boolean` on MySQL and SQLite or remove the entire line, as PostgreSQL and SQL Server have `boolean` and `bit` for this (i.e. if someone has `tinyint` column on PostgreSQL, he definitely wants `integer` not `boolean`).

hafezdivandari (changes requested):

src/Console/ModelsCommand.php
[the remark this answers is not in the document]
barryvdh: Hmm yeah but there were a few issues about it before. But might be just doctrine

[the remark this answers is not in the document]
hafezdivandari: Yeah it was Doctrine issue, I've added schema name support for these methods on PostgreSQL (L10) and SQL Server (L11).

barryvdh:
So except for the binary -> string change, this doesn't seem to break anything in our current test setup. So LGTM

hafezdivandari (approved):

hafezdivandari:
@barryvdh you missed https://github.com/barryvdh/laravel-ide-helper/pull/1512#issuecomment-1934154779?

barryvdh:
Do we need to group the types by driver still? How does custom mapping work anyways, php handles it the same anyways right? If you want a different type, you should cast it. If it's different for a driver, we should just fix it on our side?

hafezdivandari:
> If you want a different type, you should cast it. If it's different for a driver, we should just fix it on our side?

It seems right.

The only edge case comes to my mind is PostgreSQL array types, we can simply fix that, but it's also `string`:

```php
DB::statement('create table "test" ("arr" int[])');
DB::statement('insert into "test" ("arr") values (array[1,2,3])');

var_dump(DB::table('test')->value('arr')); // string(7) "{1,2,3}"
```

So I think we can remove `custom_db_types` as you said?

hafezdivandari:
> Or maybe the root issue is to default to the mixed type. I can imagine that the default actual type usually is string anyway.

I agree. We can assume DB always returns `string` unless it's explicitly `integer`, `boolean`, or `float`.

KentarouTakeda:
I have tested this on a private project with 70 tables and 680 columns, including PostgreSQL proprietary types. It produced exactly the same results as the previous version. LGTM.

I agree with treating unknown types as `string` rather than `mixed`.Because PDO's implementation is such, we receive those values ​​that Laravel cannot interpret as `string`.

pgsql:
https://github.com/php/php-src/blob/php-8.3.3/ext/pdo_pgsql/pgsql_statement.c#L673-L674

mysql:
https://github.com/php/php-src/blob/php-8.3.3/ext/pdo_mysql/mysql_statement.c#L823-L825

sqlite:
https://github.com/php/php-src/blob/php-8.3.3/ext/pdo_sqlite/sqlite_statement.c#L331-L338

sqlsrv(odbc):
https://github.com/php/php-src/blob/php-8.3.3/ext/pdo_odbc/odbc_stmt.c#L645

barryvdh:
Yes removed them.

barryvdh:
Yeah bit conflicted about the 3.0 vs 2.x. It technically isn't a BC break, only if people are relying on Doctrine because of us.
I only think it's a bit annoying to force everyone to upgrade, but they would probably check their dependencies for Laravel 11 anyway.

But probably better to force the upgrade in 3.x, as it will coincide with Laravel dropping it anways. https://laravel.com/docs/master/upgrade#doctrine-dbal-removal

barryvdh:
Okay changed the branch alias to 3.0-dev

PR #1513, opened 2024-02-07
Fix lowest stability

barryvdh:
Fix the lowest stability by actually checking it. 
Needed to bump Carbon but included it in dev only because I think it only impacts user code anyway.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -18,16 +18,17 @@ jobs:
       COMPOSER_NO_INTERACTION: 1
 
     strategy:
+      fail-fast: false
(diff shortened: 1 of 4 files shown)


barryvdh:
It seems doctrine/dbal 2.x is not actually supported in Laravel 10. We need minimum of 3.1.
https://github.com/laravel/framework/issues/46238

barryvdh:
So Laravel 9 does support doctrine 2, so maybe we should bump Laravel to 10 minimum now, but then we could just as well do #1512 directly.

barryvdh:
Closing in favor of removing doctrine

Issue #1511, opened 2024-02-07
[idea] Refactor to use phpstan/phpdoc-parser


barryvdh:
Just throwing this as a reminder/placeholder.

I came across https://github.com/phpstan/phpdoc-parser which is a zero-requirements package to parse and modify phpdocs.
I've forked https://github.com/barryvdh/ReflectionDocBlock in the past because the original didn't support modifying phpdocs, but it's clearly not so up-to-date.

It would be interesting to see if we can migrate to use that library instead.

barryvdh:
I think it's not as easy, because that parser doesn't support the context. So you would need something like https://github.com/phpDocumentor/TypeResolver or https://github.com/phpstan/phpstan-src/blob/1.6.x/src/PhpDoc/TypeNodeResolver.php#L160 to resolve the type for us.

barryvdh:
Also something interesting: https://packagist.org/packages/symfony/property-info
For phpdocs we don't really want processing, just want to build it. But for types, we want to use multiple sources (we often already use both reflection, phpdocs and more..)

PR #1522, opened 2024-02-18
[3.x] Use official ReflectionDocBlock

barryvdh:
## Summary
Try to avoid using the fork of ReflectionDocBlock.

The official lib still doesn't support changing the docs but it does support creating a new one..

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/tests/Console/ModelsCommand/AdvancedCasts/__snapshots__/Test__test__1.php
+++ b/tests/Console/ModelsCommand/AdvancedCasts/__snapshots__/Test__test__1.php
@@ -17,7 +17,7 @@ use Illuminate\Database\Eloquent\Model;
  * @property \Carbon\CarbonImmutable $cast_to_immutable_date
  * @property \Carbon\CarbonImmutable $cast_to_immutable_custom_datetime
  * @property \Carbon\CarbonImmutable $cast_to_immutable_datetime
- * @property integer $cast_to_timestamp
+ * @property int $cast_to_timestamp
  * @property mixed $cast_to_encrypted
  * @property array $cast_to_encrypted_array
  * @property \Illuminate\Support\Collection $cast_to_encrypted_collection
--- a/tests/Console/ModelsCommand/ArrayCastsWithComment/__snapshots__/Test__test__1.php
+++ b/tests/Console/ModelsCommand/ArrayCastsWithComment/__snapshots__/Test__test__1.php
@@ -7,18 +7,16 @@ namespace Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\ArrayCastsWithCo
 use Illuminate\Database\Eloquent\Model;
 
 /**
- * Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\ArrayCastsWithComment\Models\ArrayCastsWithComment
- *
- * @property array<int, string>|null $cast_to_array -- These three should not be duplicated
- * @property array<int, string> $cast_to_json some-description
- * @property \Illuminate\Support\Collection<int, string> $cast_to_collection some-description
+ * @property array<int,string>|null $cast_to_array -- These three should not be duplicated
+ * @property array<int,string> $cast_to_json some-description
+ * @property \Illuminate\Support\Collection<int,string> $cast_to_collection some-description
  * @property array|null $cast_to_encrypted_array -- These three are OK (no types)
  * @property array $cast_to_encrypted_json some-description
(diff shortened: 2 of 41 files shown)


barryvdh:
Hmm it seems to show the classes for itself as FQN which is not really ideal probably.

tests/Console/ModelsCommand/ArrayCastsWithComment/__snapshots__/Test__test__1.php
```diff
@@ -28,9 +26,6 @@
  * @property string $cast_to_bool
  * @property string $cast_to_boolean
  * @property string $cast_to_object
- * @property array $cast_to_array
```
barryvdh, line 31: This is actually correct, because it is already listed.

tests/Console/ModelsCommand/AdvancedCasts/__snapshots__/Test__test__1.php
```diff
@@ -17,7 +17,7 @@
  * @property \Carbon\CarbonImmutable $cast_to_immutable_date
  * @property \Carbon\CarbonImmutable $cast_to_immutable_custom_datetime
  * @property \Carbon\CarbonImmutable $cast_to_immutable_datetime
- * @property integer $cast_to_timestamp
+ * @property int $cast_to_timestamp
```
barryvdh, line 20: `bool` and `int` are what used in Laravel anyway.

tests/Console/ModelsCommand/ArrayCastsWithComment/__snapshots__/Test__test__1.php
```diff
@@ -7,18 +7,16 @@
 use Illuminate\Database\Eloquent\Model;
 
 /**
- * Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\ArrayCastsWithComment\Models\ArrayCastsWithComment
- *
- * @property array<int, string>|null $cast_to_array -- These three should not be duplicated
- * @property array<int, string> $cast_to_json some-description
- * @property \Illuminate\Support\Collection<int, string> $cast_to_collection some-description
+ * @property array<int,string>|null $cast_to_array -- These three should not be duplicated
```
barryvdh, line 10: Whitespace changes, but I guess that doesn't matter?

tests/Console/ModelsCommand/CustomCollection/__snapshots__/Test__test__1.php
```diff
@@ -11,11 +11,11 @@
 /**
  * Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\CustomCollection\Models\Simple
  *
- * @property integer $id
+ * @property int $id
  * @property-read SimpleCollection<int, Simple> $relationHasMany
  * @property-read int|null $relation_has_many_count
- * @method static SimpleCollection<int, static> all($columns = ['*'])
- * @method static SimpleCollection<int, static> get($columns = ['*'])
+ * @method static void SimpleCollection() <int, static> all($columns = ['*'])
```
barryvdh, line 17: TBH, I'm not sure which is correct, I think neither?

tests/Console/ModelsCommand/Factories/__snapshots__/Test__testFactory__1.php
```diff
@@ -11,7 +11,7 @@
 /**
  * Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\Factories\Models\ModelWithCustomNamespace
  *
- * @method static \Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\Factories\CustomSpace\ModelWithCustomNamespaceFactory factory($count = null, $state = [])
+ * @method static \Barryvdh\LaravelIdeHelper\Tests\Console\ModelsCommand\Factories\CustomSpace\ModelWithCustomNamespaceFactory factory(mixed $count, mixed $state)
```
barryvdh, line 14: Not sure why this is changed, maybe it's not in the spec?

barryvdh:
@mfn we can remove it for the Model collector at least, except there a few changes (see comments mostly). Not sure if we need to change those, but it would be nice if we could just use the upstream package.

But we would need to do the same for the other generator command still.

barryvdh:
Yeah I think the upstream lib doesn't handle generics like we want to. And in general, it tries to 'fix' this for us, which we don't really need..

barryvdh:
Hmm not really happy with this as-is. Might leave it to think about it, or explore alternatives (see #1511)

PR #1523, opened 2024-02-19
Merge reset/smart reset, don't set class

barryvdh:
## Summary
Instead of having reset and smart-reset, just reset and never change the test. I don't really see why we should mention the classname in the phpdocs? So just leave the text be, or empty?

## Type of change

### Checklist
- [ ] Update the README.md


barryvdh:
I don't know why this is like it is, but it kind a feels odd.

barryvdh:
Yeah I guess we could, maybe slightly annoying but it's allowed I suppose

Issue #590, opened 2017-11-02
Is or will be supported model hint for Mongodb?


CesarScur:
Is or will be supported model hint for Mongodb?

MordiSacks:
MongoDB is shameless, what kind of hinting do you want?

mfn:
@barryvdh we support and can only support databases officially supported by Laravel + doctrine and MongoDB isn't one of them, IMHO the issue can be closed :}

PR #1528, opened 2024-02-22
Allowing table inspection to gracefully fail so other IDE helpers for models may still be generated

clayzar:
## Summary
This adds logic to allow table inspection to gracefully fail. More specifically, this allows other the doc block generation to continue and generate other helpers that aren't database related. 

A use case for this is for NoSQL database drivers, for which schema inspection doesn't make sense or doesn't have a convention to determine columns, yet is still helpful to have a handful of the other IDE helpers that this package is capable of generating. Currently, when using the latest Laravel MongoDB package the corresponding driver does not implement the `compileColumns` method. So we see an error thrown that mentions this undefined method when running the command to generate doc blocks for models configured to use a Mongo DB connection.

 I've linked a couple of issues that mention this. 

I've also added silence-able warnings if table properties could not be inspected for a given model. A warning is output as the command generates output and another at the end of the output denoting the models whose table could not be inspected, to ensure the developer is aware to scroll up should the output be many lines long. See attached screenshots

https://github.com/barryvdh/laravel-ide-helper/issues/590
https://github.com/mongodb/laravel-mongodb/issues/1785

## Type of change
This could be thought of as a bug fix, new feature, or misc change – depending on how you look at it.

### Checklist
- [ ] Update the README.md

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -11,7 +11,7 @@ return [
     |
     */
 
-    'filename'  => '_ide_helper.php',
+    'filename' => '_ide_helper.php',
 
     /*
     |--------------------------------------------------------------------------
@@ -152,6 +152,19 @@ return [
     */
 
     'ignored_models' => [
+
+    ],
+
+    /*
+    |--------------------------------------------------------------------------
+    | Models to silence warnings
+    |--------------------------------------------------------------------------
+    |
+    | Define which models should have warnings silenced
+    |
+    */
+
+    'silenced_models' => [
 
     ],
 
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -252,6 +252,8 @@ class ModelsCommand extends Command
             $this->laravel['config']->get('ide-helper.ignored_models', [])
         );
 
+        $skippedGeneratingTablePropertiesForModels = [];
+
         foreach ($models as $name) {
             if (in_array($name, $ignore)) {
                 if ($this->output->getVerbosity() >= OutputInterface::VERBOSITY_VERBOSE) {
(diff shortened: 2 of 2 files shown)


clayzar:
Great thoughts here! I've made those adjustments and also tucked away the updates more neatly. Just to regurgitate what you had in mind:

If the `ide-helpers.silenced_models` value is null or otherwise missing (such would be the case for current projects using the package).. no behavior changes, exception will be thrown.

Otherwise, this will gracefully handle and exception when inspecting the table and provide warnings in the output. Those warnings can be silenced by specifying the FQCN of the class in `ide-helper.silenced_models`.

I've added the exception class and message to the output. Here's a sample of the output in an example project with 2 models that use a MongoDB connection

Screenshot 2024-02-23 at 4 52 32 PM

clayzar:
I feel that setting the value of `silenced_models` to an empty array to allow graceful error handling isn't very intuitive.

With the warning collection, I was trying to map out something that could be used in case other warnings were ever possible, but I realize this may be a little overblown for the need here.

Thinking back on the primary goal here, I think it may be simpler to just use a config setting named `skip_model_tables` that defaults to an empty array. This would default to the same behavior as today (disabled) and must be enabled deliberately. The items in the array would be FQCN for any model for which table inspection should be skipped. The value could also be set to `*` to skip all models without having to explicitly list each one, helpful for large code-bases exclusively using an unsupported database driver.

Thoughts?

config/ide-helper.php
```diff
@@ -11,7 +11,7 @@
     |
     */
 
-    'filename'  => '_ide_helper.php',
+    'filename' => '_ide_helper.php',
```
hassan7303, line 14: nice

PR #1563, opened 2024-06-20
Update view "version" variable name to avoid potential conflicts

pb30:
## Summary
We inherited a number of projects that have a global `version` variable shared with all views that is an array of things like asset versions, etc.

When running `php artisan ide-helper:generate` it results in an error since `$version` is used in this packages views.
```
Array to string conversion at vendor/barryvdh/laravel-ide-helper/resources/views/helper.php:17
```

This PR renames the variable to `$laravel_version` to avoid conflicts since it may be a common global variable name.

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 - Fix return value of query scopes from parent class [#1366 / sforward](https://github.com/barryvdh/laravel-ide-helper/pull/1366)
 
 ### Changed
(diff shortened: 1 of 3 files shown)


barryvdh:
Hmm I think it's a bit odd that we need to account for all possible variables. If someone (slight chance) has overriden the template, this would be considered BC break.

barryvdh:
Merged in https://github.com/barryvdh/laravel-ide-helper/pull/1637

PR #1566, opened 2024-07-07
Update README.md

Abassion:
## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ If for some reason you want manually control this:
     }
   }
   ```
+Only if you are using Laravel 10 and **below**:
+
 - Add the following class to the `providers` array in `config/app.php`:
   ```php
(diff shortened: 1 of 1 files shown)


barryvdh:
And what for Laravel 10+?

barryvdh:
Yeah it works out-of-the box since Laravel 6 or so, it says so above. This section is only for manual. But depending if you upgraded or not, you could still use the old style.

Abassion:
> Yeah it works out-of-the box since Laravel 6 or so, it says so above. This section is only for manual. But depending if you upgraded or not, you could still use the old style.

I see. The reason why I added the line was mainly because I faced an issue when I tried to use the manual config because it was referring that I should add the provider to "config/app.php", which does not work in Laravel 11. Maybe adding `Add the following class to the providers array in config/app.php (or bootstrap/provider.php if you are using Laravel 11):` would help.

Issue #1572, opened 2024-07-17
Issues after Laravel 11.15


Sergiobop:
### Versions:
- ide-helper Version: 3.1
- Laravel Version: 11.15
- PHP Version: 8.3

### Description:
Additionally to issues like https://github.com/barryvdh/laravel-ide-helper/issues/1571, Laravel 11.15 also ruined some other _ide_helper.php methods (find, findOrFail, etc)

I think is related to https://github.com/laravel/framework/pull/51851 and https://github.com/laravel/framework/pull/52037

Before 11.15:
```
/**
             * Find a model by its primary key or throw an exception.
             *
             * @param mixed $id
             * @param array|string $columns
             * @return \Illuminate\Database\Eloquent\Model|\Illuminate\Database\Eloquent\Collection|static|static[] 
             * @throws \Illuminate\Database\Eloquent\ModelNotFoundException<\Illuminate\Database\Eloquent\Model>
             * @static 
             */            public static function findOrFail($id, $columns = [])
            {
                                /** @var \Illuminate\Database\Eloquent\Builder $instance */
                                return $instance->findOrFail($id, $columns);
            }
```

After 11.15:
```
/**
             * Find a model by its primary key or throw an exception.
             *
             * @param mixed $id
             * @param array|string $columns
             * @return \Illuminate\Database\Eloquent\($id is (\Illuminate\Contracts\Support\Arrayable<array-key, mixed>|array<mixed>) ? \Illuminate\Database\Eloquent\Collection<int, TModel> : TModel)
             * @throws \Illuminate\Database\Eloquent\ModelNotFoundException<TModel>
             * @static 
             */            public static function findOrFail($id, $columns = [])
            {
                                /** @var \Illuminate\Database\Eloquent\Builder $instance */
                                return $instance->findOrFail($id, $columns);
            }
```

Now for example this is wrong:
$pack = Pack::query()->findOrFail($packId);

Previous to 11.15: $pack: Pack|Pack[]|Builder|Builder[]|Collection|Model|null
After 11.15: $pack: |Collection|Model|null

I don't know if we have to do something to make it work again in our end.

Thanks

joelstein:
Same here.

For example, `User::first()` returns a User object. However, after Laravel 11.15, I get an intelephense error if I try to typehint it.

> Expected type 'App\Models\User'. Found 'Illuminate\Database\Eloquent\TValue|null'.

From _ide_helper.php:

```
/**
[...]

Issue #1574, opened 2024-07-27
Get type of model when using Eloquent with methods from Query\Builder


pjio:
First thank you for this great package which helped me a lot!

### Summary

For methods from `Illuminate\Database\Query\Builder` the generated return type is `Illuminate\Database\Query\Builder`. Therefore methods like `whereNotNull()` lose information about the model:

![Screenshot-2024-07-27_13-02](https://github.com/user-attachments/assets/dbaacb0a-782d-47ce-8f98-ea7e84dc2794)

For methods from `Illuminate\Database\Eloquent\Builder` the generated return type is `Illuminate\Database\Eloquent\Builder|static`. Therefore the result is a correctly typed model or collection (see: https://github.com/barryvdh/laravel-ide-helper/commit/378adeb1f563af656fafa1adcbd893bd45e6d469)

It should be possible to use the methods with the same behavior:

![Screenshot-2024-07-27_13-04](https://github.com/user-attachments/assets/9eef2859-6163-42ee-9712-9be630c6bc26)

<details>
  <summary>Roughly 122 methods are affected</summary>
select()
selectSub()
selectRaw()
fromSub()
fromRaw()
addSelect()
distinct()
from()
useIndex()
forceIndex()
ignoreIndex()
join()
joinWhere()
joinSub()
joinLateral()
leftJoinLateral()
leftJoin()
leftJoinWhere()
leftJoinSub()
rightJoin()
rightJoinWhere()
rightJoinSub()
crossJoin()
crossJoinSub()
mergeWheres()
whereColumn()
orWhereColumn()
whereRaw()
orWhereRaw()
whereIn()
orWhereIn()
whereNotIn()
orWhereNotIn()
whereIntegerInRaw()
orWhereIntegerInRaw()
whereIntegerNotInRaw()
orWhereIntegerNotInRaw()
whereNull()
orWhereNull()
whereNotNull()
whereBetween()
whereBetweenColumns()
orWhereBetween()
orWhereBetweenColumns()
whereNotBetween()
whereNotBetweenColumns()
orWhereNotBetween()
orWhereNotBetweenColumns()
orWhereNotNull()
whereDate()
orWhereDate()
whereTime()
orWhereTime()
whereDay()
orWhereDay()
whereMonth()
orWhereMonth()
whereYear()
orWhereYear()
whereNested()
addNestedWhereQuery()
whereExists()
orWhereExists()
whereNotExists()
orWhereNotExists()
[...]

PR #1575, opened 2024-07-27
Improve replacement of return type for methods from Query\Builder

pjio:
## Summary

See issue for expected improvement: https://github.com/barryvdh/laravel-ide-helper/issues/1574

The class `\Eloquent` in `_ide_helper.php` contains methods from `Illuminate\Database\Eloquent\Builder` and `Illuminate\Database\Query\Builder`. The return type for `$this` is replaced with the real class (in case of the Eloquent\Builder already with an additional `|static`). This merge requests replaces `$this` in methods from Query\Builder with the identical return type `Illuminate\Database\Eloquent\Builder|static`. In the context of an `\Eloquent` instance this should be the real behavior and improves the derived types.

## Type of change

### Checklist

### DIFF
--- a/src/Alias.php
+++ b/src/Alias.php
@@ -395,7 +395,8 @@ class Alias
                             $reflection,
                             $macro_name,
                             $this->interfaces,
-                            $this->classAliases
+                            $this->classAliases,
+                            $this->getReplaceReturnTypes()
                         );
                         $this->usedMethods[] = $macro_name;
                     }
--- a/src/Alias.php
+++ b/src/Alias.php
@@ -18,6 +18,7 @@ use Barryvdh\Reflection\DocBlock\Tag\MethodTag;
 use Closure;
 use Illuminate\Config\Repository as ConfigRepository;
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
+use Illuminate\Database\Query\Builder as QueryBuilder;
 use Illuminate\Support\Facades\Facade;
 use ReflectionClass;
 use Throwable;
@@ -340,7 +341,7 @@ class Alias
                         $magic,
                         $this->interfaces,
                         $this->classAliases,
-                        $this->getReplaceReturnTypes()
+                        $this->getReplaceReturnTypes($class)
                     );
                 }
                 $this->usedMethods[] = $magic;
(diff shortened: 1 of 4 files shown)


pjio:
Here the old and new file for easier comparison:
[_ide_helper.php.old.txt](https://github.com/user-attachments/files/16400689/_ide_helper.php.old.txt)
[_ide_helper.php.new.txt](https://github.com/user-attachments/files/16400690/_ide_helper.php.new.txt)

eldair:
@pjio will this fix https://github.com/barryvdh/laravel-ide-helper/issues/1572 as well?

pjio:
> @pjio will this fix #1572 as well?

No, for #1572 it doesn't make a difference. The cause is a different one. This PR only affects the return type `$this`.

pjio:
Changes 03.09.24
- `normalizeReturn()` does set the return type only with `$tag->setContent()` (before it was overwritten by `$tag->setType()` which had the side effect to reset `Tag::$content`)
- It is now clearer how `$this` could be normalized depending on the alias.
- Generation of `_ide_helper.php` tested with Laravel 10 and PHP 8.1, 8.2 and 8.3
- Set "Type of change" to "New feature" (the generated file changes but it shouldn't break anything)

To test the change in VSCode the Intelephense extension needs to be downgraded to `1.10.4`. Since `1.12` there is a stricter type check which removes `static` from the union return type. With the next update for Intelephense `1.12.6` it should work again (bug report was made and it is queued for the next release).

CongAn:
Thank you very much for your valuable contributions! 
I would greatly appreciate it if the project maintainers could merge this pull request and release a new version as soon as possible.

barryvdh:
Is this fixed with 3.2 now adding the <static> type?

pjio:
> Is this fixed with 3.2 now adding the type?

Not yet. This pull request additionally adds the type for various methods of `\Eloquent` in `_ide_helper.php` (I tested with a fresh Laravel 11 project and 3.2 to see if it already would work without the change).

Examples:
```php
User::query()->where('name', 'joe')->first(); // `query()` from the model already works as of 3.2
User::where('name', 'joe')->first();          // `where` from Eloquent\Builder will work with this pull request
User::whereNull('name')->first();             // `whereNull` from Query\Builder will work with this pull request
```

PR #1594, opened 2024-10-17
fix: Model's `--write` duplicates array properties

KentarouTakeda:
## Summary

Thanks for merging barryvdh/ReflectionDocBlock#17. This pull request closes #1505 

Now that the new version has been published:

* Use the new corrected version for testing, including the `prefer-lowest` tests
* Fix the tests that were marked as incorrect to be correct

I've made the above changes.

## Type of change

- [x] Misc. change (internal, infrastructure, maintenance, etc.)

### Checklist

- [x] Existing tests have been adapted and/or new tests have been added

PR #1595, opened 2024-10-17
Fix upstream dependent test failure in Laravel 11.23 and later

KentarouTakeda:
## Summary

Fixed test failures in Laravel 11.23 and later. The cause is the following changes on the laravel/framework side:

* laravel/framework#52729
* laravel/framework#52762

I addressed the issue by changing the Laravel method that we were testing.

The following points were taken into consideration when selecting the target method:

* It must be a method of the Eloquent `Builder` class, just like before the fix.
* The content does not change between the CI options `prefer-oldest` and `prefer-stable`, and is unlikely to change in the future.
* It must be a method that accepts multiple parameters.
* Some of these parameters must have default values.

By satisfying these conditions, we can perform tests equivalent to `with()`, which was the target of tests before the fix. `upsert` met the conditions, so I selected it.

## Note

This pull request, #1594, and #1591 can all be merged independently.

IMO, the following order is best:

1. Merging 1594 should make all of this pull request's tests pass.
2. If we merge this pull request, all of 1591's tests should pass.
3. 1591 contains a breaking change. Merge it last.

## Type of change

### Checklist

### DIFF
--- a/tests/MethodTest.php
+++ b/tests/MethodTest.php
@@ -59,27 +59,28 @@ DOC;
     public function testEloquentBuilderOutput()
     {
         $reflectionClass = new \ReflectionClass(Builder::class);
-        $reflectionMethod = $reflectionClass->getMethod('with');
+        $reflectionMethod = $reflectionClass->getMethod('upsert');
 
         $method = new Method($reflectionMethod, 'Builder', $reflectionClass);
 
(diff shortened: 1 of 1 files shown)


barryvdh:
It seems to work now :) 
Guess we'll just stick with the latest version for now.

PR #1618, opened 2024-11-27
Fix PHP 8.4 deprecations

JeppeKnockaert:
## Summary
This PR removes implicitly nullable parameters because these result in deprecation warnings for PHP 8.4.

https://www.php.net/manual/en/migration84.deprecated.php#migration84.deprecated.core.implicitly-nullable-parameter

## Type of change

### Checklist
- [x] Update the README.md

### DIFF
--- a/resources/views/helper.php
+++ b/resources/views/helper.php
@@ -1,8 +1,8 @@
 <?= '<?php' ?>
 <?php
 /**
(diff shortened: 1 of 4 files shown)


dcranford-em:
This would be great to get merged in. My logs are filling up with deprecation warnings

PR #1629, opened 2024-12-11
Allow adding custom Macroable classes.

mathieutu:
## Summary
Some packages, like [Filament](https://filamentphp.com), implements their own version of the Macroable trait.
Spatie even made [their own version](https://github.com/spatie/macroable).

This PR allow adding any macro class we want to detect in the config file.

Thanks,
Mathieu.
## Type of change

### Checklist
- [x] Update the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,8 @@ Str::macro('concat', function(string $str1, string $str2) : string {
 });
 ```
 
+You can add any custom Macroable traits to detect in the `macroable_traits` config option.
+
 ### Automatic PHPDocs for models
 
 If you don't want to write your properties yourself, you can use the command `php artisan ide-helper:models` to generate
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -344,7 +344,8 @@ return [
     |
     */
     'macroable_traits' => [
-        Illuminate\Support\Traits\Macroable::class,
+        \Filament\Support\Concerns\Macroable::class,
+        \Spatie\Macroable\Macroable::class,
     ],
 
 ];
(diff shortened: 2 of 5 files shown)


mathieutu:
This PR can't be merged as it, because of these lines:
https://github.com/barryvdh/laravel-ide-helper/blob/4d2728c4f05ae7514e8a4761a2a8bcfaf8f55654/src/Alias.php#L386-L387

There are two problems with them:
- We're not looking for traits recursively, contrary to what is done in the Generator. (**Leads to bugs**)
- The Macroable class is hardcoded. (Topic of this PR)
- (The code is dupplicated)

I'm ok to continue and make the changes. 

However, could you tell me if you're interested in this or not? 
If this PR will be closed anyway, I don't invest more time in it.

Thanks!

barryvdh:
I think this is okay to add. I would prefer to have the default macro class hardcoded, only add custom classes to the config.

mathieutu:
Hey @barryvdh Thanks for your help and the merge.
FYI, it's still not working with Filament, as they don't use the same macroable code as Laravel:
Laravel: 
Screenshot 2025-01-16 at 17 22 08

Filament:
Screenshot 2025-01-16 at 17 21 59

Hence, it's obviously not compatible with this package `Alias::getMacroFunction` function... 😞

Not sure whether you'd like to provide a special treatment for them?

erikn69:
> it's still not working with Filament, as I just saw they don't use the same macroable code as Laravel

this break a lot, after regenerating my file (I don't do it often), most of my macros stopped working. 

- #1704
- #1702
- #1705
- #1690
- #1685

PR #1631, opened 2024-12-19
fix: add @template TModel of static for Eloquent

imzyf:
## Summary

fix: https://github.com/barryvdh/laravel-ide-helper/issues/1572

adding this docbloc for TModel of Eloquent class in ide-helper fix the return type.

## Type of change


joelstein:
This works for me with something like `MyModel::find(123)`.

chack1172:
Working for me

imzyf:
> Can we have a test, please?

Hi @mfn 

Regarding writing test cases, do you have any ideas or suggestions? I'm not quite sure how to write test cases for this.

barryvdh:
I think we need the other templates as well, right? https://github.com/barryvdh/laravel-ide-helper/pull/1634

chack1172:
@barryvdh I've checked and this is the only class @template used by methods if I'm not wrong. Anyway I'll make a PR using the change made in ReflectionDocBlock to dynamically detect class templates instead of hardcoding it in the view.

barryvdh:
Yeah there where a few more, TValue or something

PR #1635, opened 2024-12-28
Update testsuite for Generator, simplify service provider and mock

barryvdh:
## Summary
I think the models were not typehinted because of Lumen etc. Now that we've bumped versions, we can skip Lumen.

Also simplified the serviceprovider and the mock a bit, to allow for testing the Generator.

The Filesystem mock only mock the Put method now, rest is not mocked (partial mock).

## Type of change

### Checklist
- [ ] Update the README.md


barryvdh:
Ah yes, of course the actual file is different based on the PHP version, which makes it a bit difficult I guess 🤔

PR #1653, opened 2025-01-07
feat: use generics of return type to determine resulting models

Bloemendaal:
## Summary

PHPStan requires you to specify the generics in a docblock when using MorphTo (and other relations) like this:

```php
/**
 * @return MorphTo<LocationComplex|LocationAccommodation|LocationRoom, $this>
 */
public function location(): MorphTo
{
    return $this->morphTo();
}
```

Since this is already in the same model, we can use these generics (if specified) to determine the type of MorphTo. This would save (at least me) the trouble of having to write a custom hook for each polymorphic relation that I use. :)

This will return the following docblock:
```php
/**
 * @property-read LocationComplex|LocationAccommodation|LocationRoom $location
 */
```

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -819,10 +819,14 @@ class ModelsCommand extends Command
                                         $relation === 'morphTo'
                                     )
                                 ) {
+                                    $matches = [];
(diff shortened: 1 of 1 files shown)


barryvdh:
Thanks

PR #1672, opened 2025-02-07
Laravel 12 support

jonnott:
Just putting this out there to get ahead of the curve on supporting Laravel 12 when it's released on 24th Feb 2025.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
         "illuminate/view": "^11.15|^12",
         "mockery/mockery": "^1.4",
         "orchestra/testbench": "^9.2|^10",
-        "phpunit/phpunit": "^10.5",
+        "phpunit/phpunit": "^10.5|^11.5.3",
         "spatie/phpunit-snapshot-assertions": "^4 || ^5",
         "vimeo/psalm": "^5.4",
         "vlucas/phpdotenv": "^5"
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
         "illuminate/config": "^11.15|^12",
         "illuminate/view": "^11.15|^12",
         "mockery/mockery": "^1.4",
-        "orchestra/testbench": "^9.2",
+        "orchestra/testbench": "^9.2|^10",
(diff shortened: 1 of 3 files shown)


jonnott:
> We also need to adapt the CI workflows, otherwise we don't know if it works.

I think I've done that know .. to the best I know how, anyway. Feel free to propose further changes!

composer.json
```diff
@@ -25,18 +25,18 @@
...
         "ext-pdo_sqlite": "*",
         "friendsofphp/php-cs-fixer": "^3",
-        "illuminate/config": "^11.15",
-        "illuminate/view": "^11.15",
+        "illuminate/config": "^11.15|^12",
+        "illuminate/view": "^11.15|^12",
         "mockery/mockery": "^1.4",
-        "orchestra/testbench": "^9.2",
+        "orchestra/testbench": "^9.2|^10",
         "phpunit/phpunit": "^10.5",
```
KentarouTakeda, line 40: CI is reporting the [following error](https://github.com/barryvdh/laravel-ide-helper/actions/runs/13214322327/job/36892356377?pr=1672):

```
orchestra/testbench 10.x-dev requires phpunit/phpunit ^11.5.3
```

So the following should fix the dependency issue:

```suggestion
        "phpunit/phpunit": "^10.5|^11.5.3",
```

KentarouTakeda (changes requested):

jonnott:
> Windows seems to be a bit icky about not escaping ^ in the requirements, but should be good now :)

There's a note about that issue for Windows here: https://getcomposer.org/doc/articles/versions.md#writing-version-constraints (under the 'Caret Version Range' section) 🤯

jonnott:
> Yeah I saw that but not sure how to cleanly do that in the yaml for windows only.

Indeed, you wouldn't want that escaping in the yaml and it'd probably break for non-Windows contexts. I assume it's something that needs to be configured differently somehow by GitHub on their CI setup for Windows in terms of how it translates the version constraints to the CLI..

PR #1674, opened 2025-02-11
Add extends declaration for Macroable classes to fix missing inherited methods

KentarouTakeda:
### Overview

Currently, Laravel IDE Helper does not generate `extends` declarations for Macroable classes in their DocBlock. For example, `Illuminate\Http\Request` extends `Symfony\Component\HttpFoundation\Request`, but this inheritance is missing in the generated helper file.

As a result, tools like [PHPStan](https://phpstan.org/) or [PHP Intelephense](https://github.com/bmewburn/vscode-intelephense) sonetimes fail to detect the inherited methods and constants, triggering errors or warnings such as:

```php
use Illuminate\Http\Request;

function foo(Request $request)
{
    $request->isMethod(Request::METHOD_GET);
}
```

```
Access to undefined constant Illuminate\Http\Request::METHOD_GET
Call to an undefined method Illuminate\Http\Request::isMethod()
```

### What this PR Does

- **Adds an `extends` declaration** to the DocBlock for Macroable classes so that PHPStan, Intelephense, and other tools can properly recognize inherited methods and constants.
- **Facades are always omitted**, since they do not truly extend the underlying class but rather proxy calls to an actual instance.

### Additional Notes

- In a default Laravel configuration, `Illuminate\Http\Request` is effectively the only Macroable class that relies on a parent class. Without an `extends` declaration, any inherited methods or constants remain hidden from static analysis.
- After applying this PR, the generated DocBlock for `Illuminate\Http\Request` gains the `extends` line:

```diff
22382c22382
<     class Request {
---
>     class Request extends \Symfony\Component\HttpFoundation\Request {
```

## Type of change

### Checklist

### DIFF
--- a/resources/views/helper.php
+++ b/resources/views/helper.php
@@ -29,7 +29,7 @@ $s3 = $s1 . $s2;
 <?php foreach ($namespaces_by_extends_ns as $namespace => $aliases) : ?>
 namespace <?= $namespace === '__root' ? '' : trim($namespace, '\\') ?> {
     <?php foreach ($aliases as $alias) : ?>
-<?php echo trim($alias->getDocComment($s1)) . "\n{$s1}" . $alias->getClassType() ?> <?= $alias->getExtendsClass() ?> {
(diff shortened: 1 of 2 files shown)


KentarouTakeda:
@barryvdh 
Hello! Are you planning to release a new version with this fix?

Both versions seem to work fine, this fix and the one on top of the current dev-master. I've tried both on a ~100k line project I work on. I was able to remove over 100 instances of `@phpstan-ignore` and everything worked fine!!

Thanks for the great tool.

PR #1677, opened 2025-02-13
avoid adding duplicate tags

koshuang:
## Summary

When running `php artisan ide-helper:models --write` command, it sometimes will add duplicate phpdoc comment for the same property.

I added a check before adding a new tag for `ModelsCommand`. It will not create a new tag if it's a duplicate tag.

## Type of change


barryvdh:
ARen't we already checking if the property has a tag (the in_array check above)

koshuang:
Ah, it happens on v2.13.4. But it's fixed on v2.15.1. I will close this PR. Thanks!

PR #1695, opened 2025-04-08
fix: Add complete Request class methods to prevent Intelephense conflicts

### LINKED ISSUE #1693: Conflicting with Intelephense
### Versions:
- ide-helper Version: 2.15.1
- Laravel Version: 10.48.29
- PHP Version: 8.3.17

### Description:

I’m getting conflicts with this package, and the [PHP Intelephense][1] extension in Visual Studio Code.

IDE Helper seems to be generating a stub for the `Illuminate\Http\Request` class, that contains only a single method declaration (for the `user` method), so then Intelephense reports any other method on the `Request` class (`input`, `route`, etc) as undefined.

I’m not even sure why/when the IDE Helper package generated this stub file, as I only use IDE Helper to generate auto-completions for models using the `php artisan ide-helper:models -N` command.

[1]: https://marketplace.visualstudio.com/items?itemName=bmewburn.vscode-intelephense-client

Amirhf1:
This PR fixes the issue with Intelephense extension in VS Code by adding complete request class methods to the generated stubs.

   Changes:
   - Added new config option `include_complete_request_stubs` (default: true)
   - Added complete list of Request class methods with proper return types
   - Modified Generator class to handle Request class specially

   Fixes #1693

### DIFF
--- a/config/ide-helper.php
+++ b/config/ide-helper.php
@@ -49,6 +49,18 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Request stub generation
+    |--------------------------------------------------------------------------
+    |
+    | Set to true to generate complete Request class stubs including all methods.
+    | This helps prevent conflicts with PHP Intelephense extension.
+    |
+    */
+
+    'include_complete_request_stubs' => true,
+
+    /*
+    |--------------------------------------------------------------------------
     | Factory builders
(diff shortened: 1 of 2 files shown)


erikn69:
I think it can be improved, making it more configurable, like a list of the classes that will be completely mapped, and making it automatic, so you don't have to manually define the methods.

PR #1712, opened 2025-05-15
Add magic *_exists properties

erikn69:
## Summary
This adds `*_exists` model properties, similar to the `*_count` properties.

I also added a `write_model_relation_exists_properties` config to disable it.

## Type of change

### Checklist
- [x] Update the README.md


barryvdh:
Hmm it could be confusing when they're not actually there. So not sure if we should make it false by default

Issue #1725, opened 2025-08-13
Facade macro not generating


WalterWoshid:
### Versions:
- ide-helper Version: 3.6.0
- Laravel Version: 12.22.1
- PHP Version: 8.4.11

### Description:

```php
class MyService
{
    use Macroable;

    // ...
}

class MyFacade extends Facade
{
    protected static function getFacadeAccessor(): string
    {
        return MyService::class;
    }
}

class MyServiceProvider extends ServiceProvider
{
    public function boot(): void
    {
        MyFacade::macro(
            name: 'myMethod',
            macro: function (): void {
                dd("TEST");
            },
        );
    }
}
```

It generates a helper for `MyService::myMethod()`, but not for `MyFacade::myMethod()`.

Probably gonna open a PR in a couple of hours for a fix.

WalterWoshid:
Have some draft here, for anyone trying to complete it: https://github.com/barryvdh/laravel-ide-helper/pull/1726

Issue #1247, opened 2021-08-26
PHPDOC for custom facades


chirukinbb:
Hello, I write class with custom logics of my laravel ap
and make it as facade by this code
code of my facade
`
namespace App\Facades;

use Illuminate\Support\Facades\Facade;

class AppService extends Facade
{

    protected static function getFacadeAccessor(): string
    {
        return 'App\Services\AppLogic';
    }
}
`
code of my service provider
`
namespace App\Providers;

use App\Services\User;
use Illuminate\Support\ServiceProvider;

class AppLogicServiceProvider extends ServiceProvider
{
    /**
     * Register services.
     *
     * @return void
     */
    public function register()
    {
        $this->app->bind(AppLogic::class);
    }
}
`
in config/app.php
`
'providers'=>[
...
\App\Providers\AppLogicServiceProvider::class
]
`
methods of facade can be called as facade but I can't generate phpdocs for it.
I run
php artisan clear-compiled
and
php artisan ide-helper:generate
but in _ide_helper.php class AppService  not exists
if i write
AppService ::someMethodFromServiceLogic() it work

my question - how to generate phpdocs?

PR #1726, opened 2025-08-13
Draft: Add support for macroable facades

WalterWoshid:
## Summary
This change fixes an issue where macros registered on a Facade were not being added to the Facade's class definition in the generated `_ide_helper.php` file.

Currently, if a Facade (e.g., `MyFacade`) points to a `Macroable` service (e.g., `MyService`), the generator correctly finds the macros but only adds the method definitions to the underlying `MyService` class in the helper file. While this provides some autocompletion due to the IDE's understanding of the facade pattern, it's not a complete or accurate representation. The Facade's own class definition in the helper file remains empty of the macro methods, even though they are callable at runtime via `__callStatic`.

This fix introduces new logic to the `Generator` class. It now correctly identifies facades whose underlying service is macroable. It then copies the method definitions for each macro and adds them directly to the Facade's dummy class definition within the helper file.

The result is that both the service class and the facade alias now have the correct method definitions, providing complete and accurate autocompletion that mirrors the framework's runtime behavior.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

### Checklist
- [ ] Update the README.md

Closes: https://github.com/barryvdh/laravel-ide-helper/issues/1725
Closes: https://github.com/barryvdh/laravel-ide-helper/issues/1247 (Please don't use stale bot, was very hard to find that issue)

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -359,6 +359,14 @@ class Generator
         foreach ($macroable as $class) {
             $reflection = new ReflectionClass($class);
 
+            if ($reflection->hasMethod('getFacadeRoot')
+                && $reflection->getMethod('getFacadeRoot')->isStatic()
+            ) {
+                $method = $reflection->getMethod('getFacadeRoot');
+                $root = $method->invoke(null);
+                $reflection = new ReflectionClass($root);
+            }
+
             if (!$reflection->getStaticProperties()['macros']) {
                 continue;
             }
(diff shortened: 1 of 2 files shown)


WalterWoshid:
Macros that override existing methods also create an entry in `_ide_helper.php` which is wrong, because `__callStatic` is only called if the method doesn't exist on the underlying service class.

WalterWoshid:
Ugh, I have no clue how to solve that, and this code base (or laravel's facade system) is a mess.

PR #1739, opened 2025-10-28
Add support for `decimal` column type

### LINKED ISSUE #1738: `decimal` column type no longer being declared as `numeric`
### Versions:
- ide-helper Version: 3.6.0
- Laravel Version: 12.35.0
- PHP Version: 8.4.13

### Description:

Currently the `decimal` column type gets it's typehint as `string` instead of `numeric` as intended.

Compare with #1583 and #1582.

From what I understand decimal columns no longer have a cast associated with them but instead have a their own column type now.

So the check in the function https://github.com/barryvdh/laravel-ide-helper/blob/9ef25f60e70ced86f687ef6b9ffd9ac74a7c388a/src/Console/ModelsCommand.php#L391 no longer works (or no longer is enough) but instead/additionally the `match` statement here https://github.com/barryvdh/laravel-ide-helper/blob/9ef25f60e70ced86f687ef6b9ffd9ac74a7c388a/src/Console/ModelsCommand.php#L575-L588 needs and additional match for `decimal`, which then returns `numeric`.

I have locally tested adding the case `'decimal' => 'numeric'` and now the generated type hints are working as intended.

### Steps To Reproduce:

- Get table with a decimal column
- Generate model helpers
- Check generated typehint (should be `numeric`, is `string`)

BrainStone:
## Summary

Previously `decimal` columns were represented with a cast. This now is no longer the case and instead the column type itself is `decimal`. This change takes that into account.

This fixes #1738

## Type of change

### Checklist
- [x] Update the README.md

### DIFF
--- a/tests/Console/ModelsCommand/AllowGlobDirectory/__snapshots__/Test__test__1.php
+++ b/tests/Console/ModelsCommand/AllowGlobDirectory/__snapshots__/Test__test__1.php
@@ -42,8 +42,8 @@ use Illuminate\Database\Eloquent\Model;
  * @property float $float_not_nullable
  * @property float|null $double_nullable
  * @property float $double_not_nullable
- * @property string|null $decimal_nullable
- * @property string $decimal_not_nullable
+ * @property numeric|null $decimal_nullable
+ * @property numeric $decimal_not_nullable
  * @property int|null $boolean_nullable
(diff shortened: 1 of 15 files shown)


BrainStone:
MR is now ready for review and merging. I didn't spot anything in the README that needed updating, so no changes there.

PR #1745, opened 2025-12-09
Fix alias fake error

WentTheFox:
## Summary

The helper generation breaks when the `Laravel\Socialite\Socialite` facade's `fake` method (introduced in https://github.com/laravel/socialite/pull/756) is being called by the IDE helper, because the method has a required argument. In PHP 8.5 this causes an `ArgumentCountError` as seen below:

```
> Illuminate\Foundation\ComposerScripts::postUpdate
> @php artisan ide-helper:generate

   ArgumentCountError 

  Too few arguments to function Laravel\Socialite\Socialite::fake(), 0 passed in vendor/barryvdh/laravel-ide-helper/src/Alias.php on line 254 and at least 1 expected

  at vendor/laravel/socialite/src/Socialite.php:39
     35▕      * @param  string  $driver
     36▕      * @param  \Laravel\Socialite\Contracts\User|\Closure|array|null  $user
     37▕      * @return \Laravel\Socialite\Testing\SocialiteFake
     38▕      */
  ➜  39▕     public static function fake(string $driver, $user = null)
     40▕     {
     41▕         $root = static::getFacadeRoot();
     42▕ 
     43▕         if ($root instanceof SocialiteFake) {

      +18 vendor frames 

  19  artisan:35
      Illuminate\Foundation\Console\Kernel::handle()

Script @php artisan ide-helper:generate handling the post-update-cmd event returned with error code 1
```

ArgumentCountError screenshot

The goal of this change is to fully ignore the error so the `ide-helper:generate` command can complete without crashing, as it seems the missing `catch` block causes the error to be thrown despite being wrapped in `try`.

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/tests/AliasTest.php
+++ b/tests/AliasTest.php
@@ -9,6 +9,7 @@ use Barryvdh\LaravelIdeHelper\Macro;
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
 use Illuminate\Database\Query\Builder;
 use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Facade;
 
 /**
  * @internal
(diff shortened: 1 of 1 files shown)


Orrison:
Seeing this issue as well

FallenKnight85:
Experiencing this issue as well

WentTheFox:
@barryvdh I think the tests failed due to the
```
return ’’;
```
in the `MockFacade` class, that might have to be replaced with an actual class string. I did not see that earlier since the pipelines were waiting for approval.

barryvdh:
Yes fixed that already; https://github.com/barryvdh/laravel-ide-helper/commit/c95825a7452769cabdb10f18691d6eb1063d31e4

barryvdh:
This is tagged, so should be good now.

PR #1746, opened 2025-12-10
Skip calling fake() when required parameters exist (Socialite compatibility)

Jaspur:
This PR prevents `ide-helper` from calling `Facade::fake()` on facades whose `fake()` method requires parameters.
Laravel Socialite expects a required `$driver` argument, causing an `ArgumentCountError` during `php artisan ide-helper:generate`.

By checking the number of required parameters via reflection and skipping such facades, the generator behaves correctly without affecting other facades that support parameterless faking.

### DIFF
--- a/src/Alias.php
+++ b/src/Alias.php
@@ -239,27 +239,26 @@ class Alias
     protected function detectFake()
     {
         $facade = $this->facade;
-    
+
         if (!is_subclass_of($facade, Facade::class)) {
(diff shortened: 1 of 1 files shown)


barryvdh:
I think the error was resolved in https://github.com/barryvdh/laravel-ide-helper/pull/1745 but I guess this is the cleaner way.

PR #1774, opened 2026-03-17
Bump branch alias to 3.6

jnoordsij:
## Summary

When trying to test my application for Laravel 13 compatibility, it would not pick up the `dev-master` version of this package due to the `^3.6.1` constraint I currently have in place and the `3.5` alias in `composer.json` here.

This PR ensures that `dev-master` is a valid upgrade from the `^3.6` range. 

Alternatively, `3.7-dev` or `3.x-dev` would also seem to reasonable alternatives to me here,
so if wanted I'd be happy to change to either of these.

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,7 +66,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "3.5-dev"
+            "dev-master": "3.6-dev"
         },
         "laravel": {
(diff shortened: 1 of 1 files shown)


barryvdh:
let's make it 3.7

PR #1366, opened 2022-08-17
Fix return value of query scopes from parent class

sforward:
This PR fixes the return value of query scope methods that are defined in a parent class of the target model. At the moment the return value contains the class name where the method is located: `@method static \Illuminate\Database\Eloquent\Builder|PostParent active()`, and with this change it correctly becomes `@method static \Illuminate\Database\Eloquent\Builder|Post active()`. See included test for more context.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

### Checklist
- [x] Existing tests have been adapted and/or new tests have been added
- [x] Add a CHANGELOG.md entry
- [ ] Update the README.md
- [x] Code style has been fixed via `composer fix-style`

PR #1796, opened 2026-08-13
fix: resolve scope builder class against the model being documented

### LINKED ISSUE #1795: Missing FQDN for Builder in child Model class which doesn't import Builder
I have a `ModelA`. It defines a `scopePrivateOnly`. 

When I generate phpdocs, for `ModelA` I get the following which is fine. `ModelA.php` includes `\Illuminate\Database\Eloquent\Builder`
```
 * @method static Builder<static>|ModelA privateOnly()
``` 

scope method definition:
```
    public function scopePrivateOnly( Builder $query ): Builder
```

I have `ModelB`, which extends `ModelA`. The phpDoc gets generated as the following, despite `ModelB.php` not importing `Builder`.
```
 * @method static Builder<static>|ModelB privateOnly()
``` 

When `ModelA` _has_ the import, `ModelB` doesn't use the FQDN. When `ModelA` lacks the import, `ModelB` uses the FQDN. `ModelB` seems to follow whatever is happening in `ModelA` without deference to the imports in `ModelB`.

tldr:
when parent class file imports `\Illuminate\Database\Eloquent\Builder`, the child class scope* methods refer to `Builder` which may not be imported in the child class file.

This is kind of annoying because regenerating the docs can now lead to unknown class warnings from the IDE - because Builder isn't imported in the child classes file.

lazerg:
## Summary

Scopes inherited from a parent model were annotated with whatever name the parent file uses for the builder. If the parent imports `Illuminate\Database\Eloquent\Builder` but the child does not, the child model gets `@method static Builder<static>|Child scope()` and the IDE reports an unknown class.

The builder class is now resolved against the model the docblock is written to, the same way the model name already is since #1366.

Fixes #1795

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/src/Console/ModelsCommand.php
+++ b/src/Console/ModelsCommand.php
@@ -712,7 +712,7 @@ class ModelsCommand extends Command
                         //Remove the first ($query) argument
                         array_shift($args);
                         $builder = $this->getClassNameInDestinationFile(
-                            $reflection->getDeclaringClass(),
+                            new ReflectionClass($model),
                             get_class($model->newModelQuery())
                         );
                         $modelName = $this->getClassNameInDestinationFile(
--- a/tests/Console/ModelsCommand/QueryScopes/Models/Article.php
+++ b/tests/Console/ModelsCommand/QueryScopes/Models/Article.php
@@ -1,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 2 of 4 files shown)


afk11:
Confirmed fixes #1795 , thanks @lazerg 🙏(",)

